PR #2, opened 2017-05-01
proofing Task_List.txt

reagle:
My editor is set for spaces not tabs, if this messes up Zim, let me know.


reagle:
When will this be live? Presently, there is information that is incorrect and confusing.

jaap-karssenberg:
It's now part of the baseline code, so it will be automatically included in
the next release. As I wrote to the mailing list, I'm working on two
critical issues at this moment, as soon as I have solutions for those a new
release will be pushed out.

Regards,

Jaap

PR #20, opened 2017-05-05
Re-enable the OS X menu code

brechtm:
It is not clear why this code was commented out in be1e73ff9.

### DIFF
--- a/zim/gui/__init__.py
+++ b/zim/gui/__init__.py
@@ -370,29 +370,29 @@ class GtkInterface(gobject.GObject):
 
         # Adapt the GUI to OS X conventions
         ### TODO adapt this to support multiple toplevel windows per process ###
-		#~ try:
-			#~ import gtkosx_application
-			#~ macapp = gtkosx_application.Application()
-#~
-			#~ # move the menus to the OS X menu bar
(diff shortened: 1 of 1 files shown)


brechtm:
Oops. I missed to spot the TODO.

jaap-karssenberg:
Probably what we need is to make this code into a small object on it's own, including it's own functions to handle the menu items. This object should be loaded by the ZimApplication object defined in `zim/main/__init__.py`, e.g. in the `add_window()` method when the first window is loaded. Giving the object a reference to the "_windows" attribute should allow the menu actions to identify what window has the focus and delegate the call to that window.

I can make a proposal, but will need help testing is as I have no OS X device ..

jaap-karssenberg:
P.S. key design change is that this "menu object" should not be tied to a single window, windows can come and go, but the menu should stay as long as the application is running

brechtm:
I've had a quick look at the code, but I'm not sure how to proceed since I'm not familiar with GTK. If you can provide a basic implementation of your suggestion, I can test and possibly debug it.

jaap-karssenberg:
Closing request in this form - please open an issue to request the fix, I will look into providing the boilerplate code for it

PR #59, opened 2017-05-23
bug: cursor, selection bounds, and scroll values are lost when append…

Osndok:
…ing text via a plugin

See:
https://bugs.launchpad.net/zim/+bug/1631238

### DIFF
--- a/zim/gui/pageview.py
+++ b/zim/gui/pageview.py
@@ -684,6 +684,11 @@ class TextBuffer(gtk.TextBuffer):
 
 		@param tree: a L{ParseTree} object
 		'''
+
+		# Since we cannot hold onto a TextMark or TextIter through a clear(), we must use offsets
+		cursorOffset = self.get_iter_at_mark(self.get_insert()).get_offset()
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
Need to review carefully whether there are no unintended side affects in this case. When using ``set_parsetree()`` resetting the buffer state is the expected behavior.

Probably you should use ``insert_parsetree()`` instead. Looking at the implementation it should actually remember the cursor position. 

https://github.com/jaap-karssenberg/zim-desktop-wiki/blob/master/zim/gui/pageview.py#L697

Osndok:
I'll definitely look into that! Being some time since I wrote this patch, I don't remember specifically why.... but I think that the insert function either delegated to the set_parsetree function (at the time), or that insert was not available through the plugin interface.

jaap-karssenberg:
Afraid I can't merge this change because the "set_parsetree()" method is the wrong place to handle this logic. Would suggest it is done in "set_page()" to be triggered on page changes. For appending it would not be needed at all if you use the "insert" method.

PR #60, opened 2017-05-23
better pasted image file names

Osndok:
See:
https://bugs.launchpad.net/zim/+bug/1631502

### DIFF
--- a/zim/gui/clipboard.py
+++ b/zim/gui/clipboard.py
@@ -11,6 +11,7 @@ straight forward API.
 
 import gtk
 import logging
+import time
 
 from zim.fs import File, Dir, FS
 from zim.newfs import LocalFolder
@@ -236,7 +237,10 @@ def parsetree_from_selectiondata(selectiondata, notebook=None, path=None):
 			# but is quite large to store, so compress by using png
 			format, extension = 'png', 'png'
 
-		file = dir.new_file('pasted_image.%s' % extension)
+		prefix = path.name.replace(':','-')
+		nameAndTime = prefix+'-'+("%x" % time.time())
+
+		file = dir.new_file(nameAndTime+'.'+extension)
 		logger.debug("Saving image from clipboard to %s", file)
 		pixbuf.save(file.path, format)
 		FS.emit('path-created', file) # notify version control


jaap-karssenberg:
I like the proposal. Using timestamps is straight forward. The basename however I'm more reluctant since it is hard to oversee unintended side effects. For one I'm thinking there should be a max length on the name somehow. Maybe just using basename of the page?

Osndok:
I agree that a limit on the filename length might be wise, but I would very much like to preserve as much of the path as acceptable.

If you would like to dictate that maximum length in characters... I'll work on a way to cap it the shallowest or deepest segments that still keep it under that threshold (unless it is one long pagename, I guess).

I would also be interested in hearing if you think the truncated name should be the "shallowest N":
**:A:B:C -> A-B-592374e6.png**

...or the deepest:
**:A:B:C -> B-C-592374e6.png**

Osndok:
For way of a reference point "**Journal-2017-05-22-592374e6.png**" is **32 characters**.

jaap-karssenberg:
Max length is arbitrary, but would keep it a readable length. My logic would be to only use the basename of the page, because that is a length that for sure can be used. Also it still makes sense after moving a page (but admittedly not after renaming). 

Thinking about it more maybe I'm more in favor to keep "pasted image" because that is what it is - anything else is a guess (unless you get some meta-info from the clipboard). But I can imaging making this configurable somewhere as a preference.

Btw. sorry for not reading carefully before, but why are you doing timestamp as hex? Just a date-time string would be better readable for the user and sort better.

For comparison, screenshot on Gnome results in: "Screenshot from 2017-05-23 17-40-36.png"

Osndok:
> My logic would be to only use the basename of the page,

That really doesn't work for me, because I have many fine-detail page names, which are often deepest with just a number (like the journal pages), and would leave nearly no information in the final file name, like: **5-592374e6.png** (which is not useful at all)

> [Base name] still makes sense after moving a page (but admittedly not after renaming).

This is a matter of state, so anything (once decided) has the possibility of becoming stale. IMO, even a stale/old/inaccurate description is better than one without meaning!

>  I'm more in favor to keep "pasted image" because that is what it is.

But that is not correct, and only makes sense from the context of Zim development. "pasted image" is only the **providence** of the image, and it says nothing about what is actually represented **inside the image**.... which I think should be obvious, because it is just a static string, so it cannot really add any value (unless, as a developer, you routinely reprocess or exclude all pasted images, for example).

The "page name" may not be exact, but it is at least an approximation.

Here is an example... suppose someone is collecting cat memes under ":Cat:Meme", and they use the paste workflow... under my proposal, all the filenames will be of the form "Cat-Meme-XXXXXXX.png"... which is, exactly, what the image contains. So if he/she sends that file to his phone, or a friend, or uploads it to a website... there is **already** a good hint of what the file is with *zero* effort from the user.

And there are other benefits...
* when doing a filename search a static string will match either none or all; whereas searching for "*cat*.png" ... **might** be useful to someone
* when "cleaning up", an experienced user might be able to detect a distinct pattern of files that "came from zim", and confidently delete them (because they are safe in zim's git repo).

> why are you doing timestamp as hex? Just a date-time string would be better readable for the user and sort better.

My first attempt used a standard decimal time format, and the results (particularly for journal pages, which already include the date) were not satisfactory.

It's primary use is a **discriminator** (for when people are working on the same repo in parallel), rather than being readable... the hex string is more compact (esp. since I needed seconds-level precision), and I had actually considered at one time only using  a portion of it (the least signifant digits), but settled on the fact of it being reversible "if needed" might have some value where a modulus would not.

As for sorting, I'm quite sure that the hex code would sort just as well as the decimal (assuming most-siginificant digit first)... as the numerals (0-9) come before letters (a-z) in the ascii table.

> For comparison, screenshot on Gnome results in: "Screenshot from 2017-05-23 17-40-36.png"

Quite right; and that is because gnome really doesn't know "what" is under an arbitrary rectangle of the screen, but (if we are pasting it into a Zim page)... we sorta do!

jaap-karssenberg:
> "pasted image" is only the providence of the image, and it says nothing about what is actually represented inside the image.... which I think should be obvious, because it is just a static string, so it cannot really add any value (unless, as a developer, you routinely reprocess or exclude all pasted images, for example).
> 
> The "page name" may not be exact, but it is at least an approximation.

I do follow your reasoning, but not sure I fully agree.  E.g. in your example of pasting "Bill-Clinton.png" I might just as well paste an image of the white house to that page and the name would still be loosely related to the content of the image. So for me keeping the info that it is a pasted image helps to signal it might not be accurate.

And yes, it is reasonable to want to be able to parse the name from a programmatic point of view. E.g. a plugin that cleansup all pasted images that are no longer referred to in a page makes sense to me.

How about "pasted image -- pagename -- datetime" ? This keeps the info that it was pasted (and thus "pagename" could not be what is in the image after all.

You could also query the clipboard for textual information. At least noticed that when I paste an image from firefox the original image name is also available through the clipboard. Of course, this original name couls also turn out to be something like "image 1", so no guarantees that it is meaningful.

>> For comparison, screenshot on Gnome results in: "Screenshot from 2017-05-23 17-40-36.png"

> Quite right; and that is because gnome really doesn't know "what" is under an arbitrary rectangle of the screen, but (if we are pasting it into a Zim page)... we sorta do!

I intended this as an example of usage of date and time. 

Bit surprised that while for the name you emphasis user readability, you choose a different direction for the time. If you include time, you might as well make it such that it is readable (and searchable!) for the user.

Osndok:
> So for me keeping the info that it is a pasted image helps to signal it might not be accurate.

That's really interesting. Personally, I don't consider pasted content to be in any way more deprecated or inaccurate.

> How about "pasted image -- pagename -- datetime" ? This keeps the info that it was pasted (and thus "pagename" could not be what is in the image after all).

I think that it's better than the current scheme. "Image" doesn't really add much value b/c the file extension indicates if it is an image or not.

I'm interested that you would want "pasted" to be the most prominent part of the file name. It seems really important to you, and like unwanted noise to me... therefore, I get the impression that we use zim quite differently, somehow... of differ in our view of the content that we create therein.

I have a similar dislike for all the screenshots being named "screenshot", as in all my zim usage patterns, using the screenshot function for a "screenshot" (such as, to document an interface, record overall screen state, ask for help, etc) accounts for <1%.... almost always it is an instant "crop-and-capture" that bypasses an editing stage.

...so the only 'official' way to insert an image into zim (without them all being labeled 'pastes' and 'screenshots') is also, by **far** the longest (in terms of steps and time) and also the most rote (I prefer dense, terse, effective tools... over guided interaction wizards).

> You could also query the clipboard for textual information.

That may be a good part of a solution, but I think you would still need to add a discriminator. Originally it was an incrementing integer (which doesn't work at large), and I changed it to a hex-time in seconds (which is less broken).

Another, more far-out, idea... would be if something like Google's machine-vision apis to **actually** describe the main object in the image.

> Bit surprised that while for the name you emphasis user readability, you choose a different direction for the time. If you include time, you might as well make it such that it is readable (and searchable!) for the user.

Ha ha... yes... I can see why you would say that...

Of course, **I** don't see it as any sort of contradiction or self-inconsistency. To me... **when** the image was pasted (or screen-capped) is just as much of the (comparatively irrelevant) **providence**... the acquire date does not actually relate to the image (in terms of referencing), and can be recovered by the git revision control system (which also highlights contemporaneous changes).

Said another way, when I instruct zim to insert an image... I see myself as declaring that the image belongs in zim, in that notebook, **on that page**, at a certain text position... I'm not declaring the pasting important... or the screenshotting... or the attaching....

Given this conversation, I'm a bit surprised that http://zim-wiki.org/ (which I thought was composed and maintained with zim) does not have all it's image filenames as: "pasted_image*", "screenshot*", and "attachment*"... :)

jaap-karssenberg:
Summarizing discussion:
* keep "pasted_image" prefix - page name may be misleading in some cases (unless file name available from clipboard)
* use page name - truncate on the left if more than x characters
* hex to avoid merge conflicts, not to indicate time of pasting

Osndok:
That all sounds fine to me, provided that you don't mind the prefix string being a config item (so I can make it the empty string locally).... In which case, I guess I need to code up the changes, huh?

jaap-karssenberg:
Sure, we could even make the whole string format a config item. (In which case I would suggest "%" escapes for the format fields - but maybe this becomes bit complex if just a prefix string would do fine.

Up to you if you want to code it up - we can also put it on the total 0.68 release list, which means I will get to it at some point, but no promise when exactly.

stewit:
Just one additional idea to avoid merge conflicts: Why not use a hash of the image file as a component of its file name?

In addition I would like to point out, that giving somehow "unique" names to pasted image files in combination with storing them in a seperate, notebook-global directory (maybe directly under the notebooks root directory) would allow to automatically "repair" broken image links that occur when copy-pasting content between ZIM wiki pages. 

The (backwards-compatible) logic would be:
if an image-link points to the attachment dir of the page and the basename starts with pasted_image and the file does not exist there, make the image link point to %GLOBAL_PASTED_IMAGES_DIR/same_filename

The same thing could be considered for equations: store them with a hash of their tex in their file name and in  a global directory to avoid issues mentioned [here](https://github.com/jaap-karssenberg/zim-desktop-wiki/issues/45)

Osndok:
Using a hash in this case is probably not optimal... you might make a half-step towards sub-second collision avoidance, but you would lose natural file sorting, and have much longer file names.

However, if Zim accepted this patch and the **only difference** was using a hash rather than a hex(unix_time)... I would probably accept it, and drop that part of my parallel branch; I guess I just got scared when Jaap suggested & backed some counter formats which I thought would actually be a step backwards (IMO).

As for the other thing... ordinarily, I'm **all about** the content-addressed-storage!

I like the way you think, but if Zim were to go CAS, it would be moving away from one of the key things that attracted me to zim: that it stored all of it's files in an understandable layout, rather than hidden behind a binary database or needing to be meticulously 'reconstructed' if Zim (for whatever reason) died or did not work out.

...and technically, if you are saving your notebook to a git repo, you already have a CAS backend :)

Back to Jaap's point on configuration... it seems to me that this is a similar problem to "code style" holy-wars. That people may be very opinionated, or have specific use cases for a variety of formats; which might also mean that it is both global (in the case of one **creating** new notebooks) and notebook-specific (for one that might be **sharing** a notebook and not want it polluted with "incorrectly named images")... and at that point, it is too much work for me to recommend (when I can just have it set right for myself and no one else is complaining).

jaap-karssenberg:
Patch needs to be updated to reflect discussion outcome

PR #94, opened 2017-06-01
Cleanup: Make code more compliant to pep8

genodeftest:
I included all non-controversial changes suggested by autopep8 and ran it on the Zim source tree.

Note: These changes might need to be redone since git master changed a lot. Please tell me when to do this, because probably any other change will need to be rebased after merging this pull request, as it touches many lines of many source code files.

__EDIT:__ Please note that the commits included in this pull request touch so many lines that they will make most git GUIs go haywire.


jaap-karssenberg:
Yes I think these make sense, especially as part of preparing the code base for moving to python3 at some point in the future.

Maybe the right place would be shortly after the final 0.67 release before starting new round of changes.

I did not review all of the filters, but I noticed that the indentation "fix" in many places changes manual indenting I did specifically for readability of the code. 

For example:

if ... \
and ... :

becomes

if ... \
   and ... :

which is not what I intended.

I fully realize this is taste without an objective way to decide. Maybe do the other fixes first, so it is easier to cherry pick.

Cheers,

Jaap

genodeftest:
> Maybe the right place would be shortly after the final 0.67 release before starting new round of changes.

Ok. Just give me a hint then (e.g. post a comment here) and I'll try to rebase/regenerate the changes.

> I did not review all of the filters, but I noticed that the indentation "fix" in many places changes manual indenting I did specifically for readability of the code.

I've seen that with [Exaile](https://github.com/exaile/exaile) (Gtk+/Python based media player) too. I can skip them easily.

genodeftest:
As 0.67 is released, are you interested in this pull request for the next few days?

jaap-karssenberg:
Yes, please  go ahead,  I'll hold my commits until this one is in.

Could you also check in a shell script under "tools/" in case we want to re
run it after merging branches ?

Thanks,

Jaap

genodeftest:
> For example:
> ```
> if ... \
> and ... :
> ```
> becomes
> ```
> if ... \
>    and ... :
> ```
> which is not what I intended.

Hm, pep8 is quite clear about that: https://www.python.org/dev/peps/pep-0008/#indentation. Does this change your mind? Or should I remove that change from all the cleanup stuff here?

genodeftest:
Next try done. I'll clean up unwanted octals after this change has been merged, as that is only a minor cosmetical issue.

jaap-karssenberg:
I reviewed the changes (skimming through the largest ones):
- tools: Add code cleanup file b3814d5 -- Good
- Fix line separation characters  … 4bfa0a8 -- Good
- Autopep8: Indentation contains mixed spaces and tabs  … 0159c64 -- NOK
    - Issue: seems like it replaces all tabs with spaces rather than just finding mixed cases
- Autopep8: Statements on separate lines  … 15744c0 -- Good
- Autopep8: More Statements  … 9f90b5c -- Good with fix
    - Amend: dates in test suite contain unintended octals (2 files!)
- Autopep8: Deprecation warning  … 1cab590 -- Good
- Autopep8: The backslash is redundant between brackets  … 6a03191 -- Good
- Autopep8: Imports  … 0ee14cb -- OK
- Autopep8: Whitespace in code  … 3848a1d -- OK
- Autopep8: Whitespace in comments  … 8f75e7f -- OK with fix
    - Revert changes of "`#:`" to "`# :`" these are inline doc strings used by epydoc
- Autopep8: Blank line  … d670f97 -- OK
- Autopep8: Intendation in comments  … 13399c9 -- NOK
    - Discussion: Removes indenting of "hanging comments" - these are intended as line comments for previous line, not sure if removing indent improves readability here
    - Issue: Removes indenting of commented-out-code - this is NOK because code no longer works after removing the comment char
- Autopep8: Intendation in code  … 972bb49 -- discussion, see below

Notes:
- remove `contrib/zim2trac.py` - not used anywhere, and could be on the wiki for those who want it (honestly forgot myself it was in the code base ...)
- Also check pep on formatting doc strings with `"""` instead of `'''` -- take along when gettin grid of epydoc format in doc strings

So apart from the issues with indentation I would like to check in the other patches. Since the indentation touches every file, I don't think I can cherry-pick. Probably the easiest if I exclude the indentation check and than regenerate using the tooling script, do you agree?

Will also look into adding test cases to prevent checking in new code that breaks the conventions:
* Current `tests/package.py` already contains a test for mixing whitespace and tabs -- thought I also put a `\r` test in, but seems I'm mistaken :(`
* Try using the [pycodestyle module](pycodestyle.readthedocs.io) to generate test cases or run autopep script and show proposed change as error message
* Special focus on "deprecation" warnings, make these errors such that we remain as close as possible to python3 to allow future back-porting of patches


-----
My issue for Intendation in code
--------------------------------

Statements like this:
```
if some rather long statement \
and some other statemend:
    so something
```

Now become:
```
if some rather long statement \
        and some other statemend:
    so something
```
Which to my taste reduces the readability considerable.

Would like to avoid these replacements, or rewrite them in a more
readable form. Not sure what would work best.

TODO: check how many occurences these really are - can we deal with enclosing brances to avoid triggering the indent rule ?


Manual tweaking for Intendation in code
---------------------------------------

```
some_variable = some_fucntion(arg1,
    arg2,
    arg3
)
```

becomes:
```
some_variable = some_fucntion(arg1,
                              arg2,
                              arg3
                              )
```

Here I would prefer the also compliant:
```
some_variable = some_fucntion(
    arg1,
    arg2,
    arg3
)
```

Need to review the patch and manually fix cases where this indenting is added to go the other way.

genodeftest:
> TODO: check how many occurrences these really are - can we deal with enclosing brances to avoid triggering the indent rule ?

I doubt that this works. I think, autopep8 will detect the braces and remove them.

> * Autopep8: Indentation contains mixed spaces and tabs … 0159c64 -- NOK
>   * Issue: seems like it replaces all tabs with spaces rather than just finding mixed cases

What is the problem with that? I think this behaviour is intended. The rule refers to files with mixed indentation, not lines (I guess). Files with mixed indentation are bad for readability, because editors don't agree on how wide a tab should indent (2, 4 or 8 characters). To make sure python code does not visually break, it really should have only one kind of indentation per file, tabs or spaces. For files which only use tabs for indentation and no spaces at all, replacing tabs by spaces is still useful because if you open that file the next time, your editor might intend by spaces, because that's the default in python.

I'll revert the last two changes (indentation stuff) and do the manual fixes, so most clean-ups can be merged. Then I'll try to re-run the indentation fixes with your notes above.

genodeftest:
> I'll revert the last two changes (indentation stuff) and do the manual fixes, so most clean-ups can be merged. Then I'll try to re-run the indentation fixes with your notes above.

Done.

> So apart from the issues with indentation I would like to check in the other patches. Since the indentation touches every file, I don't think I can cherry-pick. Probably the easiest if I exclude the indentation check and than regenerate using the tooling script, do you agree?

I basically agree (except for the indentation with tabs), and removed those changes for now, so you don't need to cherry-pick.

> * remove contrib/zim2trac.py - not used anywhere, and could be on the wiki for those who want it (honestly forgot myself it was in the code base ...)

If it is useful, why not leave it there and link to it from documentation? Having it in the wiki means having it in Github-controlled space, and I don't trust them too much. Having it in git makes sure everyone who `git clone`'d your repo has it.

To make sure that these changes don't get overridden, continuous integration tools like travis might be useful, preventing check-in if it breaks some rules.

jaap-karssenberg:
@genodeftest maybe I'm not looking at the right version of the changes, but looks to me this pull request still updates each and every line in the code base replacing tab with space ?

genodeftest:
@jaap-karssenberg : I've written this above:

> > * Autopep8: Indentation contains mixed spaces and tabs … 0159c64 -- NOK
> >   * Issue: seems like it replaces all tabs with spaces rather than just finding mixed cases
> 
> What is the problem with that? I think this behaviour is intended. The rule refers to files with mixed indentation, not lines (I guess). Files with mixed indentation are bad for readability, because editors don't agree on how wide a tab should indent (2, 4 or 8 characters). To make sure python code does not visually break, it really should have only one kind of indentation per file, tabs or spaces. For files which only use tabs for indentation and no spaces at all, replacing tabs by spaces is still useful because if you open that file the next time, your editor might intend by spaces, because that's the default in python.

And I think it still is the case. Having files with mixed indentation breaks readability, which is especially bad in python where indentation is important for control flow. If you are sure you still want me to *not* fix this, please refer to this comment.

jaap-karssenberg:
> What is the problem with that? I think this behaviour is intended. The rule refers to files with mixed indentation, not lines (I guess). Files with mixed indentation are bad for readability, because editors don't agree on how wide a tab should indent (2, 4 or 8 characters). To make sure python code does not visually break, it really should have only one kind of indentation per file, tabs or spaces. For files which only use tabs for indentation and no spaces at all, replacing tabs by spaces is still useful because if you open that file the next time, your editor might intend by spaces, because that's the default in python.

Well, as far as I checked most of my files use tabs only and not a mix. Using tabs instead of spaces never gave me any problems. Just replacing them by spaces because the pep prefers that is fixing something that for me is not an issue. I never understood this recommendation and the rationale is weak in my mind, but probably it is a matter of taste. Either way, I prefer not to patch each and every line in the project just replacing spaces and tabs.

On the other hand, there may be few cases were accidentally spaces and tabs were mixed. Always good to clean these up because they can lead to unexpected behavior. But finding these cases I'll replace them by tabs to be consistent with the rest of the code base.

genodeftest:
If you insist, here is the same set of changes but without fixing mixed spaces and tabs. Note that there are a few files with mixed indentation (e.g. `tools/generate_test_notebook.py` or `zim/inc/xdot.py`), but they are not covered by this change.

jaap-karssenberg:
Thanks for your patience. I realize my tab usage is debatable and depends on taste. But at this point I don't want to merge a whitespace replace over the whole code base.

Afraid that the result of the latest version is still replacing all (or many) tabs with spaces, so net effect is the same as before. Going through the individual commits, it seems that "Autopep8: Indentation in comments"  - 43077e2 - is the commit that does the big replace.

-- Jaap

jaap-karssenberg:
Committed the set of changes that I feel comfortable with as 4d8406ce86e217ce92cee9bb3c17b273ce6891bc

PR #98, opened 2017-06-07
ImageGenerator extras: access to current "Page" & optional default text/template

Osndok:


jaap-karssenberg:
Straight forward change to merge. Just wondering what plugin you are planning that would use these ?

Osndok:
My first use (which I have a working prototype for) is a change to my local diagram plugin, that provides a template sorta like this:
```
digraph PageBaseName
{
    node[...styles-i-like...];

}
```
So, you see, it needs both the "page.basename" (for the graph's name) and the default text (to reduce rote typing).

jaap-karssenberg:
Committed as 2a72d2ddcaaec3077c63ce6a7436314a64abb69e

PR #100, opened 2017-06-08
shorter links from Journal

e3rd:
I'd like to have shorter links from Journal. So I've added new plugin option.

image

If this option is set to True, when copying a link from Journal, only the short form gets pasted. 

image

For this case, only "diplomka" is pasted instead of "Journal:2015:05:diplomka" when copying location.

image

Please advise me, how to read the plugin preferences correctly? Now, I'm just examining ui from ZIM_APPLICATION.

### DIFF
--- a/zim/gui/clipboard.py
+++ b/zim/gui/clipboard.py
@@ -13,6 +13,7 @@ import gtk
 import logging
 
 from zim.fs import File, Dir, FS
+from zim.main import ZIM_APPLICATION
 from zim.newfs import LocalFolder
 from zim.notebook import Path
 from zim.parsing import is_url_re, url_encode, link_type, URL_ENCODE_READABLE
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
I understand the idea, but afraid I can't merge this change like this. You are putting plugin specific logic in one of the core modules, that is a bad practice. 

Roughly what would be needed here is that the PageView defines a hook which has the page as an argument and takes a 2-tuple of page and link text as return value. Now plugins can connect to this signal and provide an implementation.

My second question would be whether this is indeed specific for the journal. Wouldn't it make sense as a generic option? Of course you can have both - using the hook above for plugin overrides and provide a default implementation in case no one handles on the hook.

Have a look at the "suggest-link" signal defined in the notebook class.

Regards,

Jaap

e3rd:
Thanks for the guidelines, Jaap, I wanted indeed to discuss it first with you. I'll try something and pull another draft.

PR #121, opened 2017-06-15
Insert-link (and probably all other) popup page suggestions are much faster

Osndok:
For me, this lowers the worst-case punishment (e.g. for typing "a" in the search box), from two minutes to three seconds; and speeds up the common cases considerably too.

### DIFF
--- a/zim/gui/widgets.py
+++ b/zim/gui/widgets.py
@@ -1996,15 +1996,14 @@ class PageEntry(InputEntry):
 		if not self.notebook:
 			return # no completion without a notebook
 
-		text = self.get_text()
-		if self._current_completion:
-			if self._current_completion.startswith(text) \
-			and not ':' in text[len(self._current_completion):]:
-				return # nothing to update
-			else: # Clear out-of-date completions
-				model = self.get_completion().get_model()
+		if self._current_completion and self.get_completion():
+			# Clear out-of-date completions
+			model = self.get_completion().get_model()
+			if model:
 				model.clear()
-				self._current_completion = None
+			self._current_completion = None
+
+		text = self.get_text()
 
 		if not text or not self.get_input_valid():
 			return # can't complete invalid input
--- a/zim/gui/widgets.py
+++ b/zim/gui/widgets.py
@@ -2075,8 +2075,15 @@ class PageEntry(InputEntry):
 			return href.to_wiki_link()
 
 		model = completion.get_model()
-		childpos, peerpos = 0, 0
+		count, childpos, peerpos = 0, 0, 0
 		for p in self.notebook.pages.search_pagename_substring(text):
+
+			count=count+1;
+			# BENCHMARK: 1 second of UI hang for every 100 entries.
+			if count > 300:
+				logger.debug("truncating long results list to prevent GTK hang")
+				break
+			
 			link = relative_link(p)
 			if link.startswith('+'):
 				model.insert(childpos, (link, p.basename))
--- a/zim/gui/widgets.py
+++ b/zim/gui/widgets.py
@@ -2075,20 +2075,18 @@ class PageEntry(InputEntry):
 			return href.to_wiki_link()
 
 		model = completion.get_model()
-		lowertext = text.lower()
 		childpos, peerpos = 0, 0
-		for p in self.notebook.pages.walk():
-			if lowertext in p.basename.lower():
-				link = relative_link(p)
-				if link.startswith('+'):
-					model.insert(childpos, (link, p.basename))
-					childpos += 1
-					peerpos += 1
-				elif not ':' in link:
-					model.insert(peerpos, (link, p.basename))
-					peerpos += 1
-				else:
-					model.append((link, p.basename))
+		for p in self.notebook.pages.search_pagename_substring(text):
+			link = relative_link(p)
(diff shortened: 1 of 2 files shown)


zim/notebook/index/pages.py
```diff
@@ -539,6 +539,17 @@ def walk_bottomup(self, path=None):
 		page_id = self._pages.get_page_id(path) if path else ROOT_ID # can raise
 		return self._pages.walk_bottomup(page_id)
 
+	def search_pagename_substring(self, nameFragment):
+		candidates = self.db.execute(
+			'SELECT * FROM pages WHERE LOWER(name) LIKE ? ORDER BY parent,sortkey',
```
jaap-karssenberg: Original code matches on "basename" - matching on "name" gives different behavior. When you match pages with many children, matching "name" will include all children as well, which is not the intended behavior.

Afraid I don't know immediately the best way to match only the basename in an SQL query. Might need adding another column or using a SQL function to generate the basename.

Probably it would make sense to also add a unittest for this method in order to explicitly check the behavior. See ``tests/indexviews.py`` for similar tests on the ``PagesView`` object.

zim/notebook/index/pages.py
```diff
@@ -539,6 +539,17 @@ def walk_bottomup(self, path=None):
 		page_id = self._pages.get_page_id(path) if path else ROOT_ID # can raise
 		return self._pages.walk_bottomup(page_id)
 
+	def search_pagename_substring(self, nameFragment):
```
jaap-karssenberg: For consistency in the class I would suggest calling this `list_pages_match_substring(self, substring)`

zim/notebook/index/pages.py
```diff
@@ -539,6 +539,17 @@ def walk_bottomup(self, path=None):
...
+		candidates = self.db.execute(
+			'SELECT * FROM pages WHERE LOWER(name) LIKE ? ORDER BY parent,sortkey',
+			('%'+nameFragment.lower()+'%',)
+		).fetchall();
+
+		logger.debug("%d page-names contain '%s'", len(candidates), nameFragment);
```
jaap-karssenberg: Would remove these kind of debug messages before merge - they are useful while developing, but later on only add noise to the error log

zim/gui/widgets.py
```diff
@@ -2060,35 +2060,44 @@ def _fill_completion_for_anchor(self, path, prefix, text):
...
-					peerpos += 1
-				else:
-					model.append((link, p.basename))
+		count, childpos, peerpos = 0, 0, 0
+		for p in self.notebook.pages.search_pagename_substring(text):
+
+			count=count+1;
+			# BENCHMARK: 1 second of UI hang for every 100 entries.
+			if count > 300:
+				logger.debug("truncating long results list to prevent GTK hang")
```
jaap-karssenberg: Would remove these kind of debug messages before merge - they are useful while developing, but later on only add noise to the error log

zim/gui/widgets.py
```diff
@@ -1998,7 +1998,7 @@ def update_completion(self):
 
 		text = self.get_text()
 		if self._current_completion:
-			if text.startswith(self._current_completion) \
```
Osndok: Ha ha... yeah... I stared at that one cross-eyed for a while too... I think that the variable is serving two purposes. When **one** match is found, it is actually **longer** than the accepted text (so typing semi-clobbers it as you go). Either that, or it is just guarding an unintuitive break condition.

I noticed this after I tried to ignore one-character searches. With the original logic, it would try once, and never again (because "TestPage" **always** starts with "T"); with this swap.... it "works as expected".

zim/notebook/index/pages.py
Osndok: I'll think about this too... for now, though, I'm just happy to have the productivity boost!

zim/gui/widgets.py
jaap-karssenberg: Don't think that is right - only line where it is to a value other than `None` is here: https://github.com/Osndok/zim-desktop-wiki/blob/5dcb1500acdd5099423e46ca5161efc840320215/zim/gui/widgets.py#L2041

So should always be either the latest input text, or `None`. Have to admit that the variable name is confusing here.

Osndok: Well, then, I suppose there is something high-level about this check that is wrong, then. As I read it now, it seems to say "if the user added a non-colon, then we don't have to update the list of completions"... which is clearly incorrect, as (at the very least) adding characters is **expected** to further constrain the list of candidates.

jaap-karssenberg: No, this part of the code populates the list, but the filter function determines what is shown. So typing more text will constrain the results as the filter function selects matching entries. This optimization just means we don't bother to re-populate with a sub-set as the filter function already selects a subset.

nasedil:
Hi guys.  The issue with `link to...` and `jump to...` affects me pretty significantly too, so I wonder how can I help?  I wonder first of all, what is the state of this branch.  Are there functional problems that stop it from being used or we need to polish it?  Or maybe you have any comments or work-in-process regarding the issue that you would like to share before I start looking at the code?

jaap-karssenberg:
Out of date, performance of completion should be improved already since this request was opened

PR #122, opened 2017-06-15
append_text_to_page should call insert_parsetree if page is currently opened

Osndok:
Replaces:
* https://github.com/jaap-karssenberg/zim-desktop-wiki/pull/59
* https://bugs.launchpad.net/zim/+bug/1631238


jaap-karssenberg:
Needs to be updated to latest version.

Longer term plan is to make the "Page" object the owner of the "TextBuffer", that should also simplify this fix

PR #127, opened 2017-06-16
reveal an extra path segment, if elements of the pathbar would be ambiguous

Osndok:
It's not a 100% solution, as I suppose for this idea to be pragmatically correct we would have to iterate on the names, adding segments until they became unambiguous... but +1 segment I think is a step in the right direction.

### DIFF
--- a/zim/gui/pathbar.py
+++ b/zim/gui/pathbar.py
@@ -7,6 +7,8 @@ import gtk
 import gobject
 import logging
 
+from collections import Counter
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
Thanks - will merge after releasing 0.67

jaap-karssenberg:
Committed as bc27b6db6b294d961ad51616a2b304558b8ff5ca

PR #161, opened 2017-07-19
Add navigation.home for export templates

rolfkleef:
Make navigation.home available for export templates.
Also, add a remark to the Templates help page that navigation.up is not yet implemented.


jaap-karssenberg:
Committed as 0ad2197853de3420f393dfecd9ac2958686ebfa9

PR #176, opened 2017-08-03
Use buffer_initialize for gtkspellcheck>=4.0.3 - fixes #168

### LINKED ISSUE #168: Memory leak in spell checker plugin leads to huge memory usage
Without further ado here is a typical graph of memory usage

RAM usage stays constant when zim is minimized. Apparently new RAM is allocated during interaction with zim, though I was not able to track down specific action that causes leak.

At first I've noticed this behavior with zim 0.65, but considered it a bug specific to my desktop PC. But recently I've upgraded my laptop to Ubuntu 17.04, so it got zim 0.65 too and apparently it has same problem.

Versions 0.66 and 0.67 leaks too.

Maximum memory usage I got since the beginning of tracking zim memory is 3.8 gigabytes. This leak is severe!

Probably my primary notebook size affects leak rate. It is pretty big:
```
$ find "$NOTEBOOK" -name '*.txt' | wc -l
1568
```

Self-Perfection:
Also test for spell checker is expanded to check for memory leak.

For fixed plugin memory usage difference after stress test is about
600-800 KiB. On the other hand unfixed plugin leaks 90 MB or 210
depending on system on which I run test. Therefore 10 MiB is chosen as
safe intermediate threshold to catch leaks but stay silent without leak.

### DIFF
--- a/tests/spell.py
+++ b/tests/spell.py
@@ -48,13 +48,14 @@ class TestSpell(object):
 			ext.toggle_spellcheck()
 			ext.toggle_spellcheck()
 
-			ui.open_page(Path('Foo'))
-			ui.open_page(Path('Bar'))
-			ext.toggle_spellcheck()
-
-			ui.open_page(Path('Foo'))
-			ui.open_page(Path('Bar'))
-			ext.toggle_spellcheck()
+			import resource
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
Committed refactored version as d67726c7124fc371ea6828440f76c6a896bf2361

Issue #163, opened 2017-07-20
Retry button when failing to save page


lrq3000:
I'm trying to use Zim Notes on a shared disk, like SpiderOak Hive.

Often, a synchronization issue will just delete the latest version of the file and replace it with an older one. I am not sure what is causing this, I will try to contact SpiderOak about this.

Anyway, Zim Notes detect this type of issue and offers me to save a copy on my computer, this is SUPER useful and robust.

However, it is very cumbersome to have to compare everytime the changes and recommit myself, is it possible to add a retry button (which would simply retry the saving process, and re-display the same error dialog if the issue happens on retry)? See the attached screenshot for the proposition. I tried to do it myself but I know nothing about GTK :-(

![retry](https://user-images.githubusercontent.com/1118942/28439805-d6a895dc-6da3-11e7-990d-88301e7016d0.png)

Thanks a lot for the great software, it changed my life!

lrq3000:
Also, maybe you could add in the help message that "closing the window allows to continue working on the page in memory", this is a very helpful trick.

lrq3000:
Also to add more info, when I close the window, continue to write down, and then try to save again (so I think it simulates a "retry" option), it still fails to save! How is that possible? I thought Zim Notes was just copying the in-memory page from the temporary storage space to the final destination when saving, to do an atomic save, so how can it repeatedly fail?

lrq3000:
I think this might be related to #135

jaap-karssenberg:
What happens here is that when zim writes the page it checks the timestamp of the file, if that changed since the last read/write it also checks the content (by md5 sum). If the content changed, zim refuses to overwrite the file, as it may contain changes made somewhere else that you don't want to loose inadvertently.  Retry gives the same error since nothing changed (retry only makes sense for temporary file system mallfunction).

Thinking out loud how to improve the situation. Probable we should have a dialog specifically for this situation and keep the current dialog for other malfunctions.  The dialog should  explain the situation, show attributes of both files, probably allow comparison, and offer an "overwrite" option. (Maybe instead of real overwrite, move the old file out of the way so it can be recovered.) Ideal would also be a "merge" option, but no clue how to do that reliable without falling back to an underlying version control system.

Important detail for the case here is that if the new file is older, probably the timestamp is also older, this could be highlighted in the dialog: "Seems like the file you are editing has been replaced with an older version, what would you like to do?"

[...]

PR #178, opened 2017-08-06
Initial implementation of monitoring for external page changes, autom…

joukhaha:
…atic merging of the changes and a dialogue to choose the method of merging

Hi,
As discussed in issue #163 I made an initial implementation to merge externally modified files and buffer to a) enable multiple people working/viewing the same page  more or less at same time (but not quite real time) (my goal) and b) to enable discarding an externally saved file version (Irq3000's goal).

Based on the testing I've done this far, the implementation seems to work more or less as intended, but I've already ironed out several bugs and I'm sure others still exist. Not really being very familiar with zim's architecture etc, I'm still unsure whether my choices in the implementation are acceptable, and whether it'll be possible to add this to the official zim at some point.

I've basically expanded the presently used SavePageErrorDialog to also give options to discard the buffer or the file, or to merge them (once or from there on). If automatic merging is enabled the page will then be merged to include the changes on disk automatically. The page files are actively monitored and also polled for changes.

Please let me what you think, and if you have any suggestions! I'll be happy to provide additional details if needed.


lrq3000:
Thanks a LOT Jouhaka! That's an amazing work!

Sorry for coming on this PR so late, I did not have the time to test it before.

I tried and it would be perfect for my needs too, if the bugs are ironed out indeed :-)

I noticed three major bugs:
* the "Discard Disk" button does not work on my computer.
* the detection of simultaneous changes always forget the latest changes in "Buffer", see the picture below:
sans titre
* finally, merging does not always work correctly (at least merge once, i did not try automerge, because I don't know if it will stay activated forever or just until next time I relaunch Zim Notes, which would be cool). Indeed, in the above example, I did "merge once", and I got only the Disk changes ("zzzzzz" and not "aaaaa").

lrq3000:
And I think that a "merge with external tool" button would be useful for non trivial changes, such as modifications on the same line. The proposed diffs.py seems to only be able to merge per line, not per character (but I think it is a wiser choice to do per line, this ensures the most coherency), but for people working on the same document and paragraphs, an external manual diff tool would be useful I think (it would allow to solve any type of conflict, this is kind of the last resort option apart from saving a copy and doing the diff manually).

zim/gui/pageview.py
```diff
@@ -4863,22 +4897,61 @@ def discard(self):
 			pageview.set_page(page)
 			self._done = True
 
+		def discard_disk(self):  
+                        logger.debug('allowing to overwrite the file')
+			text, page._last_etag = page.source_file.read_with_etag()
+                        page._store_last_lines(page.format.Dumper().dump(self.get_parsetree(),file_output=True))
```
lrq3000: Found the culprit for the "discard disk" button not working: change this line to:

    page._store_last_lines(page.format.Dumper().dump(page.get_parsetree(), file_output=True))

lrq3000: Also could you add a few comments to explain what each line does? The first line seems to be the one that keeps the buffer and discards the disk, so why do we need the other lines of code?

lrq3000:
I am testing this in production (back to working with Zim Notes everyday, feels great! :-D), I will let you know if anything else pops up, but so far it works great!

joukhaha:
Nice to hear this might have some use for others as well! I'll try to look at the problems you identified at some point (might take a while though), and better comment the code I added as well.

We have also been using this regularly since summer and we haven't noticed any major issues, except that the approach I took does not and probably cannot be made to work reliably for pages that include tables (or other multiline objects).

Regarding the bugs (did not check any of them yet, will do it later):

1. 'Discard to Disk' not working - Fully possible, and I'll need to look into that. I think it was at least working at some point of the process, but it sounds like you solved this already.

2. 'detection of simultaneous changes always forget the latest changes in "Buffer"'. Could you describe how to reproduce this in a bit more detail? If you mean that you simultaneously add the aaaaaaaa through the zim page and zzzzzzzz externally, then aaaaaaaaa already is in the buffer (i.e. buffer means the content of the zim page at the time conflicts are detected) and the diff works correctly when not showing it. In other words, the diff shows the changes to be made to present contents of the page, not the common ancestor as in normal 3-way merge (this allows to update the page using zim's own update functions which also enables at least limited undo). A quite cool thing is also that you can continue editing the disk, and the diff will continuously update (or at least it should - this has been tested very little :)

3. 'merging does not always work correctly'. Sounds like this may be related to the previous issue? In principle there's no reason not to activate automerge as it is page specific and closing a page/opening another page will always reset it (if I remember right already reloading the page (ctrl-R) does it).

Which OS are you using? Windows? I only occationally use windows computers through rdesktop or similar, so I haven't tested that ~at all personally. Some of my colleagues do use it though, but they haven't reported any serious issues.

lrq3000:
Hello @joukhaha , thank you very much for following this up :-D

Yes I am using Windows 7 64-bits, but I don't think these issues are due to the OS.

After a few weeks of intense usage, this patch (with the fix mentioned above) works very well! I only use the "Discard to Disk", but it totally fulfills what I needed :-) Thanks a LOT @joukhaha! The diff is also very clear, one just needs to get a bit used to the format (with colors highlighting it would be better but this is only cosmetic, it works already so well!).

I will try to reproduce 2 & 3 in the following days, I'll keep you updated :-)

lrq3000:
After more months of daily usage, I cannot reproduce 2 & 3, everything is running very fine!

This patch is awesome, I love how it allows to quickly fix merging & syncing issues :-) This should be merged in the main branch, I'm pretty sure a lot of users will find it useful (for me it's a killer feature!).

joukhaha:
Thanks, that's nice to hear! We've also been using the modified version quite actively without any issues (albeit overlapping modifications are eventually encountered relatively rarely in our use).

As for trying to merge with the main branch, there are a few issues I'd like to have feedback from Jaap/more knowledgeable people before using more time to refine the code...

First, as mentioned previously I'm unsure how many architecturally/python-wise idiotic choices I've made in making the patch, so some comments and pointers on how to improve the quality and meet the general requirements for the code base would be needed. 

Second, in order to merge with the main branch, it'd probably be necessary to somehow gracefully handle tables and other items (if other such items exist) without one-to-one correspondence between the buffer and text-file line numbers. Getting to the bottom of this without pointers from Jaap or others would likely take more time than I have available for now, at least if there are no easy shortcuts (like disabling the merge for pages with tables).

Third, I guess that the possibility to refine this further will require at least a green light for eventual merging from Jaap. Without it there's (at least presently) not much reason to work on the previous points.

lrq3000:
@joukhaha hehe don't worry I know :-) My message was more targeted at @jaap-karssenberg since he's the project author and maintainer :-)

I think the most important thing is that a PR functionally works correctly, and here that's the case, we both tested it on the field for months without any issue. The tracking of exernal changes to zim notes, even when looking at another note, is very reliable. However, it's true I do not use much the automatic merge either, so I will try to test that out in my test environment.

Code-wise, I guess the best would be to turn this PR into a plugin, but I'm not sure Zim plugins have the ability to touch to core functionalities as this PR needs to.

About the auto-merge of tables, I think the current way is OK: the merge is done line-by-line, so in the case of a table it would merge row by row. Unless two users are modifying multiple cells on the same row at the same time, it's unlikely that an automerge of a table would result in an issue, or only minor (spurious zim markup due to the saving of an unfinished row markup that the user was in the process of typing - except if using the table plugin, then there would be no such issue as the table is always correctly structured).

For the other objects, as far as I saw they are all similar to how tables are handled: specific markup text, usually row-by-row, but no specific object, such that any merging based on text, as your PR does, should normally be fine :-) The Zim approach to formatting makes it easier to handle merging!

joukhaha:
Actually, the case of tables seems to be a bit more complicated than that, at least if the table editor plugin is in use. Then the GUI treats the table as a an element which only occupies a single line in the buffer and this messes up the line numbering on which the merge is based on. Ideally, a fairly simple way to solve this might be to temporarily return the buffer to pure text mode (ie automatically disable the table editor before merge and re-enable it afterwards). If I remember right, however, this would affect e.g. the undo capabilities and possibly also some other functions. Anyway, I didn't really investigate this in that great detail. Also equation editor and other such elements could very well have issues and to my knowledge remain mostly or entirely untested. Getting these right would probably call for additional testing and a broader vision on how things are tied together.

Plugin realization would be an ideal approach in principle, but (not actually knowing how the plugins work) I also suspect it might not be possible to realize this as a plugin.

lrq3000:
Ah good catch then, I didn't look at how complex UI objects were managed internally, I thought Zim worked on pure text all the time. In this case indeed I agree that in case of conflicts, Zim should try to turn everything to pure text before the comparison and merge, I think it should fix most issues (but additional calls might be necessary for specific objects such as equations as you point out, the equation should be stored in plain text in any case but the image might be lost, in this case Zim should regenerate the image).
In any case, it should be possible to fix most of these issues, automatic conflict management in Zim is definitely possible using this PR as a basis :-D

lrq3000:
Hey guys, I've discovered a limitation of this amazing patch that I use pretty much everyday: it monitors reliably only the page you're currently opening.

This means that there are two possible shortcomings, where an external change won't be caught:
* modifying a note, closing Zim and then externally revert the note to a previous state. This is not so bad, I mean it's to be expected that if you close the application then it cannot watch anymore the changes...
* modifying a note, then open another note. In this case, the behavior seems inconsistent: sometimes, the external changes will be caught, sometimes not. This means that if you multitask a lot (ie, writing on multiple different notes), you might lose some changes.

The latter case is for me the most worrying. If I understood the code correctly, normally there should be a monitoring thread maintained for all notes opened so far in the current Zim session (ie, since the last time Zim program was launched), is there maybe a timeout mechanism that maybe unloads too fast the notes from the monitoring thread?

lrq3000:
And thank you very much @nouser for updating the patch! :D

lrq3000:
A small feedback after using this patch extensively: make sure to keep the page you modify opened for a few minutes before closing. If you do modifications, then close Zim notes or switch to another page, then your modifications can be lost (because it seems that only the active page is monitored for changes). It would be nice to keep the monitoring active for all pages opened in the current session, until Zim is closed.

lrq3000:
Still using it on a daily basis, still very happy ;-) @joukhaha ported it to the latest branch of Zim in #688 :D

InstanceOfMichael:
is this merged into 0.74.3? If so, how do you activate it? I was assuming it would auto-magically work or offer the merge button, but it looks like the older behavior.
Screenshot at 2022-03-14 13-29-28

joukhaha:
Yep, does not work with current version... However, I recently realized that the functionality can be relatively easily converted into a plugin. I already made a test version that is missing the dialogues, and just does the 3-way merge without asking the user anything. On a very quick test this seems to work, and adding the options to merge automatically or to show the diff before merging should also be relatively straightforward (although I need to figure out how to use the dialogues correctly).

I **hope** to have a running version ready in a couple of days/weeks. That would certainly mean that the old version should be completely discarded and the plugin used instead.

lrq3000:
Not implemented yet since the transition to gtk3, that's why I am still using the old version, and it still works wonderfully well. -------- Message original --------De : InstanceOfMichael ***@***.***>Date : lun. 14 mars 2022 à 18:32À : zim-desktop-wiki/zim-desktop-wiki ***@***.***>Cc : Stephen Karl Larroque ***@***.***>, Comment ***@***.***>Objet : Re: [zim-desktop-wiki/zim-desktop-wiki] Initial implementation of monitoring for external page changes, autom… (#178)
is this merged into 0.74.3? If so, how do you activate it? I was assuming it would auto-magically work or offer the merge button, but it looks like the older behavior.

Issue #162, opened 2017-07-20
Customizable font also for code/preformatted text


tuxmobil:
I'd love to also customize the font for code or preformatted text. It seems this is currently only possible for regular text. (Linux)

jaap-karssenberg:
Have you seen the "styles.conf" configuration file?  If not, you can find details in the manual. Does that address your question?

tuxmobil:
Yes, I did see that file and that I'm referring too. According to your documentation (and application behaviour), the "font" property is available only for the [TextView] section. I would like to apply a different font to the [Tag code] and [Tag pre] sections, but the font property is without effect on these.

jaap-karssenberg:
There is the "family" attribute that can be used to set a font name. 

The "font" attribute is not exposed for individual tags, but we can change that with a patch.

If you are running from source, you could open up `zim/gui/pageview.py` and modify the definition with "tag_attributes" around line 592 like this: https://github.com/jaap-karssenberg/zim-desktop-wiki/pull/209/files

tuxmobil:
To my understanding you can set a family only (monospace, sans) with the family attribute, but not a specific font like "Arial 10" for example. I'm not running from sources but thanks for the pointer.

PR #209, opened 2017-09-21
Define "font" attribute for text styles

jaap-karssenberg:
Proposed fix for issue #162


jaap-karssenberg:
Committed as b5b2436ee678e66315b442e7bc08766ff43dd52b

PR #232, opened 2017-11-21
Fixed the title in rst default template

spollard:
The overlines and underlines for page titles were not long enough sometimes when exporting in reStructuredText format. I made the overline and underline lengths depend on the title length by updating the Default.rst template.


jaap-karssenberg:
Will merge in "develop"

PR #243, opened 2017-11-29
Always sort notebooks in Open dialog

xHire:
I use couple of notebooks (> 10) in zim and they are always in random order in the Open dialog, which is confusing and bothersome. This patch makes sure the list there is always sorted.

### DIFF
--- a/zim/notebook/info.py
+++ b/zim/notebook/info.py
@@ -344,6 +344,8 @@ class NotebookInfoList(list):
 				info = NotebookInfo(uri)
 			self.append(info)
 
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
The idea is that the list is re-orderable, so you can put the most often used notebooks on top. The suggested patch would break that functionality.

Maybe instead you could add an option to the menu to sort the list ?  Could fit e.g. in context menu popup.

xHire:
Ah, I never noticed that there is this feature. You’re right that my patch doesn’t fit into this at all. I’ll try optional one-time on demand sort.

Issue #15, opened 2017-05-02
Update versioncontrol plugin to play nice with git


jaap-karssenberg:
Current version control plugin does not handle well the concept of the index used in git.

For example preview of changes does not take into account that files will be added before commit - should use commit dry run instead.

Refactor class with git logic to support operations needed more directly. Keep current classes for other systems to avoid updating all at once.

PR #250, opened 2017-12-02
versioncontrol improvements for git staging

### LINKED ISSUE #249: versioncontrol: zim is too aggressive in updating the git index
If a zim notebook is part of a larger git repo, it can be a hassle to have zim updating the index while you are working on things for a commit. I think it would be better if instead of updating it when pages are changed, it simply did a `git add $NOTEBOOK_ROOT` before committing with the "save a version" menu item. This has the following benefits:
* the user has full control over the git index in larger repos
* you don't need to use `-a` when committing (because of the `git add`), which will avoid capturing unrelated repo changes in a zim commit
* there is no change to the workflow for someone whose repo only contains a zim notebook

raincomplex:
Change stage() to mean "prepare the repo for a commit". It is called by SaveVersionDialog before getting the status of the repo. It is only implemented for git, where it does a 'git add' on the whole notebook, so that the status message accurately reflects what will be committed.

Remove '-a' from 'git commit'. Both stage() and VCSBackend.commit() call add() before commit(). Having '-a' here might capture unwanted changes in a zim commit, if the notebook is part of larger repo.

Remove use_staging. It was only used with git and the new way of staging means it isn't necessary.

Fixes #249

### DIFF
--- a/zim/plugins/versioncontrol/__init__.py
+++ b/zim/plugins/versioncontrol/__init__.py
@@ -105,11 +105,6 @@ class NotebookExtension(ObjectExtension):
 	def detect_vcs(self):
 		dir = self._get_notebook_dir()
 		self.vcs = VCS.detect_in_folder(dir)
-		if self.vcs and self.vcs.use_staging:
-			# HACK - FIXME use proper FS signals here
-			# git requires changes to be added to staging, bzr does not
-			# so add a hook for when page is written, to update staging.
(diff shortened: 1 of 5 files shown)


Osndok:
Assuming it's all good, I think this would also address:
* #15 
* https://bugs.launchpad.net/zim/+bug/1633134

jaap-karssenberg:
Patch makes sense, will merge.

One defect is the test case should also be updated, will do that myself on merge.

PR #251, opened 2017-12-02
When removing a symlinked directory, make sure it's empty.

### LINKED ISSUE #238: symlink to notebook is destroyed with Zim 0.67
I have my actual Notebook directory located outsize of ~/Notebooks, and I access it via a symlink.

i.e.:  
~/Notebooks/Work -> ~/.somedirectory/Work

I do this so my 'Work' directory can be part of a larger git repo.

This works fine in Zim 0.65, but when I upgrade to Zim 0.67 the link is occasionally wiped out while Zim is running, and a new empty ~/Notebooks/Work directory is created.  This normally happens to me when moving pages between directories in my notebook.

raincomplex:
Fixes #238

### DIFF
--- a/zim/newfs/local.py
+++ b/zim/newfs/local.py
@@ -34,7 +34,7 @@ def _os_lrmdir(path):
 	try:
 		os.rmdir(path)
 	except OSError:
-		if os.path.islink(path) and os.path.isdir(path):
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
Patch makes sense, will merge.

One defect is the test case should also be updated, will do that myself on merge.

PR #275, opened 2017-12-29
pathbar "Copy Location" contextual menu action

Osndok:


jaap-karssenberg:
Needs update for latest version

PR #278, opened 2017-12-31
Updated link to DokuWiki in manual

Self-Perfection:


Self-Perfection:
Fixes link that leads to nonexistent page.

Issue #282, opened 2018-01-01
Не работает сочетание клавиш "Ctrl +C" 


it19862:
Не работает сочетание клавиш "Ctrl +C"

banan314:
For those who don't know :ru:. If you run the application from terminal, hitting `Ctrl+C` doesn't work (which means nothing happens). :link: [The issue is described on StackOverflow](https://stackoverflow.com/questions/16410852/keyboard-interrupt-with-with-python-gtk).

Please see my PR #288. It solves the problem on Ubuntu, as after running `./zim.py` I can terminate using `^C`.

jaap-karssenberg:
Just tested here, but even without your patch `^C` works fine for me.

banan314:
I just cloned the repository and wrote 
```
./zim.py
```
No window appeared and `^C` didn't work. Then I wrote `./setup.py build`. Afterward `^C` worked.

But today I cannot reproduce it as well, even checking out the commit it was the day I opened the issue.

sojusnik:
Is this somehow connected to https://github.com/jaap-karssenberg/zim-desktop-wiki/issues/81?

genodeftest:
@sojusnik : #81 is probably not related as I see the issue with a different (latin) keyboard layout. Also, #81 is about GUI shortcuts whereas this happens on the console.

sojusnik:
@genodeftest Thanks for pointing that out.

Sandman44:
That simple:
Edit - Preferences - Editing - Default format for copying text to the clopboard  -change from Text to Wiki!

jaap-karssenberg:
This discussion is confusing "^C" in the application for copy-paste and "^C" in the terminal for killing the process.

Current master branch both work correctly.

PR #288, opened 2018-01-06
fix issue #282

banan314:
added signal handler for `Ctrl+C`

### DIFF
--- a/zim.py
+++ b/zim.py
@@ -53,7 +53,10 @@ except ImportError:
 try:
 	encoding = sys.getfilesystemencoding() # not 100% sure this is correct
 	argv = [arg.decode(encoding) for arg in sys.argv]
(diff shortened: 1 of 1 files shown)


zim.py
```diff
@@ -53,7 +53,10 @@
 try:
 	encoding = sys.getfilesystemencoding() # not 100% sure this is correct
 	argv = [arg.decode(encoding) for arg in sys.argv]
-	exitcode = zim.main.main(*argv)
+	if __name__ == '__main__':
```
jaap-karssenberg: What is the purpose of the `__name__ ==  '__main__'` wrapper here ?

This file is always run as the main program.

jaap-karssenberg:
Current master branch works correctly for ^C

PR #307, opened 2018-01-31
Change metainfo ID to 'org.zim_wiki.Zim'

cheese:
Hyphens are discouraged to be used with respect to recent version of AppStream,
and 'org.zim-wiki.Zim' failed to be a valid Flatpak ID.


jaap-karssenberg:
For future reference: changing "-" by "_" in the metainfo id is according to the recommendation in the spec.

See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

PR #312, opened 2018-02-09
Made "Looks like you found a bug" window resizable

mystyc:
Made the "Looks like you found a bug" dialog window resizable.
The default size is still the same because the default values also sort of serve as the minimum size that the window can be resized to.


jaap-karssenberg:
See f40e9aa1d5ef067411af152da34b0aed2bca94ea

PR #356, opened 2018-02-24
Reduce status bar height; unbreak mnemonics in context menu

### LINKED ISSUE #342: Gtk3: improve style of statusbar in mainwindow
The statusbar at the bottom of the mainwindow used to be rendered small, but now takes full size + a border. At least remove the border, check what else can be done to reduce the size a bit.

genodeftest:
By making the largest widget in Zim's status bar smaller, it gets as tiny as
with Gtk+ 2.x.

Fixes #342.


jaap-karssenberg:
Rebased and small touch up but result is merged

PR #367, opened 2018-03-02
search: add middle click to insert link at current cursor

WillForan:
option to insert link from search dialog with middle mouse click


jaap-karssenberg:
I understand what you are trying to do, but implementing on middle click seems kind of arbitrary. At least will not be intuitive for other users.
Open to other suggestions how to integrate such feature with a user interface that is more clear.

PR #377, opened 2018-03-09
Screenshot Plugin: Support for KDE Spectactle and additional configuration options

TobiasWinchen:
Hi,
I added some functionality to the screenshot plugin:
* Added KDE spectacle to choices of screenshot tools. The other tools produce flickering when used together with vnc connections using krdc
* Added a toolbar icon to make screenshots
* Added configuration options to suppress the dialog in favor of default selections


jaap-karssenberg:
Afraid patch needs to be regenerated against latest master, too many conflicts

jaap-karssenberg:
Afraid the change set is not clean, seeing unrelated changes as well

jaap-karssenberg:
Unfortunately this pull request cannot be merged as is. Without further activity or new insight it will be closed in a couple of months. See #1610 for background on cleanup.

Issue #385, opened 2018-03-16
Gtk3: Drag&Drop of pages fails


genodeftest:
Steps to reproduce:
1. have more than one page in Zim
2. try to drag&drop one page into another one (left sidebar)

What happens:
drag&drop does not happen. Traceback:

```
__init__.py:495:do_drag_data_get:AttributeError: 'SelectionData' object has no attribute 'target'

Traceback (most recent call last):
  File "/path/to/zim/plugins/pageindex/__init__.py", line 495, in do_drag_data_get
    assert selectiondata.target == INTERNAL_PAGELIST_TARGET_NAME
AttributeError: 'SelectionData' object has no attribute 'target'

Local variables in innermost frame:
time: 0
info: 2
selectiondata: <Gtk.SelectionData object at 0x7fcc18324828 (GtkSelectionData at 0x5560f8512b00)>
dragcontext: <__gi__.GdkWaylandDragContext object at 0x7fcc1833de10 (GdkWaylandDragContext at 0x5560f8192830)>
self: <pageindex.PageTreeView object at 0x7fcc183140d8 (zim+plugins+pageindex+PageTreeView at 0x5560f81d83e0)>
```

Version:
Latest git revision from `next` branch as of now.

genodeftest:
I created a fix (valid and correct in terms of Gtk+/GObject documentation), but it still fails due to a bug in GObject python bindings: https://gitlab.gnome.org/GNOME/pygobject/issues/104 and https://gitlab.gnome.org/GNOME/pygobject/issues/90

genodeftest:
https://gitlab.gnome.org/GNOME/pygobject/issues/104 has a workaround which can be found at https://github.com/virtuald/pygi-treeview-dnd. See also: exaile/exaile#200

jaap-karssenberg:
Believe this issue is closed, so issue #390 for blocking issue on dnd

swestlake:
version 0.71.1 has the same regressive bug in that attempting to moving nodes in the index pane does not perform the move of the topic page to another location.

libgtk-3-0:amd64    3.22.11-1
python3-gi              3.22.0-2 

debian linux 64-bit..

Issue #332, opened 2018-02-20
Gtk3: need testing of rich copy-paste *to* zim


jaap-karssenberg:
As reported in bug #326 rich copy paste *from* zim is broken. Rich copy paste *to* zim should still work, but cannot be tested automatically.

Need (manual) testing for pasting into a page
- images
- links
- text

Automated tests using other tools to set the clipboard are welcome as well (and may provide clues to solving #326 as well ...)

muguu:
Test for pasting an image into a page: FAIL
`DEBUG: Action: paste
DEBUG: Targets available for paste: (True, targets=[Gdk.Atom.intern("TIMESTAMP", False), Gdk.Atom.intern("TARGETS", False), Gdk.Atom.intern("MULTIPLE", False), Gdk.Atom.intern("SAVE_TARGETS", False), Gdk.Atom.intern("text/html", False), Gdk.Atom.intern("text/_moz_htmlinfo", False), Gdk.Atom.intern("text/_moz_htmlcontext", False), Gdk.Atom.intern("image/png", False), Gdk.Atom.intern("image/bmp", False), Gdk.Atom.intern("image/x-bmp", False), Gdk.Atom.intern("image/x-MS-bmp", False), Gdk.Atom.intern("image/x-icon", False), Gdk.Atom.intern("image/x-ico", False), Gdk.Atom.intern("image/x-win-bitmap", False), Gdk.Atom.intern("image/vnd.microsoft.icon", False), Gdk.Atom.intern("application/ico", False), Gdk.Atom.intern("image/ico", False), Gdk.Atom.intern("image/icon", False), Gdk.Atom.intern("text/ico", False), Gdk.Atom.intern("image/jpeg", False), Gdk.Atom.intern("image/tiff", False)]), we want parsetree
WARNING: Could not paste - no compatible data types on clipboard`


(Same test with master was successful): 
`DEBUG: Requesting data for image/png
DEBUG: Saving image from clipboard to /home/murat/Notebooks/Notes/Home/pasted_image.png
DEBUG: Pasting link: file:///home/murat/Notebooks/Notes/Home/pasted_image.png (type: file, isimage: True)`


Complete Debug:
```
INFO: This is zim 0.68-gtk3
DEBUG: Python version is sys.version_info(major=3, minor=6, micro=3, releaselevel='final', serial=0)
DEBUG: Platform is posix
DEBUG: No bzr version-info found
[...]

PR #386, opened 2018-03-17
Plugin PageIndex: Partial fix to DnD issues

genodeftest:
See #385 for more details.


genodeftest:
I'll give this another try also fixing #332.

PR #388, opened 2018-03-17
ImageView widget: Adapt to new API and fix removing timeout

### LINKED ISSUE #387: Gtk3: ImageView widget uses old API which has been removed
Steps to reproduce:
1. Enable plugin "Equation editor"
2. open any Zim page
3. Insert an equation
4. in the dialog, type any valid TeX "code"
5. press the `Preview` button

What happens:
Nothing visible. Traceback:
```
  File "/path/to/zim/plugins/base/imagegenerator.py", line 354, in preview
    self.imageview.set_file(self.imagefile) # if None sets broken image
  File "/path/to/zim/gui/widgets.py", line 4096, in set_file
    self.set_pixbuf(pixbuf)
  File "/path/to/zim/gui/widgets.py", line 4107, in set_pixbuf
    self._render()
  File "/path/to/zim/gui/widgets.py", line 4129, in _render
    allocation = self.allocation
AttributeError: 'ImageView' object has no attribute 'allocation'
```

What should happen:
Show equation

Additional info:
Having a look in `/tmp/zim-[username]/`, the equation is parsed and rendered correctly, i.e. there is a valid `.dvi` and a valid `.png` file. Displaying it fails though.

The way drawing works has changed quite a lot between Gtk+ version 2 and 3 ([documentation](https://developer.gnome.org/gtk3/stable/gtk-migrating-2-to-3.html)). There's probably no need to use custom drawing any more.

genodeftest:
Fixes a stream of warnings about a GObject timeout source which has already been removed.

Fixes #387.

### DIFF
--- a/zim/gui/widgets.py
+++ b/zim/gui/widgets.py
@@ -4112,6 +4112,7 @@ class ImageView(Gtk.Layout):
 		# remove timer if any
 		if self._render_timeout:
 			GObject.source_remove(self._render_timeout)
+			self._render_timeout = None
 
 		if not self._pixbuf \
 		or (allocation.width, allocation.height) == self._render_size:
@@ -4124,6 +4125,7 @@ class ImageView(Gtk.Layout):
 		# remove timer if any
 		if self._render_timeout:
 			GObject.source_remove(self._render_timeout)
+			self._render_timeout = None
 
 		# Determine what size we want to render the image
 		wwin = self.get_property('width')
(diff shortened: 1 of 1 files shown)


genodeftest:
Note: The equation editor is still broken visually, but only in equation editor plugin code.

jaap-karssenberg:
Afraid this patch doesn't work - "width" and "height" properties reflect size, not allocation. Preparing alternative patch using "get_allocation()". But it helped me to figure out what is going on.

Thanks for putting me on the right track solving this one !

genodeftest:
Oh, thanks, I didn't know that!

genodeftest:
You might still want to cherry-pick the second commit though, it fixes a different issue.

genodeftest:
> You might still want to cherry-pick the second commit though, it fixes a different issue.

Oh, I've seen you did that in 3127afe87a7c73a622fedce7328d0a5d677655ff already.

Issue #390, opened 2018-03-18
Gtk3: Gtk.SelectionData.set() is broken in python bindings


genodeftest:
`Gtk.SelectionData.set()` cannot set selection data for dragging within file browser or attachment browser. See also: #385
Upstream bug reports:
* https://gitlab.gnome.org/GNOME/pygobject/issues/104 has a workaround which can be found at https://github.com/virtuald/pygi-treeview-dnd. See also: exaile/exaile#200
* https://gitlab.gnome.org/GNOME/pygobject/issues/90

jaap-karssenberg:
Trying to get this to work using pygi-treeview-dnd but currently blocked by this issue: https://github.com/virtuald/pygi-treeview-dnd/issues/4

virtuald:
Weird, I'll take a look later tonight.

jaap-karssenberg:
Recently it seems this is now working. Keeping bug open for monitoring the situation, but apparently something changes upstream.

Gtk / python-gi versions of working situation:
- libgtk-3-0/bionic,now 3.22.30-1ubuntu1
- python3-gi/bionic,now 3.26.1-2

murattepegoz:
Folks, dragging files from attachment panel into a wiki page is also not working for me. The problem started with version 0.70. I am using zim on mac with homebrew. Let me know if I need to create a new issue for this.
Here is the snapshot summarizing the problem.
![image](https://user-images.githubusercontent.com/3308687/56845551-f017cf80-6877-11e9-944f-b39124534156.png)

jaap-karssenberg:
This is a separate issue please report in a new ticket.

Regards,

Jaap

Op za 27 apr. 2019 08:08 schreef murattepegoz <notifications@github.com>:

jhe-:
The issue is still happening for me.

I'm on:
- Zim 0.71.0
- libgtk-3-0/bionic,now 3.22.30-1ubuntu2
- python3-gi/bionic,now 3.26.1-2

jhe-:
#827 is a hacky but minimal workaround that fixes drag-and-drop in the index pane in my case.

arigit:
Issue still present on 0.71.1, running from source tree, MacOS 
Drag & drop from Finder to a zim page results in nothing. The same action on the same maching on 0.69.1 (also running from source tree) works perfectly.

Running zim in debug mode (python3 zim.py --debug) shows no messages at all when dropping the file.

[...]

Issue #326, opened 2018-02-20
Gtk3: rich copy-paste from zim to clipboard broken with port to Gtk3


jaap-karssenberg:
Zim's clipboard function uses the gtk method [`Gtk.Clipboard.set_with_data()`](https://developer.gnome.org/gtk3/stable/gtk3-Clipboards.html#gtk-clipboard-set-with-data). This method was supported by the "pygtk" bindings, but is not supported in "gir" bindings which are used for the Gtk3/python3 port. The reason is that the method signature seems incompatible with the "gir" annotation.

As a result, the copying of rich content **from** zim **to** the clipboard is broken. This affects also internal copy-paste of e.g. page links. (Copying into zim should work, but tests are broken due to a lack of a way to set the clipboard ...)

A potential work around would be to use `Gtk.TextBuffer` serialization API. However, this is broken as well, see [Gnome bug 791443](https://bugzilla.gnome.org/show_bug.cgi?id=791443.) Also this serialization API is due to be removed in Gtk4, so would only be a temporary workaround.

So both methods used setting any other content on the clipboard other than plain text (or images) are broken.

For internal copy paste we can probable think of workarounds, but pasting  e.g. HTML formatted text outside of zim can not be done with the available API.

What is needed to fix this is a way to directly access the `Gtk.Clipboard.set_with_data()` method from python. Potentially the ctypes library could be of use here, further research is needed.

genodeftest:
Having `Gtk.Clipboard.set_with_data()` is blocked due to this Gtk+ bug: https://bugzilla.gnome.org/show_bug.cgi?id=656312

genodeftest:
See also: #390

anxiousmodernman:
Am I reading this right: that they _won't_ be adding this method to GTK3, but rather 4?

jaap-karssenberg:
Attempt to get this to work using ctypes here: https://gist.github.com/jaap-karssenberg/4f7d3dba7d5bc725dacf28a382893d69

Got stuck debugging this - all help is welcome to get this to work!

arigit:
Pointers on ctypes and gnome introspection / to test later

https://stackoverflow.com/a/43212363

arigit:
Some positive results. The original code was failing this assert
`assert text == 'Test 123'`

Disclaimer - I'm just dabbling into this so sorry of the results I got are obvious - really hoping the below actually helps.

[...]

PR #392, opened 2018-03-18
Gtk3: Some fixes for clipboard (copy, paste) and DnD (drag and drop)

genodeftest:
# What works

Clipboard:
* These changes make all the tests in #332 work, i.e. copying from outside into Zim works fine now.

DnD:
* These changes make drag&drop from outside (e.g. gedit, nautilus, firefox) to Zim work.

# What doesn't work
Clipboard:
* Copying from Zim to anywhere (Zim or outside) is still broken. This is due to broken Gtk+ Phthon bindings, see #326 for details.

DnD:
* Drag in Zim, drop outside of Zim. 
* Some situations of drag in Zim, drop in Zim (e.g. #385).
In both cases, the Gtk+ Python bindings are to blame, see #390 for details.

### DIFF
--- a/zim/gui/clipboard.py
+++ b/zim/gui/clipboard.py
@@ -119,12 +119,12 @@ def pack_urilist(links):
 		if is_url_re.match(link):
 			link = url_encode(link, mode=URL_ENCODE_READABLE) # just to be sure
 		text += '%s\r\n' % link
-	return text
+	return text.encode()
 
 
 def unpack_urilist(text):
 	# FIXME be tolerant here for file://path/to/file uris here
-	text = text.strip('\x00') # Found trailing NULL character on windows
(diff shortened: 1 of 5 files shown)


jaap-karssenberg:
Merged as e42ebc605620567cb7ecc16a8551b038d1c36aae

genodeftest:
Thank you!

PS: If you use a different merge option ("Create merge commit"), you won't have to manually close the pull request but it will close automatically. This looks nicer in GitHub because the merged pull requests actually show as being merged, not as being closed.

jaap-karssenberg:
I realize that, but for "small" series of patches I prefer to rebase and
keep the commit in serial order. If anything it helps me to check the diffs
on top of the latest version.

Therefore I only use merge commits for widely varying branches, e.g.
syncing between the gtk2 and the gtk3 branches.

-- Jaap

PR #394, opened 2018-03-19
Fix server crash by non-English page names.

ainoneko:
Fixes https://bugs.launchpad.net/zim/+bug/1698784

(I tested with pages named "English", "French", "Hebrew", "Russian" and "Japanese"  (in the corresponding languages), and found that each one except "English" crashed the server in _resolve_pagename().
After I found that some names are not unicode and fixed it there, only the "Hebrew" page crashed the server — now in makeValidPageName(), because the non-unicode  string was damaged by _pagename_invalid_char_re.sub().
)
zim_server_index

### DIFF
--- a/zim/notebook/index/pages.py
+++ b/zim/notebook/index/pages.py
@@ -424,6 +424,9 @@ class PagesViewInternal(object):
 		pagename = parent
 		page_id = self.get_page_id(parent)
 		for i, basename in enumerate(names):
+			if isinstance(basename, str):
+				logger.debug('basename "%s" was not Unicode in _resolve_pagename' % basename)
+				basename = basename.decode('utf-8')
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
I believe this to be no longer valid on master because we now use python3 and there str is always unicode by default.

You may want to reopen the request targeting the "python2-gtk2" branch

ainoneko:
> You may want to reopen the request targeting the "python2-gtk2" branch

No, I don't want to ^_^: you seem to have fixed it somehow in the 0.69 release.
(At least I tried the current version in the "python2-gtk2" branch with python2, and it didn't crash.)

jaap-karssenberg:
OK, thanks for testing !

PR #399, opened 2018-03-28
Rename AppData file according to spec

Qantas94Heavy:
For desktop AppData files, the file extension ".appdata.xml" should be used instead of ".metainfo.xml".

More info: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent


jaap-karssenberg:
Merged as dffc113d39e309709046eda1c9ee2d76b4cff4c1

PR #402, opened 2018-03-30
Changed "egistry" to "registry"

kennethjhim:
Corrected the "egistry" to "registry" that shows when the app is about to be installed.


jaap-karssenberg:
Fixed in 4abb1da1616c7438326e877b4921e2fb8ced18c6. Should have used this merge but didn't see it in association with the issue report. Sorry

Issue #401, opened 2018-03-30
Customizable wrap mode in Code Block


aaferrari:
It would be very useful if in the Source View plugin options you could configure the lines wrap mode so you don't have to use the horizontal scroll bar when a code has very long lines. I found out that it can be done by adding the function `set_wrap_mode()` to sourceview.py in this way:
```
@@ -294,6 +294,7 @@
 		self.view.set_right_margin_position(80)
 		self.view.set_show_right_margin(True)
 		self.view.set_tab_width(4)
+		self.view.set_wrap_mode(Gtk.WrapMode.WORD_CHAR)
 
 		# simple toolbar
 		#~ bar = gtk.HBox() # FIXME: use gtk.Toolbar stuff
```
Instead of `Gtk.WrapMode.WORD_CHAR` you can also choose between `Gtk.WrapMode.NONE` (wrap lines never), `Gtk.WrapMode.CHAR` (at character boundaries) and `Gtk.WrapMode.WORD` (at word boundaries).

jaap-karssenberg:
Patch merged

PR #403, opened 2018-03-31
Text wrap mode properties

aaferrari:
Checking the source code of a couple of plugins I discovered how to add what I had proposed in #401. As far as I've tried it works smoothly


jaap-karssenberg:
Patch needs to be re-generated for latest version, changes for Gtk3 usage needed

PR #407, opened 2018-04-04
Correct spelling errors

banan314:


banan314:
Only the last commit (f36b212) is important, the first 3 are merges.

tests/actions.py
```diff
@@ -12,15 +12,15 @@
 class TestPrimaryKeyBinding(tests.TestCase):
 
 	def runTest(self):
-		for accel in ( # Ctrl-A or Command-A
```
banan314: I know it's a shortage, but in my opinion `accelerator` is better.

zim/config/dicts.py
```diff
@@ -689,7 +689,7 @@ def setdefault(self, key, default, check=None, allow_empty=False):
 
 class SectionedConfigDict(ControlledDict):
 	'''Dict with multiple sections of config values
-	Sections are auto-vivicated when a non-existing item is retrieved.
```
banan314: there is no such a word

zim/config/dicts.py
```diff
@@ -880,8 +880,8 @@ class HierarchicDict(object):
 	There is a special member dict stored under the key "__defaults__"
 	which has the top-level fallback properties.
 
-	Child dicts are auto-vivicated, so this object only implements
```
banan314: there is no such a word

jaap-karssenberg:
Looks like this closed automatically when I dropped the "next" branch -- could you re-open the request for "master" ?

Thanks,

Jaap

PR #408, opened 2018-04-04
Remove redundancy in regular expressions

banan314:
Redundancy:

- 2 consecutive spaces: `  +`
- unnecessary escapes like `\%` or `[\.\/]` (in brackets it's not needed)

I didn't change the behavior of the regexes.


jaap-karssenberg:
Looks like this closed automatically when I dropped the "next" branch -- could you re-open the request for "master" ?

Thanks,

Jaap

PR #424, opened 2018-04-17
a fix in documentation that helps others when using equation plugin

warlord500:
by including this in the documentation, it makes it easier for users to debug with why inserting an equation wont work.  this is at least partly discussed in this [bazaar bug](https://bugs.launchpad.net/zim/+bug/1237909)


krisztianmukli:
Thanks for the tip! I've updated the Community Wiki Page: https://github.com/jaap-karssenberg/zim-wiki/wiki/Install-the-full-Zim-Desktop-Wiki-Suite-on-Windows

jaap-karssenberg:
No problem in adding this in the documentation, but would like to understand why the "zhmetrics" package is required. It isn't used in our templates as far as I'm aware. So what pulls in this dependency?

lrq3000:
Thank you so much! I never used the equation editor as I never found a way to make it work, I might give it a try again now!

jaap-karssenberg:
Unfortunately this pull request cannot be merged as is. Without further activity or new insight it will be closed in a couple of months. See #1610 for background on cleanup.

PR #465, opened 2018-06-08
Calendar plugin - "Go to Today" boost

e3rd:
having some text selected, it is replaced with a link and moved at the end of today's page


e3rd:
I have to re-write the code for Gtk-3, right?

PR #468, opened 2018-06-12
multiformatting! FIX #305

### LINKED ISSUE #305: Allow combining formatting like italic, bold, underline
In practice, I experience that e.g. italic & bold cannot be mixed, i.e. nested behaviour is not supported.

Version: V0.65 (Ubuntu).

e3rd:
Finally it is possible to apply multiple text styles to a text. Ex: Links can be nested under a stroke text and still be clickable.
Works quite reliable!
This is a big change so I'm ready to your comments / change requests.

### DIFF
--- a/zim/gui/pageview.py
+++ b/zim/gui/pageview.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 
 # Copyright 2008-2017 Jaap Karssenberg <jaap.karssenberg@gmail.com>
 
@@ -122,7 +123,7 @@ KEYVALS_BACKSPACE = list(map(Gdk.keyval_from_name, ('BackSpace',)))
 KEYVALS_TAB = list(map(Gdk.keyval_from_name, ('Tab', 'KP_Tab')))
 KEYVALS_LEFT_TAB = list(map(Gdk.keyval_from_name, ('ISO_Left_Tab',)))
 
-#~ CHARS_END_OF_WORD = (' ', ')', '>', '.', '!', '?')
+# ~ CHARS_END_OF_WORD = (' ', ')', '>', '.', '!', '?')
 CHARS_END_OF_WORD = ('\t', ' ', ')', '>', ';')
 KEYVALS_END_OF_WORD = list(map(
 	Gdk.unicode_to_keyval, list(map(ord, CHARS_END_OF_WORD)))) + KEYVALS_TAB
@@ -489,7 +490,7 @@ class TextBuffer(Gtk.TextBuffer):
 	@signal: C{inserted-tree (start, end, tree, interactive)}:
 	Gives inserted tree after inserting it
 	@signal: C{textstyle-changed (style)}:
-	Emitted when textstyle at the cursor changes
+	Emitted when textstyle at the cursor changes, gets the list of text styles or None.
 	@signal: C{link-clicked ()}:
 	Emitted when a link is clicked; for example within a table cell
 	@signal: C{clear ()}:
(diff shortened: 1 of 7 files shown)


zim/parser.py
```diff
@@ -334,7 +334,7 @@ def _process(self, builder, text):
 		# default action for matched text
 		if self.descent:
 			builder.start(self.tag)
-			self.descent(builder, *text)
+			self.descent(builder, text)
```
e3rd: I haven't found a use case of the `descent` feature in the present code. I think the "*" was here by a mistake.

zim/gui/pageview.py
```diff
@@ -1489,10 +1496,10 @@ def get_textstyle(self):
 		'''
 		tags = filter(_is_style_tag, self._editmode_tags)
 		if tags:
-			assert len(tags) == 1, 'BUG: can not have multiple text styles'
-			return tags[0].get_property('name')[6:] # len('style-') == 6
+			# X not anymore assert len(tags) == 1, 'BUG: can not have multiple text styles'
```
e3rd: I may remove that line if you want.

zim/gui/__init__.py
```diff
@@ -1836,7 +1836,7 @@ def statusbar_element(string, size):
 			return label
 
 		# specify statusbar elements right-to-left
-		self.statusbar_style_label = statusbar_element('<style>', 100)
+		self.statusbar_style_label = statusbar_element('<styles>', 110)
```
e3rd: status bar space was too small... it might be dynamic or so but I just made it a little wider

jaap-karssenberg:
Hi Edvard,

This patch looks promising. Lot of users will be happy with the improvement.

Next steps in my mind should be:
- Please rebase the patch against the "next" branch. This is branch will merge with master soon and has a lot of changes (all the python3 / Gtk3 work)
- Since this is a large change, need to update the test cases to ensure that nested tags work as expected in all export formats
- Document what nesting is allowed and which isn't. Probably need some test cases to check the edge cases.

Regards,

Jaap

e3rd:
Sorry, I overlooked the next branch. I'll try to realize your points in the next few weeks!

e3rd:
> branch

Merging to `next` branch now :).

>  need to update the test cases 

Created four test cases in tests dir. Are they alright?

> Document

Where would you document that? On website/pages or just in CHANGELOG.txt? (I put a longer description in CHANGELOG.)

jaap-karssenberg:
I've added this merge to the "next feature release" milestone. Still pushing myself to get a fully Gtk3 based release working. Tempting as it is, I won't put any new features in to make testing really focus on regressions of the porting. Given that I will be off for the next month, I hope this will happen in August - but these timings are usually overly positive. Anyway, I can merge this one once that is out the door.

For testing what I had in mind is to test also the export formats (html / latex / ...) handle nested tags well. Have a loop at "tests/formats.py" - it uses a set of files in "tests/data/formats/" maybe you can duplicate that test and use a test file that exercises various combinations of nested formats?

Documentation wise, I think this should land somewhere in the manual. Especially the combinations that are allowed, e.g. headings not being allowed in nests. This should be information for users, not developers. Some users may be surprised about what does/doesnt work and look in the manual for guidance.

Regards and many thanks,

Jaap

e3rd:
Ok, I have to add tests for export formats (html / latex / ...) and add it better to the documentation and then re-open this PR again.

I admit I didn't do this work before merging GTK-3 related changes back into master.

Issue #487, opened 2018-07-14
Linking improvement suggestions


igormcampos:
Hello, the link to another page function is the one I most use on this software, so I'm going to list some things that annoys me a little, maybe there's a way to circle these, so please let me know if you have a way to do that.

1. after using a subpage indication (:) the Ctrl+ L (linking pop-up) doesn't fuzzy search anymore, more than that it even becomes case-sensitive. Example:
`
If I want to find a page that is called "The Citadel" inside the page "Maps" I could do the following to link it:
Ctrl + L  and type "cit" and it will pop up at the bottom as a suggestion
or
CTRL + L and type "maps:cit" (but this doesn't work, doesn't show as suggestion or anything, even if you type "maps:the cit"  it won't work
it will only work if you type exactly the title "Maps:The Cit" that will suggest the complete title
`

2. There's no way to check the name of the page if you are linking one with a path too long. If the path is nothing like ":Places:Navigation:Maps:The Citadel" it will probably not show the last part on the sentence and if there's something like "The Bridge" also there it will probably be hard to find yourself, I think this image explains it better:
![image](https://user-images.githubusercontent.com/8935865/42724624-05e793bc-874c-11e8-9fb7-64d8567cebba.png)

3. Text should reflect the title of the page, not the path. Using the prior example it is something like this
It should be like this:
![image](https://user-images.githubusercontent.com/8935865/42724641-5ff43252-874c-11e8-9189-45abe86276cb.png)

And not this:
![image](https://user-images.githubusercontent.com/8935865/42724644-778a2868-874c-11e8-99db-9d9f8365ea54.png)

Having to manually cut that part of the path every time I link something is very tiring.
Thanks for reading.

igormcampos:
@Osndok I see where you are getting with the 3, but I still think it would be a great change overall, or maybe an option to customize that on preferences?

Abou the mass moving, yes I suffered from that once, I suddenly notice some strange pages that shouldn't be there, but I was able to move them because I noticed early on hahaha
My way to make that less annoying is using complete paths as you said, to make that easier I use macros on AHK to the most common pages that I have as "directories" or "groups", I also have a macro to circle 3, that removes all the path from a string, but it was a suggestion anyway hehe

igormcampos:
Not sure if someone would want this, but I'm gonna leave this here if someone is also troubled with the paths and page name, it's the macro I made for AHK, if you have it installed it's pretty easy to setup and I commented what each part does.
[...]

Issue #477, opened 2018-06-30
Moving a tree part of a book note in another one leaves shadow contents


lalebarde:
Starting state:
```
Book Note 1          Book Note 2
a                           x
  aa                         xx
  ab                         xy
b                           y
  ba                         yx
  bb                         yy
```
Move tree _y_ of Note Book 2 into Note Book 1
`mv mybooks/Book Note 2/y mybooks/Book Note 1/`
Reopen zim and re-index both notebooks

Observed End state:
```
Book Note 1          Book Note 2
a                           x
  aa                         xx
  ab                         xy
b                           y
  ba                         yx
  bb                         yy
y
  yx
  yy
```
Tree _y_ is still here in Note Book 2 but _yx_ and _yy_ apear in gray. At least, contents are not here anymore as expected.

Expected End state:
```
Book Note 1          Book Note 2
a                           x
  aa                         xx
  ab                         xy
b                           
  ba                         
  bb                         
y
  yx
  yy
```

jaap-karssenberg:
This means there are still links towards the pages that you moved in the notebook. You can find those by looking at the "backlinks" and remove them.

PR #492, opened 2018-07-15
Do not suggest or create floating links for the user.

Osndok:
Floating links routinely become broken and misdirected during page renames.

Futhermore, the algorithim used to compute a floating link between two
pages has a high runtime cost, and does not scale well with a large
number of pages, particularly repos with many 'deep' pages.

This commit removes Zim's ability to generate floating links, while not
removing general support (both for data-in-place and as the user might
manually enter). Without any particular user interaction, links that
were originally created to be floating will gradually be replaced by
absolute links as their target pages are moved and renamed.

As a result of this commit, users might be happy to find considerably
faster "insert link" and "edit link" dialog box operations (specifically
with the building of the autocompletion suggestions), faster page move
and rename operations, and (in the long term) far fewer broken links.

A downside, to many users, might be a considerable increase in the
number of longer/absolute links. This might be ameliorated by using
the page title or 'n' of the last path segments as the visible link
text.

The generation of links between pages that do no share a parent are
practically not effected (differing only with a preceding colon ':'),
except for the fact that they are no longer vulnerable to the page
movement bug.

This may be related to the following existing tickets:
* https://bugs.launchpad.net/zim/+bug/501496
* https://bugs.launchpad.net/zim/+bug/1086343
* https://bugs.launchpad.net/zim/+bug/1133365
* https://bugs.launchpad.net/zim/+bug/1419531
* #477
* #487

### DIFF
--- a/zim/notebook/index/pages.py
+++ b/zim/notebook/index/pages.py
@@ -683,43 +683,7 @@ class PagesView(IndexView):
 		elif target.ischild(source):
 			return HRef(HREF_REL_RELATIVE, target.relname(source))
 		else:
-			href = self._find_floating_link(source, target)
-			return href or HRef(HREF_REL_ABSOLUTE, target.name)
-
-	def _find_floating_link(self, source, target):
-		# Relative links only resolve for pages that have a common parent
-		# with the source page. So we start finding the common parents and
-		# if that does not resolve (e.g. because same name also occurs on a
-		# lower level) try one level up to "anchor" the link.
-		# It is absolute must to use resolve_link() here - this ensures the
-		# outcome is always consistent between these functions.
-		parentnames = []
-		for n1, n2 in zip(source.parts, target.parts):
-			if n1 == n2:
-				parentnames.append(n1)
-			else:
-				break
-
-		def try_link(names):
-			assert names
-			href = HRef(HREF_REL_FLOATING, ':'.join(names))
-			pid, pagename = self._pages.resolve_link(source, href)
-			return href if pagename == target else None
-
-		relnames = target.parts[len(parentnames):]
-		if not relnames: # Target is direct parent
-			relnames.insert(0, parentnames.pop())
-		href = try_link(relnames)
-		if href is not None:
-			return href
-		else:
-			while parentnames:
-				relnames.insert(0, parentnames.pop())
-				href = try_link(relnames)
-				if href:
-					return href
-			else:
-				return None # no floating link possible
+			return HRef(HREF_REL_ABSOLUTE, target.name)
 
 	def list_recent_changes(self, limit=None, offset=None):
 		assert not (offset and not limit), "Can't use offset without limit"


Osndok:
I say 'floating' b/c that is what it says in the source, and refers to a link that does not contain a ":" or "+" prefix. IIRC, a relative link (strictly speaking) is one with the plus prefix.

myrdd:
@Osndok thank you a lot for this PR, I'm using it already. (Well, thanks Jaap as well for Zim!)

There's one change I'm currently missing in this PR: The change of the »Edit > Copy Location« functionality (`Shift+Ctrl+L`). IMHO it should prepend a colon (`:`) at the beginning when this patch is active. what do you think @Osndok?

Osndok:
@myrdd - I'm glad you find it useful! I'm not exactly sure what you mean, because "copy link" always seems to copy an absolute link for me... it might be new code interacting with the patch b/c I'm a bit behind the master branch ATM.

myrdd:
@Osndok you're right, the info I provided was incomplete. Mainly I forgot to mention `Ctrl+V`. Let me try to explain:

1. create this root-level hierarchy:

```
.
├── A
│   └── a
├── B
```

2. go to `:A:a`
3. `Shift+Ctrl+L`
4. paste (`Ctrl+V`) at `:A:a`
5. paste (`Ctrl+V`) at `:A`
6. paste (`Ctrl+V`) at `:B`
7. paste (`Ctrl+V`) in a different program, e.g. vim

**the results are:**

4. `[[a]]` at `:A:a`
5. `[[+a]]` at `:A`
6. `[[:A:a]]` at `:B`
7. `A:a` in vim

On any Zim page with the same root-level parent you'll get relative paths.

I've tested this on your branch (which is based on 0.68) as well as on a rebase on 0.69.

So IMHO the result of step 4 should be the same as that of step 6.
Step 7 is debatable, maybe preferably like 6 as well.

myrdd:
whoops, still not exactly what I've been doing…

”step 4“ above is this edge case:

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/6fcdd35688edcc0d1f3c68801c6f1f8e8fe47f75/zim/notebook/index/pages.py#L748-L749

well, that's actually really unusual. (still, we could change that line of code for the patch.)

In fact I've been doing this:

1. `Shift+Ctrl+L`
2. `Ctrl+L` on any other page (Insert > Link...)
3. paste the link into the ”Link to:“ field

And there you'll get `A:a` instead of `:A:a`. That's what I was always referring to. :see_no_evil: 

---

BTW, shouldn't ”_don't use floating links_“ rather be an (advanced) *option*?

Osndok:
> [given the above setup] **the results are:**
>     1. `[[a]]` at `:A:a`

Yes, this is a strange edge case. Without html-like fragment links, I really don't know why the user would link to the current page, or what to do in that case. I would assume that a page referencing itself would be the safest kind of floating link, if ever there was one.

>     2. `[[+a]]` at `:A`

I really have not had much issue with these sort of relative links, so... I'm fine with this!

>     3. `[[:A:a]]` at `:B`

Good! This is the intended change. Originally zim would plop a floating link in there ("A:a").

>     4. `A:a` in vim

I can't replicate this. When I paste links out of zim I get a full path with preceding ':'.

>  3. paste the link into the ”Link to:“ field ... And there you'll get ```A:a``` instead of ```:A:a```

Nope. I get ":A:a" here too... there must be something different between our two setups, but I don't know what that would be. :man_shrugging:

> BTW, shouldn't ”don't use floating links“ rather be an (advanced) option?

If floating links where entirely fixed, then maybe... but in general I would say that there should not be an option that is known to cause data loss.

To you and I, ":A:a" page link has a direct analog to "/A/a" in the unix world, and we are **very familiar and comfortable** with absolute paths; in the zim world, it kinda starts with the assumption of a flat pagespace... ("A" & "B") with sub pages as an extension ("A:a")... so I think that part of the struggle here is that all the internals of zim are geared to use "A:a" as the canonical pagename, even though to a unix-centric mind that is an incomplete & relative path).

Well... and that there is no unix equivalent to the progressively-shallowing-resolution (like if you were make a link to "B" in the ":A:a" page), that it might refer to either ":B" or ":A:B", depending on the current state of the notebook... in unix that is generally handled with "../B" style links, and carries it's own set of problems.

myrdd:
> > 3. `[[:A:a]]` at `:B`
> 
> Good! This is the intended change.

I see! Haven't been aware of that. :) (Note to myself: confirmed)

> > 2. `[[+a]]` at `:A`
>
> I really have not had much issue with these sort of relative links, so... I'm fine with this!

Actually I prefer relative links, and I've got some feature ideas, but this isn't the place to discuss them. ;)

> > 3. paste the link into the ”Link to:“ field ... And there you'll get `A:a` instead of `:A:a`
> 
> Nope. I get ":A:a" here too... there must be something different between our two setups

Hmm... maybe some library with a different version.. don't have the time right now to research this, and it's not that important. In fact, I sometimes write `A:a` manually (using the keyboard), and I'd like to have an absolute link as the (suggestion) result. However, when I write `A:` or `A:a` (on Zim w/ this PR) into the ”Link to:“ field, the suggestion is a floating path. See below.

Type `A`:

spectacle r31022

Type `A:`

spectacle r31022

IMHO this is a missing change in this PR.

---

Regards the ”_don't use floating links_“ option I don't have a clear opinion.. Still, for my self I love to have a ”huge“ list of options – given they aren't buggy, of course.

jaap-karssenberg:
Sorry to close such long conversation, but for me the way forward is to make floating links work correctly, not disable them.

myrdd:
@jaap-karssenberg that's a perfectly acceptable design decision imho. Still, after some weeks using this patch<sup>(1)</sup>, I actually _prefer_ non-floating links, **except** relative links (`+...`). I'd like to have an (advanced) option called ”_don't use floating links_“ or similar. Should/may I create an issue for that?

I already suggested to the OP (@Osndok) to make this PR not change the default behavior, but rather introduce such a ”_don't use floating links_“ option, but he opinionated against it:

> > BTW, shouldn't ”don't use floating links“ rather be an (advanced) option?
> 
> If floating links where entirely fixed, then maybe... but in general I would say that there should not be an option that is known to cause data loss.

— https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/492#issuecomment-466675061

<sup>(1)</sup> FYI: I am using zim very frequently; and its linking feature and ”jump to“ dialog heavily.

jaap-karssenberg:
I'm open to a patch that makes such behavior optional.

Regards

Jaap

Op do 4 apr. 2019 16:50 schreef Martin Kimmerle <notifications@github.com>:

introt:
> This commit removes Zim's ability to generate floating links, while not
> removing general support (both for data-in-place and as the user might
> manually enter)

> However, when I write A: or A:a (on Zim w/ this PR) into the ”Link to:“ field, the suggestion is a floating path. See below.

Here that counts as manual and is a bit trickier to get properly rid of. The easiest way seems to be to make the dialog parse the floating links and write absolute links in their place when closed (better UX would be after selection), which is what I've done in [my patch](https://github.com/introt/zim-desktop-wiki/commit/bb3c61aba8166eb8d199bab13e3a0233ae4ecdbe).

> Without html-like fragment links...

Now that we have anchors, I've also made same-page links absolute.

While writing the patch I came across some smelly code: inside `InsertLinkDialog.do_response_ok` there's a variable named `href`, which is a `str`, not a `HRef`- I tried to debug this with the `type` built-in, but the name was used to store a string returned by `link_type`... Before I finish this I'm going to submit a PR for raising the required Python version to a non-EOL one with type hinting.

PR #507, opened 2018-08-11
Task list: setting to display due date in sidepane view.

fhaftmann:
For my personal task list workflow *both* priority and due date are essential properties, hence I need the ability to sort after either of these in the side pane view. Maybe that is appropriate and worthy to be included into zim in general.

This discontinues the TODO comment to split the different views of the task list into different classes: different column layouts can be specified by an explicit enum parameter.

I am open to criticism and would contribute a German translation after merge.


jaap-karssenberg:
Needs further review & testing

PR #522, opened 2018-09-02
Advise about PyGTKAllInOne package for Windows

bimlas:
Installing PyGTK seems to be weird in the description on http://www.pygtk.org/; using PyGTKAllInOne (as you mentioned on https://github.com/jaap-karssenberg/zim-wiki/wiki/Install-the-full-Zim-Desktop-Wiki-Suite-on-Windows) is far easier.


lrq3000:
Yes that's what I did to compile a binary for Windows of Zim, but Jaap is migrating the whole app to GTK3 so I'm not sure if these instructions will still work.

jaap-karssenberg:
I believe this request has become outdated with the move to Gtk3

PR #558, opened 2018-11-12
Alt-F12 shortcut for "Migrated" checkbox

Osndok:
I suggest, that to continue the pattern of checking checkboxes using the keyboard, we should have a keyboard shortcut to check a box for "migrate" too.


jaap-karssenberg:
Sorry to reject but would prefer a patch that toggles through states with a single keybinding - same as what the mouse click does.

Just merged keybinding config in preferences menu, will make it easier to customize

Issue #563, opened 2018-11-22
svg attachments crash (segfault) Fedora 29 Linux (Zim 0.68) -- solution here


BehindTheBrain:
svg attachments crash Zim, even when the are not displayed as images in the page.

The fix is already noted in the code (from Windows):

In /usr/lib/python2.7/site-packages/zim/plugins/attachmentbrowser/thumbnailer.py
line 29: -- change it to this:
 or (file.basename.endswith('.svg')):

jaap-karssenberg:
Need at least an option for this

jaap-karssenberg:
Dev notes:
  * Add option to thumbnail svg, default off
  * Re-test svg support on windows with Gtk3 under msys

PR #564, opened 2018-11-22
SVG causes segfaults in Linux too (fix)

BehindTheBrain:
Seems like this now needs to mimic the Windows condition.


Osndok:
For the sake of organization, I thought I might link this pull request to the issue you just opened... #563

jaap-karssenberg:
Afraid I have to reject this merge request. This is just disabling SV thumbnailing for all uses. Should at least be optional, now also users for who all works fine loose functionality.

PR #641, opened 2019-01-28
Fix noteobok typo

rolfkleef:


jaap-karssenberg:
Will add to 0.71 list -- not merging for 0.70 anymore since it will break
translations.

-- Jaap

PR #649, opened 2019-01-31
Patch/regex

banan314:
#408 but directly into `master`


jaap-karssenberg:
Will merge into develop

Issue #410, opened 2018-04-04
Deprecated tags in HTML exported document


banan314:
Regards exporting a Zim notebook to HTML document.

Some of the tags are deprecated:

- `<u></u>` (highlight, as defined in CSS), https://www.w3.org/TR/html52/textlevel-semantics.html#the-u-element:
  > In most cases, another element is likely to be more appropriate: for marking stress emphasis, the em element should be used; for marking key words or phrases either the b element or the mark element should be used, depending on the context; for marking book titles, the cite element should be used; for labeling text with explicit textual annotations, the ruby element should be used; for technical terms, taxonomic designation, transliteration, a thought, or for labeling ship names in Western texts, the i element should be used.

- `<s></s>` (strike through), https://www.w3.org/TR/html52/textlevel-semantics.html#the-s-element

  > The s element is not appropriate when indicating document edits; to mark a span of text as having been removed from a document, use the del element.

jaap-karssenberg:
Current templates are HTML 4. To support HTML 5 there should be a template option to inform the dumper about the html version required and based on this option output valid html 5. Without such option existing html 4 templates will break.

PR #678, opened 2019-02-19
Issue #410

laeccan:
Apply recommended changes for 'good first issue'.

### DIFF
--- a/zim/formats/html.py
+++ b/zim/formats/html.py
@@ -41,8 +41,8 @@ class Dumper(DumperClass):
 	TAGS = {
 		EMPHASIS: ('<i>', '</i>'),
 		STRONG: ('<b>', '</b>'),
-		MARK: ('<u>', '</u>'),
-		STRIKE: ('<s>', '</s>'),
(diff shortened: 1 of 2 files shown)


zim/formats/html.py
```diff
@@ -41,8 +41,8 @@ class Dumper(DumperClass):
 	TAGS = {
 		EMPHASIS: ('<i>', '</i>'),
 		STRONG: ('<b>', '</b>'),
-		MARK: ('<u>', '</u>'),
-		STRIKE: ('<s>', '</s>'),
+		MARK: ('<em>', '</em>'),
```
Self-Perfection: This change probably calls for updating HTML template CSS to make `<em/>` more visually prominent to match default zim style.

jaap-karssenberg:
Afraid this patch needs to be a bit more complex:

Current templates are HTML 4. To support HTML 5 there should be a template option to inform the dumper about the html version required and based on this option output valid html 5. Without such option existing html 4 templates will break.

PR #684, opened 2019-02-25
Convert large notebook script to python3

mexicarne:
Make zim-desktop-wiki/tools/create_large_test_notebook.py work with python 3

### DIFF
--- a/tools/create_large_test_notebook.py
+++ b/tools/create_large_test_notebook.py
@@ -4,8 +4,8 @@ import sys
 import os
 
 if len(sys.argv) != 2:
-	print 'Usage: %s DIRECTORY' % sys.argv[0]
-	sys.exit(1)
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
Thanks for the fix. Looks like the diff has some whitespace changes - is indenting consistent over the file using TAB ?

mexicarne:
> Thanks for the fix. Looks like the diff has some whitespace changes - is indenting consistent over the file using TAB ?

No, the whole file was converted to space indenting per pep8 and because most lines in this file already was indented by space. Do you prefer to convert it tab instead?

jaap-karssenberg:
Yes, please, zim code base ignores that part of pep8 :)

For the source code there is a test to enforce the indenting - probably I
should make it cover the tools section as well.

mexicarne:
Done in 0b6223a

PR #688, opened 2019-03-02
Improved dialog feedback when file save failed (#163 and #178 modified to merge with 0.70-rc3)

joukhaha:
Updated the PR addressing #163 and #178 to work with the 0.70 version.

### DIFF
--- a/data/templates/html/Default_with_index_post.html
+++ b/data/templates/html/Default_with_index_post.html
@@ -1,0 +1,147 @@
+<html>
+<head>
+	<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+	<title>[% title %]</title>
+	<meta name='Generator' content='[% generator.name %]'>
+	<style type='text/css'>
+		a          { text-decoration: none      }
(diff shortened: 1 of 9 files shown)


jaap-karssenberg:
Please check your merge request, looks like there are other changes in this request as well. So needs to be cleaned up to only have the feature mentioned.

joukhaha:
Sorry about that. Those changes only related to www.py (and a new html-template). Removed those so now the changes should only be addressing the enhancement at hand...

joukhaha:
Oh, and probably goes without saying, but please take the changes with a grain of salt. They likely need further attention and care before being ready for including in a stable version.

lrq3000:
Thank you so much @joukhaha ! I'll try to test this version when I'll get the time, but first, since I'm running on Windows, I'll have to find a way to compile the new Zim for Windows, and it seems it's not that simple since the previous maintainer gave up...

joukhaha:
This functionality would now be provided by a plugin, that should be much cleaner in implementation, and can be easily configured/activated... See #1986

PR #693, opened 2019-03-04
In Flatpak use flatpak-spawn to call host command if possible.

### LINKED ISSUE #686: Zim Verison plugin does not see git installation
Summary:
The Verison plugin does not detect git from the plugins menu, even after using bash to git init a repo manually.

Current System:

- Fedora 29, up-to-date (as of this posting)
-Git, version 2.20.1
-Zim 0.69.1 (installed through the flatpaks option in the Fedora "Software" app).

Steps taken:
-Attempted to add Version plugin from Zim menu, says Git it not detected.
-From bash cli, I made a "git init" from the desired notebook @ cd: "/Notebooks/Notes"
-"which git" confirmed git was present. Plugin did not work.
-I then did "git add" and "git commit" on available files. Plugin was still unavailable.

Let me know the best next steps for me to resolve this, thank you!

cheese:
Flatpak application does not see real host command. Host command
can be executed only through D-Bus. But the interface is
asynchronous. We use the 'flatpak-spawn' utility to call host
command synchronously.

Fixes #686

### DIFF
--- a/zim/applications.py
+++ b/zim/applications.py
@@ -30,6 +30,21 @@ TEST_MODE_RUN_CB = None
 
 _ENCODING = locale.getpreferredencoding() # FIXME: assume processes to adhere to same preference ...
 
+_FLATPAK_HOSTCOMMAND_PREFIX = ("flatpak-spawn", "--host")
+
+def _check_flatpak_host_command():
+	# Detect whether we are running in Flatpak and can call HostCommand
+	if os.path.exists("/.flatpak-info"):
+		try:
+			subprocess.check_call(_FLATPAK_HOSTCOMMAND_PREFIX + ("which", "which"), stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
+			return True
+		except subprocess.CalledProcessError:
+			pass  # Not privileged to call host command or "which" not found in the host
+		except OSError:
+			pass  # Failed to execute flatpak-spawn
+	return False
+
+_CAN_CALL_FLATPAK_HOST_COMMAND = _check_flatpak_host_command()
 
 def _main_is_frozen():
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
I like this approach as it seems a good way to address this issue without to much burden on the zim code.

My only concern is whether it makes sense to use `try ... except ...` if we can call the command without "flatpack-spawn" but it fails for another reason, it is called again with "flatpack-spawn".

To me it seems better to directly go for the "flatpack-spawn" variant if you are sure enough that you are inside the flatpack environment.

Does that make sense?

cheese:
> I like this approach as it seems a good way to address this issue without to much burden on the zim code.
> 
> My only concern is whether it makes sense to use `try ... except ...` if we can call the command without "flatpack-spawn" but it fails for another reason, it is called again with "flatpack-spawn".
> 
> To me it seems better to directly go for the "flatpack-spawn" variant if you are sure enough that you are inside the flatpack environment.
> 
> Does that make sense?

Commands may be available within flatpak environment and should take priority.

Another scheme instead of 'try ... except' is to check and remember command's running method at first lookup/run. Then if the `Application` object is kept, further running will directly use the right method.

jaap-karssenberg:
No, that kind of caching makes it more complex without removing the
concern, as it still goes wrong on the first run. Although it might be good
idea from a performance point of view. E.g. the version control plugin runs
git often enough that you might see it if you try every time.

What is the specific exception when a command can not run inside the
flatpack environment? If the "except" is selective for the specific error,
it makes the change of catching unrelated errors smaller.

-- Jaap

cheese:
Enhanced the patch to catch specific exceptions.

PR #704, opened 2019-03-11
Journal plugin – "Go to Today" boost

e3rd:
having some text selected, it is replaced with a link and moved at the end of today's page

(This is a reopening of #465 that was closed probably by mistake when changing the Gtk3 branch from `next` to `master`.)

### DIFF
--- a/zim/gui/pageview.py
+++ b/zim/gui/pageview.py
@@ -47,7 +47,7 @@ from zim.gui.widgets import \
 	Dialog, FileDialog, QuestionDialog, ErrorDialog, \
 	IconButton, MenuButton, BrowserTreeView, InputEntry, \
 	ScrolledWindow, \
-	rotate_pixbuf, populate_popup_add_separator, strip_boolean_result
+	rotate_pixbuf, populate_popup_add_separator, strip_boolean_result, Calendar
 from zim.gui.applications import OpenWithMenu, open_url, open_file, edit_config_file
 from zim.gui.clipboard import Clipboard, SelectionClipboard, \
 	textbuffer_register_serialize_formats
@@ -6914,7 +6914,6 @@ class InsertDateDialog(Dialog):
 		self.uistate.setdefault('linkdate', False)
 
 		## Add Calendar widget
-		from zim.plugins.journal import Calendar # FIXME put this in zim.gui.widgets
 
 		label = Gtk.Label()
 		label.set_markup('<b>' + _("Date") + '</b>') # T: label in "insert date" dialog
@@ -7827,6 +7826,7 @@ class MoveTextDialog(Dialog):
 		)
 		self.pageview = pageview
 		self.notebook = notebook
+
 		self.page = page
 		self.buffer = buffer
 		self.navigation = navigation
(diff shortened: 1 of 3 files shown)


jaap-karssenberg:
I understand the use for the feature, but I believe this should be a separate action. Now the "go today" action has very different behavior depending on selection state, which I believe will not be intuitive to many users.

Option would be to do this as an extension of the "move selected" text feature in the pageview context menu.

e3rd:
Since it doesn't concurrent any other feature I thought no user would be badly surprised – any backwards incompatibility.

I can hardly image how could I put this into `move selected`. Do you believe I could map this to an accelerator (so that I am not condemned to open a new "Move text to other page..." dialog)? I really like to hit `Alt<d>` on the text selection that I want to archive to a journal page.

Or would it sound better to you if I would make it a checkbox in a preferences/plugins/journal/configure or  notebook/properties/journal that a new user should opt-in to enable the feature? :) That would be much easier for me!

dpalmigiano (approved):

luffah:
As an user, it is aside of my use.
Moving text to another pages is an operation i do only when i archive tasks to journal.
A suggestion is to add a kind of button "In journal" in "Move selected" wizard to select the targeted journal note with the calendar (preselected on today).

jaap-karssenberg:
Afraid I cannot merge it in it's current form. The function is hard coded outside a plugin, which breaks the basic separation between core modules and plugins.

The proper way to implement this would be to put it in the plugin and then use an Extension class to add this functionality to the pageview. The easiest is probably to define a separate dialog and add it to the context menu by listening to the "populate-popup" signal.

Hope this helps,

Jaap

e3rd:
Another year, another request change (2018, 2019, 2020, 2021 :D ).

By plugin you mean an internal one, shipped with Zim, or should the functionality be realized as a separated plugin in an independent repository?

jaap-karssenberg:
My apologies for the long time outs in the discussion. Going forward, I'll really try to do better and better support pull requests.

You always have the option to go with a separate plugin and be free to customize it to your liking. Within the main repository I would want this as part of the journal plugin. 

In essence the change you did by adding the functionality to the "MoveText" dialog is OK from a user point of view. What I object to is that instead of adding the code to the plugin, you added the code to the dialog directly and check for the plugin in a round-about way. So not asking to change how it looks, just where the code is placed.

This can be implemented in two ways:
1. Copy the modified MoveText dialog code to the journal plugin and define an menu action to trigger it. On top of that the extension can listen for the popup signal and add this to the context menu as well
2. Add an "`@extendable`" decorator to the default MoveText dialog and in the plugin code define an Extension specific for that dialog. This avoids duplication of the dialog and allows you to add on to the dialog from the plugin

As long as your changes are in line with the intended use of the base MoveText dialog, option (2) would be preferred. If you want a complete different dialog, better to go with option (1).

(For 2, you would define an empty class "`MoveTextDialogExtension(DialogExtensionBase)`" in the pageview code. In the plugin you import this class and use it as a base class for your extension. The "`@extendable`" decorator links the extension class to the dialog itself, and then the plugin framework takes care of loading your extension every time the dialog is shown.)

Hope this helps,

Jaap

e3rd:
Jaap, thanks a lot for the detailed comment. I fully understand the situation of maintaining such a big project. With the details you've provided, I'll try to implement it in the months to come.

PR #706, opened 2019-03-11
accelmap migration info

e3rd:
I've been quite surprised that I have to renew the file `accelmap` since old accelerators stop working as of Zim 0.70. It may be a good idea to inform users they have to re-do they accelerator. I don't know where to put the information so I created a migrating section in the FAQ.


e3rd:
cool! :)

PR #707, opened 2019-03-12
Status bar change on keyboard Fix #452

### LINKED ISSUE #452: Show link destination for keyboard navigation
I love using Zim for years!  

When mouse cursor is on a link, its destination is visible in the status bar.  
However, when I navigate keyboard caret on a link, nothing changes.

In what files would you advise me to change it in so that I may prepare a pull request?

e3rd:
Show link destination for keyboard navigation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ Jaap Karssenberg <jaap.karssenberg@gmail.com>
 This branch is the Python rewrite and starts with version 0.42.
 Earlier version numbers for zim correspond to the Perl branch.
 
+## next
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
I believe there are better signals available to connect to for changes in cursor position rather than directly tying into the keybinding.

Have e.g. a look at [GtkTextBuffer mark-set ](https://developer.gnome.org/gtk3/stable/GtkTextBuffer.html#GtkTextBuffer-mark-set) (the cursor is managed by a mark) or use the "notify:cursor_position" signal

e3rd:
Thanks, in few weeks I'll try to submit an update!

PR #709, opened 2019-03-14
accel without menu item can be called again

e3rd:
Hello,
there was a regression since e7e8f42 – old `uimanager_xml` menu style formatting was removed and Bookmarks disappeared from menus. However, accelerators to that bookmarks stopped working. With the hereby pull-requested change the accelerators work again without being displayed in the menu. (But after 3 hours of debugging I really don't know why :( )

I am open to all comments and ideas how should I do that better :)


e3rd:
thanks!

PR #716, opened 2019-03-25
Fix crash in Edit Image dialog when the image file does not exist

qtonthat:
When the image file does not exist, Edit Image dialog failed to come up.
It was caused by ZeroDivisionError exception (shown by zim -V).

This simple change will the problem.

### DIFF
--- a/zim/gui/pageview.py
+++ b/zim/gui/pageview.py
@@ -6925,6 +6925,9 @@ class EditImageDialog(Dialog):
 		else:
 			width.set_sensitive(True)
 			height.set_sensitive(True)
+			if w <= 0 and h <= 0:
+				# image file does not exist
+				return
 			self._block = True
 			width.set_range(0, 4 * w)
 			width.set_value(w)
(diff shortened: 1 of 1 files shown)


qtonthat:
Missing image files can be common problems when users move sections between pages. I would suggest to offer users a chance to easily correct the problem using a dialog (as my commit does) rather than just raise exception and force them to edit the source text directly. It seems natural for users, seeing a broken icon in a page, to pop up the dialog to fix it.

jaap-karssenberg:
Sorry, from your patch that behavior was not obvious, because the other
path this may occur (the except block) was unchanged.

-- Jaap

qtonthat:
> Sorry, from your patch that behavior was not obvious, because the other path this may occur (the except block) was unchanged.

You're right.

jaap-karssenberg:
Looking at this again it seems with my patch you can still edit images that do not exist. Can you clarify the issue you see with the patch?

qtonthat:
Your patch was fine. It was my thinking that wasn't.

PR #750, opened 2019-04-03
Add gnome-screenshot to screenshot plugin.

qtonthat:
The existing screenshot commands (import, scrot) do not work properly
on all systems (artifacts).


jaap-karssenberg:
Merged in develop branch

jaap-karssenberg:
Merged in develop

Issue #443, opened 2018-05-13
Daily due dates for weekly journal


nelas:
Dear Jaap,

I've made a [crude proof of concept](https://github.com/nelas/zim-desktop-wiki/commit/001d8f18785f51f1f7fb773c66aeed1962a95ea0) for extracting daily due dates from a weekly journal (see [original thread](https://lists.launchpad.net/zim-wiki/msg04589.html) in the mailing list), and would like some input on the implementation.

**What I did:**
1. Get header text from `head` in [_parse_heading()](https://github.com/nelas/zim-desktop-wiki/commit/001d8f18785f51f1f7fb773c66aeed1962a95ea0#diff-210049b160d4db9dd5b0e106c609469aR345).
2. Use `datetime.strptime` to [parse the date](https://github.com/nelas/zim-desktop-wiki/commit/001d8f18785f51f1f7fb773c66aeed1962a95ea0#diff-210049b160d4db9dd5b0e106c609469aR353) from the header (e.g. "Monday 14 May"). This is done in a separate function [_parse_heading_due_date()](https://github.com/nelas/zim-desktop-wiki/commit/001d8f18785f51f1f7fb773c66aeed1962a95ea0#diff-210049b160d4db9dd5b0e106c609469aR361).
3. Modify `_parse_heading()` to also return the [parsed date](https://github.com/nelas/zim-desktop-wiki/commit/001d8f18785f51f1f7fb773c66aeed1962a95ea0#diff-210049b160d4db9dd5b0e106c609469aR357).
4. Update the due date of the object [defaults](https://github.com/nelas/zim-desktop-wiki/commit/001d8f18785f51f1f7fb773c66aeed1962a95ea0#diff-210049b160d4db9dd5b0e106c609469aR321), making all subsequent tasks have the new date (until the next header).

**Some issues:**
- I imported the [default datetime](https://github.com/nelas/zim-desktop-wiki/commit/001d8f18785f51f1f7fb773c66aeed1962a95ea0#diff-210049b160d4db9dd5b0e106c609469aR10) to use `strptime`. It's probably better to define `strptime` in `zim.datetimetz`.
- The headers have no year, so `strptime` sets it to 1900. To fix, I replace it by the [default_due_date.year](https://github.com/nelas/zim-desktop-wiki/commit/001d8f18785f51f1f7fb773c66aeed1962a95ea0#diff-210049b160d4db9dd5b0e106c609469aR319).
[...]

PR #780, opened 2019-04-10
Daily due dates in weekly journal

nelas:
Dear Jaap,

This is the pull request based on #443 (and [this thread](https://lists.launchpad.net/zim-wiki/msg04589.html)) to extract daily due dates from the weekly journal page (default template).

zim-daily-weekly

**Setup:**
- Properties > Journal > Use a page for each: > Week
- Properties > Task List >Use date from journal pages: > as due date for tasks

**What I did:**
1. Get header text from `head` in [_parse_heading()](https://github.com/nelas/zim-desktop-wiki/blob/a61f01c5161aa97f5888a2942097abee7b7f7afe/zim/plugins/tasklist/indexer.py#L357).
2. Use `datetime.strptime` to [parse the date](https://github.com/nelas/zim-desktop-wiki/blob/a61f01c5161aa97f5888a2942097abee7b7f7afe/zim/plugins/tasklist/indexer.py#L365) from the header (e.g. "Monday 14 May"). This is done in a separate function [_parse_heading_due_date()](https://github.com/nelas/zim-desktop-wiki/blob/a61f01c5161aa97f5888a2942097abee7b7f7afe/zim/plugins/tasklist/indexer.py#L373).
3. Modify `_parse_heading()` to also return the [parsed date](https://github.com/nelas/zim-desktop-wiki/blob/a61f01c5161aa97f5888a2942097abee7b7f7afe/zim/plugins/tasklist/indexer.py#L369).
4. Update the due date of the object [defaults](https://github.com/nelas/zim-desktop-wiki/blob/a61f01c5161aa97f5888a2942097abee7b7f7afe/zim/plugins/tasklist/indexer.py#L333), making all subsequent tasks have the new date (until the next header).

**Potential issues:**
- I imported the [default datetime](https://github.com/nelas/zim-desktop-wiki/blob/a61f01c5161aa97f5888a2942097abee7b7f7afe/zim/plugins/tasklist/indexer.py#L10) to use `strptime`. It's probably better to define `strptime` in `zim.datetimetz`.
- The headers have no year, so `strptime` sets it to 1900. To fix, I replace it by the [default_due_date.year](https://github.com/nelas/zim-desktop-wiki/blob/a61f01c5161aa97f5888a2942097abee7b7f7afe/zim/plugins/tasklist/indexer.py#L331).
- Modifying `_parse_heading()` to [return multiple values](https://github.com/nelas/zim-desktop-wiki/blob/a61f01c5161aa97f5888a2942097abee7b7f7afe/zim/plugins/tasklist/indexer.py#L326) might break something (tests went ok, though). But perhaps there is a better way.

Best,
Bruno

### DIFF
--- a/zim/plugins/tasklist/indexer.py
+++ b/zim/plugins/tasklist/indexer.py
@@ -31,6 +31,7 @@ from .dates import parse_date
 
 
 _tag_re = re.compile(r'(?<!\S)@(\w+)\b', re.U)
+_day_re = re.compile('(\d{1,2})')
 _date_re = re.compile('[<>] ?' + _raw_parse_date_re.pattern + '|\[d:.+\]')
 	# "<" and ">" prefixes for dates, "[d: ...]" for backward compatibility
 
@@ -158,8 +159,10 @@ class TasksIndexer(IndexerBase):
 			date = daterange_from_path(Path(row['name']))
 			if date and self.integrate_with_journal == 'start':
 				opts['default_start_date'] = date[1]
+				opts['daterange'] = (date[1], date[2])
 			elif date and self.integrate_with_journal == 'due':
 				opts['default_due_date'] = date[2]
+				opts['daterange'] = (date[1], date[2])
 
 		tasks = self.parser.parse(doc.iter_tokens(), **opts)
 		c = self.db.cursor()
@@ -303,10 +306,14 @@ class TaskParser(object):
 		self.task_label_re = task_label_re
 		self.all_checkboxes = all_checkboxes
 
-	def parse(self, tokens, default_start_date=0, default_due_date=_MAX_DUE_DATE):
+	def parse(self, tokens, default_start_date=0, default_due_date=_MAX_DUE_DATE, daterange=None):
 
 		defaults = [True, 0, default_start_date, default_due_date]
 					# [isopen, prio, start, due]
+
+		if daterange:
+			first_date = daterange[0] # first date
+			last_date = daterange[1] # last date
 
 		def _is_list_heading(task):
 			isopen, prio, start, due, tags, text = task[0]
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
I like the idea, but would want to make the parsing of the heading a bit more robust.

Since you already know the week, the heading only needs to give the day, so instead of strptime you might as well just use a regex to match a number. Probably more robust for locale settings etc.

nelas:
Hi Jaap, thanks for the feedback. I've altered the code to get the day using regular expression. It works, but there is a problem when there is a month transition like the one below.

Since the default due date here is `April 04`, when it fetches the `Tuesday 31 March`, it assigns it to `April 31`, which doesn't exist: `ValueError: day is out of range for month`. In addition, the wrong dates are assigned to the earlier dates March 29th and 30th (assigned to April).

Without `strptime`, there must be a way to identify the heading month. Any ideas?

```
Week 14 2020
============
29 March to 04 April

#### Sunday 29 March
[ ] Wrong due date

#### Monday 30 March
[ ] Wrong due date

#### Tuesday 31 March
[ ] Due date fails

#### Wednesday 01 April
[ ] Due date ok

#### Thursday 02 April
[ ] Due date ok

#### Friday 03 April
[ ] Due date ok

#### Saturday 04 April
[ ] Due date ok
```

Cheers,
Bruno

jaap-karssenberg:
The "daterange_from_path" gives both a start and an end date. Now we pass
along only one of these, but if we add the range as context in the parse we
can match the day numbers to dates in the range by looping through them.

That way you only need a number in this range to match to the right date.

To clarify: my objection to strptime is that you now need to match the
search pattern to the template. If you go that route, you somehow need to
pull this information from the template into the page and then into the
parser. Else the function breaks as soon as someone changes the template
(like I do fo my personal notebook ;)  )

Regards,

Jaap

nelas:
Ok, I think I got it working! I've added a `daterange` option to the `TaskParser.parse` method, and used that to loop and find the heading's date. It then sets the heading date as start or due date, depending on the values in the `defaults` list. Let me know what you think. Cheers

jaap-karssenberg:
Sorry, but it looks like you pulled a lot of other commits into your branch
- impossible to review the changes now.

Can you rebase your changes such that the merge request only show your
changes? (Or open a new request with a clean branch if that is easier.)

Thanks,

Jaap

nelas:
Hhm, ok. Just force-pushed a rebase but it's still messed up. I probably pulled the develop branch over this on the previous push... I'll open a clean branch.

nelas:
Alright! I managed to clean it up and put it in one commit.

jaap-karssenberg:
Would like to add
- test cases, specific where week covers boundary between months
- extend regex to also match fully specified dates yyyy-mm-dd or wyyww.d

jaap-karssenberg:
My apologies for leaving this pull request hanging for over a year. Started merging it with a branch that contains more tasklist improvements and realized I still don't feel comfortable with the parsing of the headers - it feels fragile e.g. in case there are numbers in the heading for another reason.

However, I also came up with the solution. Will also merge the pull request that introduces "anchors" #1293 . This allows inserting identifiers in the page that can be used for linking etc. So my best idea is to modify the journal template to have unambigous identifiers for the headings - e.g. #20210412 for today - and thus allowing to link directly to a date in a week or month page, but also allowing to parse the heading and find the correct date for the heading without any special intelligence.

Will keep this patch as you submitted it, but rework the parsing to use anchors instead of a number in the heading text.

nelas:
Hi Jaap, no problem. I also think the parsing is fragile. Using the anchor identifiers is a good solution!

Issue #84, opened 2017-05-31
A page with a "ß" in its name isn't found


sojusnik:
When a page name has a "ß" in its name, then "go to" (Ctrl+J) doesn't find that page.
Observed under v0.66 from the snapshots repository.

jaap-karssenberg:
Testing with my 0.67-rc1 snapshot I don't observe the same issue

sojusnik:
Create a "Maßnahmen" page. Then open the "go to" window (normally Ctrl+J) and type "Ma". At first you'll see that page, but when writing "Maß" the page will disappear.
Happens with the latest v0.66 from the snapshots ppa.

jaap-karssenberg:
Something weird going on here. Tested on windows system and there it works correctly. But on my linux system it doesn't. One level deeper the matching function gets passed "mass" instead of "maß", so somewhere in the library it is being more intelligent than it should. No idea if this is input method related, locale related, or something else.

sojusnik:
With v0.70 this happens also to page titles containing umlauts like ö etc.

Observed on Ubuntu 18.10

mexicarne:
Works perfectly fine here in 0.70-rc3. The same with umlauts.

jhakonen:
I found a way to fix this.

The `key` in `gtk_entry_completion_match_func` is utf8 normalized and casefolded. This is described in documentation: https://lazka.github.io/pgi-docs/Gtk-3.0/callbacks.html#Gtk.EntryCompletionMatchFunc

So, there's two possible ways to fix this (as mentioned in the documentation):
1. Instead of using `key` retrieve the key value from the entry using `get_text()`. That value is not normalized/casefolded and can be compared against `text`.
2. Normalize and casefold the `text` variable and then compare it and the `key`. At comparison the `.lower()` is not needed due of the `text` being already casefolded. Right arguments for normalizing and casefolding I checked from GTK's sources: https://github.com/GNOME/gtk/blob/3.24.4/gtk/gtkentrycompletion.c#L1290

Here's the two different options as patches:
https://gist.github.com/jhakonen/88a8e96ddaa4adf0fb558416dec919c4

I'm not sure which one you think is better, so I didn't do PR. 

Tried with pages "Tärkeä" and "Maßnahmen" using Ubuntu 18.10.

jaap-karssenberg:
[...]

PR #781, opened 2019-04-10
Fix 'jump to' not providing completions with non-ascii characters

jhakonen:
See #84 for details.

I went with the normalizing/casefolding approach as the Gtk.EntryCompletionMatchFunc's documentation seemed to prefer that. 

Includes also a few additional tests with non-ascii test strings.


jaap-karssenberg:
Merged

myrdd:
Thanks a lot @jhakonen

PR #800, opened 2019-04-16
Replace auto redirecting http:// with https://

Self-Perfection:
Replaced with https:// http:// links which redirect to https:// anyway
(Only in places which seemed safe to do this replacement)

Used following script to find such links:

```bash
#!/bin/bash

IFS= readarray -d '' FILES < <(git --no-pager grep --null -l http:// ":(exclude)*svg")
for FILE in "${FILES[@]}"; do 
    echo "$FILE"
    sed -i 's|http://www.zim-wiki.org|https://www.zim-wiki.org|' "$FILE"
    for URL in $(grep --extended-regexp --only-matching 'http://[^| ]*[^.|]' "$FILE"); do
        HTTPS_URL="https:${URL#http:}"
        REDIRECT_URL=$(curl --max-time 5 -L --head --output /dev/null --silent --write-out '%{url_effective}' "$URL")
        if [[ "$REDIRECT_URL" == "$HTTPS_URL"* ]]; then
            sed -i "s|${URL}|${HTTPS_URL}|" "$FILE"
        fi
        echo "$URL -> $REDIRECT_URL"
    done
    git add -p "$FILE"
done
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,5 @@
 # Based on template here:
-# http://editorconfig.org/
+# https://editorconfig.org/
 
 # top-most EditorConfig file
 root = true
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ and
 * Bug tracker and feature requests:
   https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues
 * Public wiki:
-  http://www.zim-wiki.org/wiki/
+  https://www.zim-wiki.org/wiki/
 * Mailing list:
   https://launchpad.net/~zim-wiki
 * IRC Channel:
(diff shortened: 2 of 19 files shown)


jaap-karssenberg:
After merging found another test hit. Did you run the test suite before committing ?

Self-Perfection:
I did not. Shame on me!

I am stuck with running tests in isolated environment in vagrant, so I relied on Travis... which never came to check latest commit in this PR.

I'll be more diligent next time.

jaap-karssenberg:
Agreed Travis should have caught it. Not sure why it did not update. Still
learning how that works.

Regards,

Jaap

Op za 27 apr. 2019 14:29 schreef Alexander Meshcheryakov <
notifications@github.com>:

Issue #795, opened 2019-04-14
Swap format and calendar boxes in Insert Date and Time dialog


nelas:
Even after years of daily Zim use, I still get confused by the insert date dialog. Because the format box comes first, I often click on it first, and this results in the accidental insertion of the date before I actually select the date I want.

I've experimented swapping the format and calendar widget in `zim/gui/pageview.py` so that the calendar comes first, and it immediately feels more intuitive. I'm getting less accidental date insertions and it also seems more natural to set the date first and choose the format later.

![Screenshot from 2019-04-15 00-34-16](https://user-images.githubusercontent.com/361179/56100393-70554100-5f18-11e9-8404-b858f69f96f2.png)

Not sure how often other people have this issue, but it's a simple change in the code that I think would improve the user experience for this dialog - which I use all the time ;)

Cheers!

Osndok:
This has always got me too... I think it's because we expect to work the dialog from the top-down. At one time, I even set out to swap the order, but I don't remember the outcome.

Please, share your patch!

PR #802, opened 2019-04-16
Swap order of the format and calendar panels

nelas:
Changes the order of the format box and calendar widget in the Insert Date and Time dialog for better user experience. See #795


jaap-karssenberg:
Merged in develop branch

Issue #128, opened 2017-06-17
"Ctrl+G" doesn't work after "Ctrl+F"


sojusnik:
Since this is a quite old and common request (we have 2 [tickets](https://bugs.launchpad.net/zim/+bug/1036821) on [launchpad](https://bugs.launchpad.net/zim/+bug/1589085) and some mentions in the mailing list) I open this issue here, in the hope to fix this issue quickly:

When using "find" (normally Ctrl+F) within a page, then searching for the next search result with "Ctrl+G" (search again) doesn't work, because the focus stays on the "find entry", as mentioned [here](https://bugs.launchpad.net/zim/+bug/1036821/comments/3).

The solution seems obvious. Maybe somebody could have a quick look on that?

jaap-karssenberg:
To be exact: if you start Ctrl-F but don't finish it by <Enter> the focus remains on the find entry and Ctrl-G doesn't work. If you close by <Enter> it will as intended.

There is another ticket open that claims correct behavior is to always keep focus on the find entry, so repeated <Enter> is repeated find. Need to take these two tickets together.

... if only I could find the ticket number ...

PR #808, opened 2019-04-18
Improve find bar behaviour (fixes #317 and #128)

### LINKED ISSUE #317: "Find" search box next occurrence shortcut
This issue is about the "Find" search box found in "Search->Find...".

Currently \<Alt\>N can be used to go to the next occurrence of the search. In most other applications \<Enter\> can be used to go to the next result (examples: Chrome, Firefox, Sublime Text, Word and many more). Currently Zim will highlight the next result when pressing enter and the find search box will loose focus. This is annoying for me since I'm used to keep pressing enter when searching through text. Since \<Enter\> will go to next occurrence in many other widely used applications I assume that I'm not the only one who often makes this mistake when using Zim.

So i suppose the following: Keep the current shortcuts and make \<Enter\> go to the next occurrence like \<Alt\>N.

seihtam:
This makes the find bar behaviour match other common applications.

* When a match is found don't defocus the find bar.
* Press Ctrl+G or Enter to go to next occurence.
* If Ctrl+F is pressed and there is already text inside the find bar then the text is highlighed (even if it is already visible).

Also fixes launchpad bugs:
https://bugs.launchpad.net/zim/+bug/1036821
https://bugs.launchpad.net/zim/+bug/1589085
https://bugs.launchpad.net/zim/+bug/1685371

### DIFF
--- a/zim/gui/pageview.py
+++ b/zim/gui/pageview.py
@@ -5339,7 +5339,14 @@ class PageView(GSignalEmitterMixin, Gtk.VBox):
 		def set_actiongroup_sensitive(window, widget):
 			#~ print('!! FOCUS SET:', widget)
 			sensitive = widget is self.textview
-			self._set_menuitems_sensitive(sensitive)
+
+			# Enable keybindings and buttons for find functionality if find bar is in focus
+			force_sensitive = ()
+			if widget and widget.get_parent() is self.find_bar:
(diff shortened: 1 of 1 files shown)


kjyv:
@jaap-karssenberg How about merging this sometime soon? It's small and fixes multiple open bugs.

PR #827, opened 2019-05-03
Workaround for #390

zznw:
Workaround for drag-and-drop in pageindex pane (#390) using a temporary class variable, since gtk would otherwise fail to pass the correct `selectiondata` and only give an empty one.

Dragging from pageindex _to_ the editor (for creating hyperlinks) is still broken.

Drag-and-drop from other sources, such as the PathBar, were never affected by the bug. They will work both with or without this fix.

The global variable isn't a problem in practice because you'll never be dragging and dropping from multiple sources at once, so by the time one operation is done the variable will have been reset. Still, it is somewhat hacky so use as you wish.


jaap-karssenberg:
The thing is, the drag & drop does work under some systems. Now the work around completely overrules the default behavior. Would prefer a work around that detects that the content is empty before it takes the short cut.

jaap-karssenberg:
Committed a workaround for this issue in the development branch just now, please test if it solves the issue

PR #847, opened 2019-05-21
Check if None and support them failing for standards out and errror's flushing. Fixes #842.

### LINKED ISSUE #842: When zim started without terminal, crashes with "input/output error"
I have zim 0.71.0 installed from APT under KDE Neon under Ubuntu Bionic.

Starting `zim` from applications (normal procedure after clicking the "window" key) makes it crash.

I tried also from the [Launcherposta](http://launcherposta.taniquetil.com.ar/) project and the behaviour is the same.

I tracked the problem until the `ZimApplication._daemonize` method, where (after deciding it's not started from a terminal) it tries to daemonize, and it just crashes with any of the two following lines: 
```
			sys.stdout.flush()
			sys.stderr.flush()
```

If I comment those two out, all works just fine

facundobatista:


facundobatista:
Note that tests are not attached as that part of the code is not yet tested (and I'm too new with the codebase to start adding tests for that special part), but this was tested IRL successfully. 

See #842  for reference, too.

Thanks!

PR #900, opened 2019-07-23
spawn_async: missing working_directory argument

RobertHaileySlalomBuild:
Pretty straight forward... we go through the trouble of passing in an optional working directory just to drop it on the floor.


jaap-karssenberg:
Merged as 1713e4977c15922dbaf43e42ac09a25839e0704e in develop branch

PR #905, opened 2019-07-31
don't use main menu mnemonics on macOS

### LINKED ISSUE #904: unable to enter euro character on macOS
My keyboard layout is german so it has the euro character on alt-e (like e.g. also italian).
Alt-e in zim however opens the Edit menu. 

It's a bit tricky for me to find out if zim or a dependency is to blame. A few things I've noticed:
- gtk3-demo does not include any menus at all that are within the application window, it uses global menus via GtkApplication/GMenu - these are proper Quartz menus without the issue. The currently used classes Gtk.UIManager, Gtk.Action, and Gtk.ActionGroup to create the main menu are apparently outdated since gtk 3.10. That way also main menu plugins would not be necessary anymore (also fixing #874).
- As per https://superuser.com/questions/303525/what-is-the-shortcut-to-access-the-menubar-in-mac-os-x, it seems not to be common that you can open the menus by underscore letter on macOS. If that functionality comes from zim directly (and if the current menus should stay), it should be disabled on macOS to allow shortcuts with alt-<letter>

kjyv:
This fixes #904. The approach is simple - it removes underscores from the menu labels after they are translated but before they are sent to gtk.


jaap-karssenberg:
Is this still needed when using the plugin proposed in #1008 ?

-- Jaap

kjyv:
The plugin is nice and doesn't have these issues. Only if it was not activated for some reason (e.g. missing dependencies). So I wouldn't say the patch is useless per se

jaap-karssenberg:
OK, I'll look into merging it

PR #909, opened 2019-08-18
Sourceview: setting to have linenumbers off for newly created items.

domenn:
Configurable, by default it is is same as it has been, linenumbers are visible. This affects behaviour for newly created blocks. After it has been created, line numbers can be displayed or hidden, just as it has been poissible until now. 

There are currently no tests - I can write some if you want and explain me basic logic of manipulating tree. 
My idea is to create codeblock, check linenumbers, change setting and create new codeblock. 
I was looking into it and don't understand how `tree.find('object')` works, and after calling action.activate for the second time, tree (tried `print(tree.tostring())`) doesn't appear to change.

### DIFF
--- a/zim/plugins/sourceview.py
+++ b/zim/plugins/sourceview.py
@@ -97,11 +97,11 @@ class SourceViewObjectType(InsertedObjectTypeExtension):
 		self.connectto(self.preferences, 'changed', self.on_preferences_changed)
 
 	def new_model_interactive(self, parent, notebook, page):
-		lang = InsertCodeBlockDialog(parent).run()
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
A better way to implement this is to add the option to the InsertCodeBlockDialog dialog. If you make it a "uistate" setting, the dialog will remember the last used value and use that by default.

Regards,

Jaap

domenn:
Changed it.

jaap-karssenberg:
Merged in "develop" branch as 9a1ff2c53eba8da1982fe816ed2b1067dc7edece

PR #910, opened 2019-08-21
Slight improvments to the statusbar

### LINKED ISSUE #618: Optimise the pane at the bottom
The sections in the pane at the bottom should be [without borders](https://i.imgur.com/NJdD6RN.png) and also slightly lower.

zappedfish:
- Gtk.Frames in statusbar are no longer inset. 
- Statusbar buttons are now CSS styled to have no borders. 
- Three dots are removed from Backlinks label in favor of an arrow by default.
- Style indicator and INS/OVR indicator switched places.

Fixes #618.

This is my first proper pull request, so I probably messed some things up :-)

| Old  |
| ----- |
| Old|

| New |
| ----- |
|New|

### DIFF
--- a/translations/am.po
+++ b/translations/am.po
@@ -685,10 +685,10 @@ msgstr "_እንደገና መጫኛ"
 #. Label for button with backlinks in statusbar
 #: zim/gui/mainwindow.py:922
 #, python-format
-msgid "%i _Backlink..."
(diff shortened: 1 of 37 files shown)


jaap-karssenberg:
merged in 008b4b8fb923e999ce0b0197dbdd3fb92ddf7b43

PR #920, opened 2019-09-06
Multiple appdata fixes

cheese:


jaap-karssenberg:
How this "releases" XML tag work and what is it being used for? Should we
update that in the release script? If so, do you replace the tag, or add
new sub-tag ?

cheese:
> How this "releases" XML tag work and what is it being used for? 

It is used by Flatpak to determine the release history and the current release.

> Should we update that in the release script?

Making the release script do the job is a good idea. I will try to update it and the PR.

> If so, do you replace the tag, or add new sub-tag ?

Not required. I deployed patches for 0.72.0 for Flathub.

jaap-karssenberg:
In my last question I meant how the script should update the info, but I'll
wait for your proposal and check it. Automating in the
"tools/create_release" script ensures we don't forget :)

cheese:
Commits updated.

jaap-karssenberg:
Had to remove the download information from the update script:
- no guarantee that the dist file that is visible when this script is run is also the final one ending up on the website
- conflict in build sequence as the (updated) appdata file is part of the download and thus changes the checksum

PR #926, opened 2019-09-12
German translation for setting to display due date in side pane.

fhaftmann:
This adds a German translation for a particular setting and contains the formal updates of the corresponding translation sources files.


jaap-karssenberg:
Please submit translation updates via: https://translations.launchpad.net/zim

This patch is hard for me to review due to the many lines that changed. Also it can be overwritten with the next import from the launchpad website. 

Thanks,

Jaap

Issue #846, opened 2019-05-16
[enhancement] custom auto correct (same as --> undated into →)


gandrille:
When typing "-->" is automatically updated into "→". This is great :-)

I would love having a dialog to add/remove other replacements, such as:

coeur into cœur (have you noticed the single unicode character for œ? It's really painful to write this kind of word in french!)
!= into ≠
<< into «
>> into »

Since the replacements are user dependent, it could be nice to have the ability to edit them.

Note: this feature exists in Word (and I have used it intensively in the past!) For example, have a look at https://www.groovypost.com/howto/howto/word-2010-autocorrect-adds-symbols-and-replaces-words/
![image209](https://user-images.githubusercontent.com/621343/57879690-5069b380-781d-11e9-8394-6db574cb7092.png)

sojusnik:
Never noticed that Zim has an auto correct feature, as depicted with -->. Are there other examples as well? Is this somewhere documented?

Would love to have an ability to add my own abbreviations.

For other examples you could have a look on LibreOffice Writer.

jaap-karssenberg:
This replacement is done by the "insert symbol" plugin. Some of your
examples can be put in the config file of that plugin (see the edit button
on the insert symbol dialog).

Op do 16 mei 2019 21:18 schreef sojusnik <notifications@github.com>:

gandrille:
Thanks for your answer, which is fine with me for most of the use cases. For sure, I will use this feature!
Feel free to close the issue, or to keep it open, if you want to address my "cœur" replacement in a future release...

And many many thanks for Zim wiki :-)

tildagail:
'œ' is in the Insert Symbol dialog. You can insert it with the shortcut \oelig by default and as a quick experiment I added an autocorrect using just oe. Just add this line somewhere to the symbols.list file and  then whenever you type oe it turns into œ:

`oe	339`

gandrille:
@tildagail : Thanks to @jaap-karssenberg answer, I have already customized my symbols.list file in that direction. It's really effective for single symbol replacement :+1:

introt:
>  Feel free to close the issue, or to keep it open, if you want to address my "cœur" replacement in a future release...

@gandrille Can this be achieved with the autocorrect plugin and a custom dictionary?

gandrille:
@introt It is possible to have it replaced, but with a manual action every time the world is typed...which is not what I'm expecting.
 
[...]

PR #939, opened 2019-09-27
Autocompletion / autocorrection plugin

Dav-d:
Hi,

I think it would be usefull to have an autocompletion plugin with the ability to modify a custom dictionary.

Thank you for your job.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,10 +2,10 @@ Contributing
 ============
 
 Thank you for considering to contribute to the zim desktop wiki project. Zim
-is an open source project and run by volunteers, zo all help is welcome.
+is an open source project and run by volunteers, so all help is welcome.
 
 
-**Help wanted:** there are many issues in the bug tracker need someone to pick
+**Help wanted:** there are many issues in the bug tracker that need someone to pick
 them up. To get started please have a look at the
 [good first issue](https://github.com/zim-desktop-wiki/zim-desktop-wiki/labels/good%20first%20issue)
 and
--- a/PLUGIN_WRITING.md
+++ b/PLUGIN_WRITING.md
@@ -1,9 +1,9 @@
 Plugins
 =======
 
-Plugins one of the two extension mechanisms supported by zim. Plugins are loaded
+Plugins are one of the two extension mechanisms supported by zim. Plugins are loaded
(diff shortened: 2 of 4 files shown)


sojusnik:
I like how Gedit implemented wordcompletion. Easily installable in Ubuntu by `sudo apt install gedit-plugin-word-completion`.

Zim could pull those words from a general dictionary and also use words from the current note (or maybe from the last 10 used notes, as an example) to interactively autocomplete them, see Gedit's plugin for example.

BTW, Zim already has some type of autocompletion, in this case it's an autocorrection, realized through the "insert symbol" plugin, as mentioned [here](https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/846).

Dav-d:
I didn't know the Gedit wordcompletion. It's a very good model. But it doesn't work from a dictionary.
So I think too that Zim could work from a dictionary and from notes.

Thank you for the "insert symbol" plugin, (but it doesn't replace a realtime word autocompletion).

Dav-d:
Here is a screeshot from gedit : 

zim-plugin-autocompletion

and a template to add new custom completions  (from LibreOffice) :
zim-plugin-autocompletion-dictionary
It could do too the job of the insert symbol plugin.

In Libreoffice there is too an usefull option : a checkbox to add or not a space after the inserted word.

Dav-d:
Actually, I think there is a difference between autocompletion and autocorrection (self-correction ?).

Autocompletion means suggestions as gedit do from note or a word dictionary.

Autocorrection means to replace a word by another after pressing the space bar ie: "zw" --> "zim-wiki", from a replacement list like the sources.list.

I think that an autocorrection plugin will be easier to develop from the insert symbol plugin, isn't it ?

Dav-d:
Here is an example from the file symbol.list to add the autocorrection ability :

```
## Suggestions for more shortcuts are welcome - please file in bug tracker ##
wd		word
wrd		word
templte	template
tpl		template
etc.
```

jaap-karssenberg:
Something went wrong here - looks like you use a merge request for an unrelated issue / proposal. Please file as proper issue.

introt:
Hi @Dav-d, have you tried https://github.com/introt/zim-plugin-text-replacer?

PR #945, opened 2019-10-01
Update info about zim in Arch Linux

Self-Perfection:
Currently zim in community repo and not in extra.


Self-Perfection:
Actually it seems to me it would be easier to replace section with mentioning zim in linux repos with following snippet:
[latest packaged version(s)](https://repology.org/project/zim/versions)

Unfortunately zim does not support embedding images from the web, so I could not create such PR.

Self-Perfection:
Oh, noticed tools/build_website.sh script, so committed version that uses repology badge. Badge for all versions seemed too tall, so used tiny version.

PR #949, opened 2019-10-06
PLUGIN_WRITING: get_extensions is find_extension

WillForan:
update to find_extension, also include where to find it.

### DIFF
--- a/PLUGIN_WRITING.md
+++ b/PLUGIN_WRITING.md
@@ -121,7 +121,7 @@ module.
 
 ## Accessing functions of other plugins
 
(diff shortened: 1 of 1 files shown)


PLUGIN_WRITING.md
```diff
@@ -121,7 +121,7 @@ module.
 
 ## Accessing functions of other plugins
 
-The functions `get_extension()` and `get_action()` can be used to access
+The functions `zim.plugins.find_extension()` and `get_action()` can be used to access
```
jaap-karssenberg: get_action also become find_action

PR #968, opened 2019-11-05
fixed syntax in documentation for headers.

ionisingTuna:
The current syntax for headers with trailing '=' marks seems to create issues on my implementation of Zim.


jaap-karssenberg:
Rather than "fixing" the documentation I think we should fix the behavior

jaap-karssenberg:
Merged alternative fix in develop branch

Issue #544, opened 2018-10-06
Source View/gtksourecview style selection


machin3io:
I'm looking for a way to set the style used by the Source View plugin. 

There are a few different ones in /usr/share/gtksourceview-2.0/styles and more can be found onlines. 

I'm just not sure how to tell zim/Source View which one to use. I'd appreciate any pointers, thanks.

jaap-karssenberg:
This would require patch of the plugin to add style selection to the preferences.

Since there are custom widget for this, will require overloading the default plugin preferences dialog.

shermanbowling:
I'd like to second this feature request. As it stands, I can't read dark blue code comments with my dark gtk theme, and the selected syntax colors are all around fugly.

PR #979, opened 2019-12-04
Add a theme choice for the gtksourceview plugin

lucvtzel:
First commit for the issue #544. Not sure that it is the best solution, but it works.

An example with the xubuntu-dark theme:
ScreenCapture_2019-12-04_18-09-44


jaap-karssenberg:
The hard coded path location is blocking for integration. Is there a way to
make this a dynamic lookup?

Op wo 4 dec. 2019 18:18 schreef lucvtzel <notifications@github.com>:

lucvtzel:
ok, i will try to correct this  ! 
:-)

lucvtzel:
It's done (a93d455).

PR #983, opened 2019-12-09
Fix #512: Allow "paragraph-background" in style.conf

### LINKED ISSUE #512: Background color for verbatim text in paragraph
Is it currently possible to define the background color of a whole paragraph with verbatim text (see `pre` tag in "Config Files" of Zim's manual) in it?

As far as I can see there is only the `background` property for a tag, which only changes the background color of a word ([example](https://i.imgur.com/agcevJT.png)) and not of the whole paragraph.

Ideally the colored area should have the width of the longest sentence written in verbatim text and not extend till the scroll bar by default.

If not, then such an option would be very useful!

stephaneseng:
## Problem

Follow-up to [#512](https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/512).

## Proposed resolution

The style.conf file is used to customize the text appearance in the main Gtk.TextView, by customizing the properties of the Gtk.TextTags of its Gtk.TextBuffer.

Among the Gtk.TextTags properties listed in [GtkTextTag documentation](https://developer.gnome.org/gtk3/3.18/GtkTextTag.html), the "[paragraph-background](https://developer.gnome.org/gtk3/3.18/GtkTextTag.html#GtkTextTag--paragraph-background)" one can be used to define the background color for a whole paragraph.

As stated in the documentation, this property was available since Gtk+ 2.8, and can now be used, now that zim targets [Gtk+ >= 3.18](https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/0.72.0/README.md#installing-from-source).

## Screenshot

Using the default [style.conf](https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/0.72.0/data/style.conf), and by adding `paragraph-background = rgba(1, 1, 1, 0.05)` to the `[Tag pre]` section as shown below:

```
[Tag pre]
# This style is used for multi-line verbatim blocks
foreground = #2e3436
paragraph-background = rgba(1, 1, 1, 0.05)
family    = monospace
wrap-mode = none
indent    = 20
```

We have the following result:

Screenshot-After


sojusnik:
Works great, thanks!

Is it possible that the paragraph background colour has only the width of the longest sentence within a paragraph and does not extend until the right scroll bar?

stephaneseng:
@sojusnik: Thanks for your feedback.

Unfortunately this doesn't seem to be possible, as it looks like none of the properties listed in the [GtkTextTag documentation](https://developer.gnome.org/gtk3/3.18/GtkTextTag.html) could be used to achieve this effect.

Issue #255, opened 2017-12-08
Add option to web server dialog to select template


jhojen26:
How to select a theme for the web server, just run me with the one that comes by default.

thank you very much

Osndok:
This is for the built-in web server, as indicated in 254... there is no option to select a template, or the like.

![no-template-selection](https://user-images.githubusercontent.com/24486935/33765624-bf179a64-dbe7-11e7-958a-964e4d71740a.png)

jaap-karssenberg:
Indeed the GUI does not have this option, should be added. When starting from commandline options you can specify the template.

PR #987, opened 2019-12-17
Add template option to web server dialog #255

stelzch:
Simple fix for Issue #255


stelzch:
Accidentally overwrote the fork with a different fix for #999

PR #1004, opened 2020-01-14
Text wrap mode properties (updated)

aaferrari:
Adapted the commit that I had left in [this pull request](https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/403) for version 0.72.1, also update #401 to demonstrate how it works with GTK3.
I am sorry that it has taken so long, for most of 2019 I have been busy with other things and I have not been able to deal with this.


jaap-karssenberg:
patch looks OK but test failed due to missing documentation - could you fix that?

aaferrari:
That should work, I'm not sure if the new options are well explained so any correction is welcome.
Sorry I didn't notice that before, I'm not very familiar with those testing tools and I don't usually pay attention to what they say.

PR #1008, opened 2020-01-19
Macos integration

brechtm:
Update the macOS Menubar plugin for the current version of Zim.

### DIFF
--- a/zim/gui/uiactions.py
+++ b/zim/gui/uiactions.py
@@ -272,7 +272,7 @@ class UIActions(object):
 		from zim.gui.templateeditordialog import TemplateEditorDialog
 		TemplateEditorDialog(self.widget).run()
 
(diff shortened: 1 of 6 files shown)


zim/gui/uiactions.py
```diff
@@ -272,7 +272,7 @@ def show_templateeditor(self):
 		from zim.gui.templateeditordialog import TemplateEditorDialog
 		TemplateEditorDialog(self.widget).run()
 
-	@action(_('Pr_eferences')) # T: Menu item
+	@action(_('Pr_eferences'), '<Primary>comma') # T: Menu item
```
brechtm: You might not want this accelerator for platforms other than macOS. I added the accelerator there because I didn't manage te add it in the macOS Menubar plugin.

PR #1029, opened 2020-03-06
Give temporary directories unique names

### LINKED ISSUE #1028: Predictable /tmp/zim-USER directory allows for denial of service
**Description**
Zim creates temporary directories with predictable names. A malicious user could predict and create Zim's temporary directories and prevent other users from being able to start Zim, resulting in a denial of service.

**Steps to Reproduce**
1. Create two users: user1 and user2.
1. As user1:
    `mkdir /tmp/zim-user2`
1. As user2:
    Start Zim

Zim will attempt to change the permissions of /tmp/user2 to 0700. Because /tmp/user2 is actually owned by user1, the call to `os.chmod()` will fail and Zim will crash.

**Workaround**
An affected user can set the $TMP or $TMPDIR environment variables to point to somewhere that is not world-writable (e.g. `/home/USER/tmp`).

**Version Affected**
0.42-beta1 and later

**Operating System**
Linux

**Severity**
Low

mssalvatore:
Fixes #1028

### DIFF
--- a/zim/fs.py
+++ b/zim/fs.py
@@ -18,6 +18,7 @@ import os
 import re
 import sys
 import shutil
+import tempfile
 import errno
 import logging
 
@@ -120,26 +121,12 @@ def get_tmpdir():
 	Used as base folder by L{TmpFile}.
 	@returns: a L{Dir} object for the zim specific tmp folder
 	'''
-	# We encode the user name using urlencoding to remove any non-ascii
-	# characters. This is because sockets are not always unicode safe.
 
-	import tempfile
-	root = tempfile.gettempdir()
-	user = url_encode(os.environ['USER'], URL_ENCODE_READABLE)
-	dir = Dir((root, 'zim-%s' % user))
+	if get_tmpdir.dir is None:
+		get_tmpdir.dir = Dir(tempfile.mkdtemp(prefix='zim-'))
 
-	try:
-		dir.touch(mode=0o700) # Limit to single user
-		os.chmod(dir.path, 0o700) # Limit to single user when dir already existed
-			# Raises OSError if not allowed to chmod
-		os.listdir(dir.path)
-			# Raises OSError if we do not have access anymore
-	except OSError:
-		raise AssertionError('Either you are not the owner of "%s" or the permissions are un-safe.\n'
-			'If you can not resolve this, try setting $TMP to a different location.' % dir.path)
-	else:
-		# All OK, so we must be owner of a safe folder now ...
-		return dir
+	return get_tmpdir.dir
+get_tmpdir.dir = None
 
 
 def normalize_file_uris(path):


mssalvatore:
> Why so complicated? Using [`tempfile.mkdtemp`](https://docs.python.org/3.6/library/tempfile.html#tempfile.mkdtemp) the whole function can be as simple as:
> 
> ```python
> if get_tmpdir.dir is None:
>     get_tmpdir.dir = tempfile.mkdtemp(prefix='zim-')  # with permissions 0700
> 
> return get_tmpdir.dir
> ```
> 
> Am I missing something?

I was trying to leave as much of the existing code intact. `tempfile.mkdtemp()` is a better solution, but all of the code is expecting a Dir object, not just a path:

```
if get_tmpdir.dir is None:
    get_tmpdir.dir = Dir(tempfile.mkdtemp(prefix='zim-'))  # with permissions 0700

return get_tmpdir.dir
```

But for some reason this causes some of the tests to fail and I haven't gotten around to digging into that yet.

hartwork:
> [..] but all of the code is expecting a Dir object, not just a path:

Good point!

> But for some reason this causes some of the tests to fail and I haven't gotten around to digging into that yet.

There is all sorts of crazy stuff going on during `Dir.__init__`:
https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/7d7d1364dc74c892bc81ef1399957406d1be4c61/zim/newfs/base.py#L230-L238

I would try using a _tuple_ (as done in the existing code) — i.e. `Dir((tempfile.mkdtemp(prefix='zim-'),))` — and see if that changes anything. If not, I'd be curious about the details of the test failure.

mssalvatore:
@hartwork, Making it a tuple seemed to work. I've amended the commit.

hartwork:
The new `Dir((tempfile.mkdtemp(prefix='zim-')))` lacks a comma though to make it a tuple: `('foo')` versus `('foo',)` — string versus 1-tuple with a string.

mssalvatore:
Weird. I don't know what was failing in the tests, then. I removed the parentheses an all of the tests passed. ¯\_(ツ)_/¯

It didn't make sense to me that the string would be an issue in the first place because `Dir.__init__` takes the same code path for tuples and strings.

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/7d7d1364dc74c892bc81ef1399957406d1be4c61/zim/newfs/base.py#L230-L231

hartwork:
> [..] `Dir.__init__` takes the same code path for tuples and strings.
> 
> https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/7d7d1364dc74c892bc81ef1399957406d1be4c61/zim/newfs/base.py#L230-L231

This code further in is only run for strings but not a list of strings:
https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/7d7d1364dc74c892bc81ef1399957406d1be4c61/zim/newfs/base.py#L123-L139

PR #1047, opened 2020-04-11
Updated translation skeleton (xgettext.sh)

fhaftmann:
This is a formal update of the translation skeleton using the tools/xgettext.sh, updating references and inserting so far missing messages to translate.

With that, missing messages to translate will appear as such in launchpad.


jaap-karssenberg:
Sorry, but no need for a pull request for that - I'll update as part of the release process

PR #1051, opened 2020-04-15
distraction free editing plugin: remove all colors of the window chrome 

rscmbbng:
The box colors are set to the same value as the background color for true distraction free editing.

Regular view: 

Distraction free before:

Distraction free after patch:

### DIFF
--- a/zim/plugins/distractionfree.py
+++ b/zim/plugins/distractionfree.py
@@ -67,10 +67,17 @@ class DistractionFreeMainWindowExtension(MainWindowExtension):
 			color: %s;
 			background-color: %s;
 		}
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
Thanks for the patch!   Would it make sense to add a separate preference
for the background color to allow different background colors for the text
and the box if a user prefers that ?

-- Jaap

rscmbbng:
Personally I don't think so. Do you prefer Zim to have more toggles/customization or more 'sane defaults?' An option could be easily added, but I think most distraction-free writing modes have a single background color and that is more in the line of expectations.

Another thing is: given a single background+box color, the option 'Border width' essentially becomes a top/bottom margin height option.

jaap-karssenberg:
Merged in develop

PR #1053, opened 2020-04-21
fixes handling of non-existent pages

bytebieger:
When the name of a non-existent page is changed, the pages that link to this non-existent page don't get updated.

Also fixes moving of non-existent pages.


jaap-karssenberg:
Can you also provide a test case as part of the patch?  That way I can
refactor the code without breaking your fix.

Regards,

Jaap

bytebieger:
Here you have the test case.

Regards,
BB

jaap-karssenberg:
Thanks!  Will merge in develop branch.

PR #1073, opened 2020-05-05
diagram default text

e3rd:
Inserting quick diagram is very handy feature. Unfortunately, we could not set up default text so that we have write every time the same line again and again.

Now we may have ex: 'digraph G {rankdir=LR}' as the base.

### DIFF
--- a/zim/plugins/diagrameditor.py
+++ b/zim/plugins/diagrameditor.py
@@ -27,6 +27,11 @@ This is a core plugin shipping with zim.
 		'author': 'Jaap Karssenberg',
 	}
 
+	plugin_preferences = (
+		# key, type, label, default
+		('default_text', 'string', _('Default text'), ''),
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
Did the test suite pass for this one? I would expect failure due to "default_text" preference missing in the manual page.

tridian-tn:
I was thinking yesterday that this would be a good feature to have, and then I found this PR. 
@e3rd if you haven't already done so, would you mind if I took this on and wrapped up the manual changes etc required to get the test suite to pass, if it doesn't already?

e3rd:
Oh, please, do it, thanks. I've already completely forgotten about this PR, blame on me.

tridian-tn:
Cheers

introt:
Closing in favor of the new PR

Issue #730, opened 2019-04-01
Keybinding to copy line


Rafal1:
Minor issue but extremely useful. Could you tell me if is it a convenient way to copy a whole line of text in zim?

jaap-karssenberg:
No such keybinding is implemented at the moment.

sojusnik:
The line sorter plugin does this. You can configure the keyboard shortcut in ~/.config/zim/accelmap

jaap-karssenberg:
It has a "duplicate line" - I read the request as a "copy line to clipboard"

sojusnik:
@jaap-karssenberg True!

Self-Perfection:
@Rafal1 though there is no "single shortcut" general text editor navigation works. So you can press:
Home -> Shift+End -> Ctrl-C

Or with mouse: triple click on any word in line to copy -> right click -> "copy"

Rafal1:
Thank you for your answers, the issue was "copy line to clipboard" :).

PR #1078, opened 2020-05-10
Add a key binding for copying the current line

eightytwo:
This change addresses issue #730 which proposes a new keybinding to
copy the current line. I'm not sure if this change is desired by many,
or if the code achieves this in a nice way, but it was a good exercise
in gaining some familiarity with the Zim codebase.


jaap-karssenberg:
Actually quite a clean implementation as you did it. Will merge in the "develop" branch

Issue #752, opened 2019-04-03
Problem with copying a header from zim to clipboard


tehkonst:
When I copy a part of a header from zim, and paste it somewhere else there are a lot of "=====" characters in the pasted text.

For example: 

> Header============

(I copied this text from zim to the addressbar of the browser)

sojusnik:
See also https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/661

From my point of view headers should be treated like normal text – if "copy as plain text" is set, headers should also be copied as plain text, without being an exclusion.

tehkonst:
The opinion of the author is strange to me. I often use programs for translating texts from one language into another and usually they translate text from clipboard. So it's no good when text comes with formatting symbols.

jaap-karssenberg:
If anyone is willing to work on a patch we can discuss.

jaap-karssenberg:
* Re-open #661, plan for making plain real plain - rename "plain" to "simple" and add real "text" as default
* Make parser round trip for both "simple" and "text"
* New item to make parser for e.g. pasted context more intelligent and also understand wiki and markdown like text

PR #1082, opened 2020-05-11
Exclude formatting when copying headings

eightytwo:
This addresses issue #752, which concerns the behaviour of copying
headings in plain text.

Prior to this change, copying a heading in a notebook and pasting it
into another text editor would include the formatting. For example, the
level three heading `Title` would be pasted into a plain text document
as `### Title`.

This is inconsistent with the copying and pasting of text with other
formatting, such as bold or italics. A bold word, such as `test`, is
pasted as `test`, not `**test**`.

This change also affects the plain text export, meaning that headings
are not exported with any formatting.

### DIFF
--- a/tests/data/formats/plain.txt
+++ b/tests/data/formats/plain.txt
@@ -1,5 +1,4 @@
 Head 1
-======
 Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
Will merge this patch.

Looking at the code, the only remaining formatting in the text format are tables. Shouldn't those be exported text only as well for consistency?

Issue #668, opened 2019-02-11
Show failed plugins in plugin list in preferences dialog


jaap-karssenberg:
When a plugin fails to load, it will not be shown in the plugin list in the preferences dialog. Errors are printed to error log, but not always visible to the user. Therefore it would be better to show all plugins in the plugin list. If plugin modules fail to load, show the error where normally the description is (with some explanation).

eightytwo:
@jaap-karssenberg: I have a couple of questions about the solution for this issue that I was hoping you might be able answer.

1. What sort of helpful error message can be shown to the user as to why a particular plugin failed to load? As far as I can tell a plugin will fail to load if an exception occurs [here](https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/7d7d1364dc74c892bc81ef1399957406d1be4c61/zim/gui/preferencesdialog.py#L370-L373), in which case there will only be an exception message available which may not be helpful to the user. That said, an exception message could be helpful for reporting issues to this repo.

2. Should the error message be shown for plugins that do load but have missing dependencies? Zim already displays missing dependencies in red so perhaps this isn't necessary. If it would be useful the error message can include this detail about missing dependencies.

PR #1086, opened 2020-05-16
Show plugins that failed to load in the list

eightytwo:
This change addresses #668, which requires failed plugins to be
displayed in the list of plugins. This is only a minor improvement
which:
* includes a plugin that failed to load in the list
* displays a generic error message in the details pane
* uses the plugin's key (file or directory name) in the list if the
  plugin's proper name can't be read due to the load problem

The generic error message is also displayed at the top of the details
pane for plugins that can't be activated due to missing dependencies.
This may not be desired or it might be useful to state that loading
failed due to missing dependencies (marked in red in the
dependencies list).

### DIFF
--- a/zim/gui/preferencesdialog.py
+++ b/zim/gui/preferencesdialog.py
@@ -296,6 +296,12 @@ class PluginsTab(Gtk.VBox):
 				buffer.insert_at_cursor(text)
 
 		buffer.delete(*buffer.get_bounds()) # clear
+
+		if not activatable:
+			insert(_('This plugin failed to load') + '\n\n', 'red')
+			if not klass:
+				return
+
 		insert(_('Name') + '\n', 'bold') # T: Heading in plugins tab of preferences dialog
 		insert(klass.plugin_info['name'].strip() + '\n\n')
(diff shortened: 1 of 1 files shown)


eightytwo:
I just saw the thumbs up reactions so have marked this PR as ready for review. Thanks @jaap-karssenberg and @e3rd.

jaap-karssenberg:
THink this one needs a small change still. The "activatable" property is now used for both plugins that miss dependencies and plugins that completely fail to load. Would be good to distinguish these in how they are presented to the user.

eightytwo:
@jaap-karssenberg: An additional sentence is now displayed when dependencies are missing, drawing the user's attention to the dependencies section of plugin view. Is this what you had in mind or would you prefer the presentation to be different?

Two scenarios are shown below.

**Missing required dependency**


**Exception when loading the plugin in `PluginsTreeModel.__init__`**

pfalcon:
Regarding last screenshot with "There was a problem loading this plugin" - I'd find that alone useless. Show me the actual problem happening.

Issue #393, opened 2018-03-18
Annoying auto-saving when leaving Zim window


CharlesNepote:
As discussed on the mailing list -- https://lists.launchpad.net/zim-wiki/msg04602.html -- I'm facing an annoying functionality/behaviour (while not so important): when I write in Zim window and **leave it for another window**, Zim **auto-save** the current content, even if I told him to save every 10 minutes in the config file.
It's annoying in my case because I synchronize Zim with other machines thanks to Seafile. The issue is perhaps also annoying with version control plugin.

From my point of view, it should be an **option which can be enable/disable**.

luffah:
Not (currently) a Zim developer, 
but it imply at least three modifications :
  - add the option, (and enclose each save procedure in code by this new condition)
  - add a popup on Zim leave (one button for each file saving , or one button for saving all files ?)
  - add  indicators in the file list  to be notified the file is not saved (because when you switch from a note to another one , it save it too) 

Personally, i find the autosave feature usefully when you have a lot of notes. (even if, most of the times, i use Vim-Zim, not directly Zim)

Osndok:
I heavily rely on the version control plugin, have my repo shared across several machines, and have no issues related to the current auto-save implementation. IMO, the real issue here is that you are trying to press zim into a low-latency collaboration mechanism (like a shared whiteboard or etherpad), and I don't think this will be your last problem going down that road. I imagine that you are also likely to get silently dropped modifications (esp. with simultaneous writes to journal pages), and ever more so as the number of people sharing the zim repo increases or (**as you are recommending**) the longer unsaved pages linger in RAM.

jaap-karssenberg:
See https://lists.launchpad.net/zim-wiki/msg04605.html

The logic should be refined to check the timer and only do the save if e.g. within 80% of the time to next save. Result would be that 10 minutes becomes somewhere between 8 and 10 minutes.

luffah:
Just for my own understanding. When will the file be saved in the following usecase ?
- t 0 : I configure the parameter on zim to 10 min. 
- t 0bis : I edit a note, it take 7 min.
- t 7 : I switch to pidgin window to say to a comrade "hey look on the shared zim !"
- t 7bis : I go sleepy on peertube... waiting for comments.
- t 17 : 10 minutes later,  I switch back on  zim window, worrying about misspelling.

jaap-karssenberg:
In the current behavior it will save the moment you switch focus away from
the editing window.

[...]

PR #1090, opened 2020-05-19
Add an option to disable autosave on focus_out in preferences.conf

alterlife:
Possible fix for #393 , this reads an option from preferences.conf under GtkInterface overrides the default foucus_out behavior. Autosave on focus out is enabled by default (as before), and is disabled if explicitly turned off in the configuration file using this option.

Another small thing I noticed is that the default autosave time out here seems to be 15 minutes, while the wiki says that the default is 10 minutes.


jaap-karssenberg:
Rather than an on/off preference, I would like to add a hidden preference for the minimum duration between saves and make it such that it does not save more often than that.
This way the issue is addressed and adds more control on the functionality than the simple on/off.

PR #1092, opened 2020-05-20
Changed my name after marriage

fmos:
Corrected my Name in all instances. Not sure if this is usual practice, but seems sensible to me.


jaap-karssenberg:
Up to you - it is your name after all ;)  Will merge before release. And
congratulations !

-- Jaap

fmos:
Thanks! :)
Fabian

PR #1122, opened 2020-06-15
Show link destination for keyboard navigation fixes #452, closes #707

### LINKED ISSUE #452: Show link destination for keyboard navigation

e3rd:
Implemented via custom mouse-independent signals. Jaap, do you think it is the right way to handle it like this?


jaap-karssenberg:
Yes, this looks pretty clean. Thanks!

PR #1123, opened 2020-06-15
paste short relative link names fix #100

e3rd:
I'd like to have shorter links from Journal. So I have added new notebook property that sets whether an internal wiki link should be pasted in shorter form. This was initially PR'd in #100 and now I've re-implemented as a generic option as Jaap suggested years ago.


e3rd:
I beg an explanation why this has been closed without a comment. I'm trying to push the feature since 2017. Have I not followed your guidelines in #100 ?

jaap-karssenberg:
If you read closely, it is closed *with a commit number* which means that
commit founds it's way to a release - thanks for contriuting :)

e3rd:
Wow, sorry that I've missed this, that's an honour, thanks a lot <3

Issue #1182, opened 2020-07-14
_checkargs prepends interpreter to argv for external commands when argv[0] ends with .py


nbeaver:
When opening an external URL, Zim runs the `spawn` function which call the `_checkargs` function.  If the first argument ends in `.py`, the `_checkargs` function may insert the python interpreter at the beginning of `argv`:

    def _checkargs(self, cwd, args):
        assert args is None or isinstance(args, (tuple, list))
        argv = self._cmd(args)
    
        # Expand home dir
        if argv[0].startswith('~'):
        cmd = File(argv[0]).path
        argv = list(argv)
        argv[0] = cmd
    
        # if it is a python script, insert interpreter as the executable
        if argv[0].endswith('.py') and not _main_is_frozen():
        argv = list(argv)
        argv.insert(0, sys.executable)
        # TODO: consider an additional commandline arg to re-use compiled python interpreter
    
        if hasattr(cwd, 'path'):
            cwd = cwd.path

        return cwd, argv

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/35679abcc52b55b4fc8045016de30d331633cf96/zim/applications.py#L194

This means that `args` such as

    ['example.py', 'example-arg1']

will be expanded to an `argv` of

    ['/usr/bin/python3', 'example.py', 'example-arg1']

Normally this is not a problem, since e.g. `http://` URLs are handled by `firefox` and `file://` URLs are handled by `xdg-open`, neither of which end with `.py`, but a URI scheme handler that is a Python script might, e.g. `my_svn_scheme_handler.py`.  In this case, if the script is in `$PATH` and so is not an absolute path, it will result in an error such as

    /usr/bin/python3: can't open file 'my_svn_scheme_handler.py': [Errno 2] No such file or directory
    ERROR: Could not open: svn://example-svn-url

One workaround is to just use the full path in the desktop file:

    [Desktop Entry]
    Name=SVN scheme handler
    Type=Application
    TryExec=/path/to/my_svn_scheme_handler.py
    Exec=/path/to/my_svn_scheme_handler.py %u

to get

    ['/usr/bin/python3', '/path/to/my_svn_scheme_handler.py', 'svn://example-svn-url']

However, this will not work if there is a mismatch in interpreters, e.g. if Zim is running `python3` and the script needs `python2` (or vice versa).

The relevant commit appears to be to allow running custom commands on Windows:

> Small fix for executing zim from source on windows

[...]

PR #1183, opened 2020-07-14
Remove the prepended interpreter.

nbeaver:
The Windows version of [Python 3.3 and later](https://docs.python.org/3/using/windows.html#python-launcher-for-windows) has native shebang support via [PEP 397](https://www.python.org/dev/peps/pep-0397/), so this mechanism is no longer necessary. Relevant bug report:

https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/1182

### DIFF
--- a/zim/applications.py
+++ b/zim/applications.py
@@ -201,12 +201,6 @@ class Application(object):
 			argv = list(argv)
 			argv[0] = cmd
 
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
Please make sure the test suite passes as well

PR #1199, opened 2020-07-20
Dark mode for equations

### LINKED ISSUE #1143: Equations hard to read in dark theme
All pictures are with `Arc-Dark` theme from [the Arc theme](https://github.com/jnsh/arc-theme), but the results are equally bad with the `Breeze-Dark` from the Breeze theme.

Here's the manual page, which is practically illegible.
image
And here's a random equation from my wiki page. This one is at least legible, but it's not great.
image

Maybe some sort of white drop-shadow, an option for a lighter/custom foreground color, or even just white/lighter text color would fix this issue.

---

System info
Arch Linux kernel 5.7.5
Theme Arc-Dark (icon theme Arc)
Zim version 0.73.0

stelzch:
Fixes #1143

### DIFF
--- a/data/manual/Plugins/Equation_Editor.txt
+++ b/data/manual/Plugins/Equation_Editor.txt
@@ -5,9 +5,9 @@ Wiki-Format: zim 0.4
 
 The equation editor is a simple dialog that allows you to insert equations into a page using **latex**.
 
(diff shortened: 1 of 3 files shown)


jaap-karssenberg:
Thanks !  Patch looks good.

PR #1223, opened 2020-08-09
Update markdown.py

Atlas48:
This pull request outlines two changes:

1. The de-facto filename extension for markdown files has changed from `.markdown` to `.md`.
2. As per [[RFC7763](https://www.iana.org/go/rfc7763)], Markdown has the default mime type of `text/markdown`


Atlas48:
This is also a minor code change.

Atlas48:
No idea why the build's failing, by the way.

jaap-karssenberg:
Master is failing as well, I need to push a fix for the test suite

PR #277, opened 2017-12-31
Load images in a thread

benzea:
I just had a quick look a loading images from a thread. I also addressed at least one crash that may happen when threading is enabled, I am not sure if there may be further errors where other workarounds might be necessary.

Anyway, take it/rework it as you like.

PR #1236, opened 2020-08-19
Use Pillow fallback to load more image formats

adynathos:
I made a change allowing Zim displays my WebP images, and I want to share it with the community.

WebP is a modern lossy image format offering superior quality and lower file size compared to JPEG.
GTK is unable to load WebP images https://gitlab.gnome.org/GNOME/gtk/issues/1403
but Pillow is capable of loading a variety of image formats https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html
With this change, when GTK fails to load an image, we let Pillow try instead.

In this version it is optional, and does something only if Pillow happens to be installed.

I added a test notebook file with an image, but I'm not sure how to make an automated test about the image being visible.

I have seen some interest in [loading images in a thread](https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/277), but threads combined with GTK image loading caused a crash. This problem could be avoided if Pillow loads on the worker thread but all the GTK calls are on the main thread.


adynathos:
Thanks for including this!

Issue #978, opened 2019-11-26
Allow to create link inside the current page


Arcesilas:
I think link inside a page could be useful sometimes. When the page is rather long, with many titles, it would be nice to be able to add a link in the text to a section of the page.
The Table of Content can scroll the page to a given point, but if there are many titles, it's much easier to just click on the text you are reading rather than search for the title in the ToC. Also, a link allows to have a text different from the title.
Using a `#` as first character of the link would indicate it's an anchor, like in HTML.

Nick222:
But what about link to certain title in TOC ?

Arcesilas:
I don't understand what you mean... Maybe you mean the same as I meant in my initial post...?

Nick222:
Then I do not understand - what it is " link inside a page" - link from ??? to ???

Arcesilas:
It's like in HTML : `<a href="#anchor">` a link inside the page. The ToC already provides this, I think it would be nice to be able to create the same links inside the page, wherever we need it.

cwross:
If we could treat headings as anchors and link to them like [[foo:bar#heading]], that would be fantastic.

stevphie:
Simple but neat feature...

pdwilso:
I've wondered about this a bit, too - does Markdown have this feature?

Arcesilas:
Markdown itself does not have this feature, because it's just a formatting syntax. Some converters extensions allow this, like https://github.com/webuni/commonmark-attributes-extension for example.
Most of the time, you need to manipulate the AST produced by the markdown parser, to add IDs on headings, for example.

PR #1293, opened 2020-10-18
Support for having anchors in wiki pages

roland-ruedenauer:
This is meant to be a starting point supporting links inside pages as requested in #978 . The changes allow creating links to headings as anchor point. Clicking onto a section heading offers 'Copy link to section" to ease the link creation process.

### DIFF
--- a/data/style.conf
+++ b/data/style.conf
@@ -41,6 +41,8 @@ foreground = blue
 [Tag tag]
 foreground = #ce5c00
 
+[Tag anchor]
+foreground = #6b017d
 
 # For the various headings we scale the size relative to the font size.
 # The color choosen is the dark green -or "Chameleon"- as defined in the
--- a/tests/clipboard.py
+++ b/tests/clipboard.py
@@ -229,6 +229,20 @@ some <b>bold</b> text
 		newtree = Clipboard.get_parsetree(self.notebook, Path('Test'))
 		self.assertEqual(newtree.tostring(), wanted)
 
+	def testCopyPageLinkWithAnchorPasteAsParseTree(self):
+		page = self.notebook.get_page(Path('Test:wiki'))
+		Clipboard.set_pagelink(self.notebook, page, 'anchor')
+		wanted = '''<?xml version=\'1.0\' encoding=\'utf-8\'?>\n<zim-tree><link href="+wiki#anchor">+wiki#anchor</link></zim-tree>'''
+		newtree = Clipboard.get_parsetree(self.notebook, Path('Test'))
+		self.assertEqual(newtree.tostring(), wanted)
+
+	def testCopyPageLinkWithAnchorAndTextPasteAsParseTree(self):
+		page = self.notebook.get_page(Path('Test:wiki'))
+		Clipboard.set_pagelink(self.notebook, page, 'anchor', 'My anchor')
+		wanted = '''<?xml version=\'1.0\' encoding=\'utf-8\'?>\n<zim-tree><link href="+wiki#anchor">My anchor</link></zim-tree>'''
+		newtree = Clipboard.get_parsetree(self.notebook, Path('Test'))
+		self.assertEqual(newtree.tostring(), wanted)
+
 	def testCopyPageLinkPasteAsText(self):
 		page = self.notebook.get_page(Path('Test:wiki'))
 		Clipboard.set_pagelink(self.notebook, page)
--- a/tests/errors.py
+++ b/tests/errors.py
@@ -105,7 +105,7 @@ class TestExceptionHandler(tests.TestCase):
 
 			# Should log one error message with traceback
 			self.assertEqual(len(records), 1)
-			self.assertEqual(records[0].getMessage(), 'Test Error')
+			self.assertEqual(records[0].getMessage(), _('Test Error'))
 			self.assertEqual(records[0].levelno, logging.ERROR)
 			self.assertIsNotNone(records[0].exc_info)
 		else:
@@ -123,7 +123,7 @@ class TestExceptionHandler(tests.TestCase):
 
 			# Should log one error message with traceback
 			self.assertEqual(len(records), 1)
-			self.assertEqual(records[0].getMessage(), 'Looks like you found a bug')
+			self.assertEqual(records[0].getMessage(), _('Looks like you found a bug'))
 			self.assertEqual(records[0].levelno, logging.ERROR)
 			self.assertIsNotNone(records[0].exc_info)
 		else:
--- a/tests/formats.py
+++ b/tests/formats.py
@@ -702,6 +702,23 @@ hmmm
 		t = self.format.Parser().parse([text])
 		self.assertEqual(t.tostring(), xml)
 
+	def testAnchor(self):
+		text = '{{#test}}'
+		xml = '''\
+<?xml version='1.0' encoding='utf-8'?>
+<zim-tree><p><anchor name="test">test</anchor>
+</p></zim-tree>'''
+		tree = self.format.Parser().parse(text)
+		self.assertEqual(tree.tostring(), xml)
+
+	def testAnchorWithText(self):
+		text = '{{test: some text}}'
+		xml = '''\
+<?xml version='1.0' encoding='utf-8'?>
+<zim-tree><p><anchor name="test">some text</anchor>
+</p></zim-tree>'''
+		tree = self.format.Parser().parse(text)
+		self.assertEqual(tree.tostring(), xml)
 
 class TestGFMAutolinks(tests.TestCase):
 	# See https://github.github.com/gfm/#autolinks-extension-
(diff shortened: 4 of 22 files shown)


pfalcon:
The title of this PR is confusing:

> Allow creating links inside the current page

It's very possible to create links inside the current page (obviously, everything is created *inside* a page, including links).

jaap-karssenberg:
Thank you for looking into this!  Sorry for not commenting earlier.

Couple of things I noticed for improving the patch:
* Would appreciate more unit testing of the newly introduced functions, to make sure there are no bugs and - more importantly - someone else doesn't break them with other patches
* When generating the anchors, would prefer the heuristics of how the anchor ID is generated to be compatible with how github and/or pandoc do it. Not sure which rule the patch uses now, would be good to explicitly document in the user documentation (manual) and have a separate unit tests covering these rules
* In the API I would prefer using "anchor" or "id" instead of "section" e.g. using `navigate_to_anchor()` instead of `navigate_to_section()` and `open_page(anchor=...)` instead of `open_page(section=...)` - the rationale being that we could later allow for other elements (figures, tables, objects) in a page that have an anchor id besides headings

Btw. as background information, my general design ideas for anchors are here: https://github.com/zim-desktop-wiki/zim-desktop-wiki/wiki/BlueprintTagsAndAnchors

Kind regards and thanks again,

Jaap

roland-ruedenauer:
Thanks for the detailed response.

The patch as it is supports creating links to headings of the form [[foo:bar#heading]] as wished in #978 and does not create actual anchor objects in the internal page representation. The result closely matches what is described in https://pandoc.org/MANUAL.html#extension-implicit_header_references .

Being able to create links to page headings is ok, but my real goal was being able to place "real" anchors into a page as you lined out in the linked blueprint document. Don't know why, but before your comment I just found the rather old and short note written down in https://github.com/jaap-karssenberg/zim-wiki/wiki/Blueprint-anchors ;-)

Now I'll have a closer look if your newer blueprint solves the issues I initially saw having support for anchors integrated into the existing wiki syntax. How would headings with explicit anchors look like was one of the questions I had.

A question that immediately occurred to me was how to visually represent anchors in the rendered page? Having them shown as text like tags would be easiest, I guess. But visually more appealing might be an option to have them being somehow hidden (although they need to be accessible in some way). Any ideas regarding this?

roland-ruedenauer:
The last commit adds this functionality:
- Adding anchors on a page with auto-formatting (regex for anchors is   r"#\w[\w_-]+")
- Support for page links containing an anchor
- Implicit page anchors for headings, where the anchor name is derived from the heading text
- Copying anchor links into the clipboard via context menu (just right-click on a heading or anchor)

I tried following the ideas provided in the [blueprint document](https://github.com/zim-desktop-wiki/zim-desktop-wiki/wiki/BlueprintTagsAndAnchors) authored by Jaap.

Test coverage still needs to be improved.

jaap-karssenberg:
I like what you are doing here - looks much better already. 

Looks like you deviated from the blueprint with respect to the wiki syntax. What I wrote there is to use "`#name`" to create a link and "`##name`" to create an anchor. Also I would really like explicit anchors to be inserted as "`{{#name}}`" instead of just "#name" since the later potentially conflicts existing page content and also because it is nicely symmetric with links written as "`[[#name]]`".

Did you consider this? If so, what is the rationale to go for the different format?

Regards,

Jaap

roland-ruedenauer:
The latest update should fix the deviations from the blueprint.

Inserting links to page local anchors via auto-formatting is now supported too.
Next I'll have a look into adding support for linking to anchors in images and objects.

Some notes about anchor support
- Not quite sure about the visual representation of anchors. In the current implementation they're displayed in purple color to make them distinguishable from regular text.
- The wiki syntax {{#name}} for anchors does not allow having a display text in addition to the anchor name.

roland-ruedenauer:
New additions
* Added support for linking to images and inserted objects via an "id" attribute.
* Modified the SourceView plugin insertion dialog to allow entering the anchor name (value of the "id" attribute).

For this PR I don't plan to add more changes besides fixes / small improvements.
It would be helpful if someone could test the current state.

pfalcon:
I'm just now manually inserting github-markdown compatible ToC to my Zim page, and figured I could use this change. (Yeah, by now using Zim to author content, and Github to publish, of course, all relies on https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/26#issuecomment-688178350)

So, the first thing I look at is the commit message(s) (not PR description, because that doesn't update automatically with the commits' evolution, and humans forget to update it manually). And what I see here is:

> Support having anchors in wiki pages

That's all, really? I'd expect for such a change, there to be 2-3KB of text, describing what is being done in that commit, and superseding any previous information available (because any previous information is already outdated, and only the actual commits contain the actual state of affirs, and should also contain the actual, and detailed, description).

pfalcon:
> Implicit page anchors for headings, where the anchor name is derived from the heading text

I was particularly interested in *how* it's derived.

roland-ruedenauer:
@pfalcon The implicit anchor name for a heading is derived by calling the function heading_to_anchor() in zim/formats/__ini__.py and currently converts the heading text to lower case, substitutes blanks by a hyphen and removes all non-word characters besides hyphen and underscore. This may not work well for some languages and might require being able to add explicit (random) anchor names to headlines.

The blueprint section https://github.com/zim-desktop-wiki/zim-desktop-wiki/wiki/BlueprintTagsAndAnchors#anchors describes what this PR is trying to implement.

roland-ruedenauer:
Deviations from https://github.com/zim-desktop-wiki/zim-desktop-wiki/wiki/BlueprintTagsAndAnchors#anchors:
* Locating an anchor on a page is simplified: explict anchors are searched first followed by implicit anchors. There's no hierarchy of anchors taken into account - the first matching anchor found an a page is returned.

Currently not supported:
* explicit anchors for headlines

Worth mentioning:
* Creating a link to an anchor using the context menu and inserting the link into the current notebook should preserve the anchor text. This is useful for links created from implicit anchors like headlines.

smac89:
Awesome work guy. Keep going!

roland-ruedenauer:
Added in the last commit: the dialog for editing image properties now offers entering an anchor name.

If an attached image has been assigned an anchor, the context menu shown when clicking on the image contains an entry to copy a link to this anchor.

I would like to also have support for explicit anchors on headings. But then, how should they be represented in the wiki syntax and what would be a reasonable way to assign or modify an anchor to a heading?

roland-ruedenauer:
"Open in New Window" now correctly works for page links with anchors. Previously the anchor part has been ignored.

jaap-karssenberg:
Working on merging this branch compared to current develop branch - thanks for putting in the effort to make the patch so complete !

jaap-karssenberg:
@roland-ruedenauer: can you explain why you implemented achors having both a name and a text element, what is an example use case where this would be useful to have?

Reading my own blueprint, I see the example of "{{id: name}}" syntax. I see how that is ambiguous - what I meant was the "id:" prefix to be literal and the "name" part to be variable. Other object types could use the same template with another prefix than "id" - e.g. "{{equation: ...}}" for inline equations etc.

If there is a good use case to have text as well, the template should probably be extended to accommodate that as well.

jaap-karssenberg:
PS. This is not a request to change the patch, I'll pick up any modifications as I already resolved conflicts in my merge branch

jaap-karssenberg:
On the topic of explicit anchors for headings, my thinking is that just adding an anchor tag on line of the heading should do the job. Wil have a look at the styling, but should not be too special.

In that case we could indeed use the heading text for copy pasting, but no need for the anchor tag itself to have a text field.

roland-ruedenauer:
The blueprint being ambiguous was the main reason to add support for having a display text with anchors. Testing and playing with anchor support in my test notebooks also made me think it really is a good idea because

* it allows to change the visual representation without breaking any anchor links
* an anchor id is restricted to conform to a required pattern - having a display text option adds some flexibility

With the explanations you provided, switching to a different inline syntax would probably make sense.

Regarding support for explicit anchors for headlines it is more the question of how to extend to UI in a reasonable way to allow adding/modifiying anchors. Expecting users to switch to wiki syntax seems not to be adequate.

Let me finally remind you of a question about visual representation of anchors I raised initially

> A question that immediately occurred to me was how to visually represent anchors in the rendered page? Having them shown as text like tags would be easiest, I guess. But visually more appealing might be an option to have them being somehow hidden (although they need to be accessible in some way).

If anchors would have no prominent visual representation (besides for maybe a small graphical hinting) and could be made visible on demand to allow creating links to anchors etc., then there would probably be no need to have a display text option. Still I am undecided if this would be a preferrable and not too hard to implement solution compared to what we have now.

jaap-karssenberg:
A pattern I see on several websites is to use a "pilcrow" sign to represent
an anchor - see e.g. the python docs like os — Miscellaneous operating
system interfaces — Python 3.9.4 documentation
<https://docs.python.org/3/library/os.html> for an example.

Within the pageview I could use the same mechanism as for checkboxes to
insert a icon of such a pilcrow sign and use a pop-up bubble to show and
modify the anchor id and trigger a "copy link" action

Does that feel like a good way to represent anchors ?

-- Jaap

jaap-karssenberg:
P.S. reading the wikipedia article on the "pilcrow" sign I even found the
use for anchor links as a recommended usage for web publishing, with a RFC
document to back it up ... so this seems the way to go

skome:
Look forward to this functionality!

roland-ruedenauer:
@jaap-karssenberg are you working on applying the changes suggested while merging the existing PR or are you waiting for me here? I am a bit confused how to go on to get the functionality integrated soon.

jaap-karssenberg:
I will pick up the changes. Cleared a whole list of other PR's this week.
Now will work on the remaining items for the 0.74 release including this
one.

Regards,

Jaap

PR #1310, opened 2020-11-02
Little fix to relative pathname computation.

a-sennov:
relative_filepath is broken on Windows because it always tries to use "/" as separator character. This is the fix.

### DIFF
--- a/zim/notebook/notebook.py
+++ b/zim/notebook/notebook.py
@@ -1054,7 +1054,7 @@ class Notebook(ConnectorMixin, SignalEmitter):
 		notebook_root = self.layout.root
 		document_root = LocalFolder(self.document_root.path) if self.document_root else None# XXX
 
-		rootdir = '/'
+		rootdir = SEP
 		mydir = '.' + SEP
 		updir = '..' + SEP
 
@@ -1078,7 +1078,7 @@ class Notebook(ConnectorMixin, SignalEmitter):
 				parent = file.commonparent(attachments_dir)
 				uppath = attachments_dir.relpath(parent)
 				downpath = file.relpath(parent)
-				up = 1 + uppath.count('/')
+				up = 1 + uppath.count(SEP)
 				return updir * up + downpath
 		else:
 			if document_root and notebook_root \


pfalcon:
@jaap-karssenberg: Hopefully, this won't be accepted, and you keep sticking to the statement: "Relative file paths are always given with Unix path semantics (so "/" even on windows)." https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/master/zim/notebook/notebook.py#L1040

It also should be noted that "/" is *the* path separator for Windows. (Only some broken Windows software insist on proprietary "\" path separator).

a-sennov:
"Relative file paths are always given with Unix path semantics (so "/" even on windows)." - in fact, this is not the case as OS-specific separator characters are in use almost everywhere. You may see it in saved pages on different OSes.
If at some point in time the mandatory conversion to system-independent '/' will be implemented, this patch will not contradict with that, as SEP variable will simply always be '/' then.

pfalcon:
> - in fact, this is not the case as OS-specific separator characters are in use almost everywhere.

There's garbage on streets almost everywhere, so what? It shouldn't be there. Again, Windows does support "/" as a path separator char. ZimWiki core (and notebook.py is the code) uses "/" as its separator char. So, this patch doesn't look to be either correct way to do it, or needed at all. **Especially** that it says that it's a "little fix", but doesn't say what problem it actually fixes.

Anyway, just my 2 cents, I'm a random user of Zim, interested in a clean codebase.

a-sennov:
Hmm... I thought description is descriptive enough. OK: method relative_filepath in Notebook class is broken. At least on Windows (I suspect on Mac too) it doesn't work as intended and that leads to multitude of problems. To verify those problems it's enough to try to drop file from one page's attachment onto another page of different depth level of notebook hierarchy (on Windows). Probably you never worked with complex notebooks with many branches and levels. The cause of errors is insane mix of usage of literal "/" and SEP (currently holding system-dependent path separator character) variable inside method's implementation.
The fix is little because it touches only 4 lines of code to clean the pile of garbage that was happily laying here for at least three years. It converges to use only SEP variable as this is the safest way to go without touching FS codebase.
PS. Windows supporting / as path separator has nothing to do with this problem - it's bug in Zim itself.

jaap-karssenberg:
By accident I already fixed this in another commit allowing both separators to work.

Issue #418, opened 2018-04-09
When cut/pasting diagrams (graphviz) into another page the images show as missing


ziggystar:
### Actions 

1. cut multiple diagrams from a page
2. paste into another page

### Result

- The diagrams are missing from the first page `p1` (expected)
- The diagrams show up as missing-image symbols in the destination page `p2`
- The dot-source cannot be edited or accessed in the `p2` (no option in context menu)
- The dot-source cannot be accessed in `p1` (items have been cut)

This is effectively data loss for the user.

### Workaround

I could repair the situation in the file-system, by creating a new attachment directory for `p2` and copy the respective files from the attachment directory of `p1`. This is quite annoying, if you have many dot files in `p1` and you have to sort out which ones are the copied ones.

While trying to fix this, I also found it difficult to associate the `*.dot` files with the figures within the page: #419

jaap-karssenberg:
For current objects this will be "won't" fix as neither solution for #45 or #46 will solve this in a generic way.

Planning to make these objects inline rather than based on external sources, that makes this easier to fix

jaap-karssenberg:
To be solved by implementing #841

pfalcon:
> To be solved by implementing #841

Even if this particular case (image-generator plugins) will be solved by #841 (which IMHO will be sad, as I explained in it), there still remains the case of the normal images inserted into page (aka image attachments) themselves. #1312 was posted which tries to solve that in a particular (among at least 2 possible) ways.

PR #1312, opened 2020-11-02
Fix for copy/paste images between pages

a-sennov:
Copying/pasting image links between pages doesn't work when pages belong to different branches or depth levels of notebook hierarchy. Relative links to images are transferred verbatim without proper resolution.

Here is the fix to the problem.

### DIFF
--- a/zim/formats/__init__.py
+++ b/zim/formats/__init__.py
@@ -469,6 +469,16 @@ class ParseTree(object):
 				filepath = element.attrib['src']
 				element.attrib['_src_file'] = notebook.resolve_file(element.attrib['src'], path)
 
+	def relative_images(self, notebook=None, path=None):
+		'''Counterpart to resolve_images
+		'''
+		if notebook is not None:
+			for element in self._etree.iter('img'):
+				if '_src_file' in element.attrib:
+					filepath = element.attrib['_src_file']
+					element.attrib['src'] = notebook.relative_filepath(File(filepath), path)
+					element.attrib.pop('_src_file')
+
 	def unresolve_images(self):
 		'''Undo effect of L{resolve_images()}, mainly intended for
 		testing.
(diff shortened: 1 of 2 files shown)


pfalcon:
There's a known issue related to these matters, https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/418, so thanks for looking into it.

However, the description of this PR is a bit light:

> Relative links to images are transferred verbatim without proper resolution.
> Here is the fix to the problem.

It says what's wrong currently, but doesn't say explicitly how it's going to be fixed and **why** (among the different possible solutions). It seems that solution being adopted is that image files should remain where they are, but links to them in a new page should be updated to point to the old image files (the links should still be relative within notebook).

But that's only one way to address it. It seems that currently following is invariant (?) in Zim:

* For any page, page's attachments (including images) are stored in a directory side by side with that page's file. (For `./Page.txt`, attachments are in `./Page/`).

If that's really the invariant, then the proper solution would be to copy/move image files to the attachment folder of the new page, not update the links in the new page.

All in all, both these options (and maybe more) should be considered, before going for a particular solution (worst case, both may need to be supported, letting user to make a choice between the two).

a-sennov:
Pages may be seen as containers for attachments, so yes, the operations on pages must be reflected to attachments too. But the scope of this PR is operations on images, not pages or even links to attachments of other types (although I aware that those links do suffer the same problem when copied or moved). The reason for this is that images are special objects: they use their own syntax, have their own representation in ZIM data model. Images on page are not actual files but always are links, or in other words - relations. So when being manipulated, actual image data should never be touched.
One day, if time will permit, I'll probably address this problem by more general approach so links to any object inside notebook will be handled the same way but for now this fix is only for images.
Another interesting area is how operations on pages affect the links to their attachments situated in other pages. But again, this is well out of scope of this PR.

jaap-karssenberg:
I'm working on an alternative patch for this. Fix as proposed can serve as
a work around but I'm afraid it will fail in a number of cases.

PR #1313, opened 2020-11-03
Fixes to TableOfContents plugin

a-sennov:
These commits make small fixes to ToC plugin:

1. Elimination of annoying exception polluting the log each time new page is opened. Actually it seems more like a workaround as tree data model is set only once at widget construction time and theoretically should never become None, Maybe clearing the model is implemented as setting it to None or something.
2. Fix incorrect sizing of ToC in floating mode. When complex page gets loaded (by complex I mean - more than 3 headings inside) ToC treeview fails to recompute it's preferred size in sync with tree data model changes. That leads to shortened ToC (it may be made to recompute size by hand by minimize/maximize window or clip by another windows etc). Commit fixes this by forcing all tree entries to expand via TreeView method call, so this is more of workaround too. I suspect that's more of Gtk+ issue when somehow TreeView and TreeModel get out of sync during model manipulations but actually I won't to dive in that muddy waters.

### DIFF
--- a/zim/plugins/tableofcontents.py
+++ b/zim/plugins/tableofcontents.py
@@ -541,7 +541,7 @@ class FloatingToC(Gtk.VBox, ConnectorMixin):
 			handler=DelayedCallback(10, self.update_size_and_position),
 				# Callback wrapper to prevent glitches for fast resizing of the window
 		)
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
For the AttributeException, can you give instructions how to duplicate it
?  Might be helpful to actually find the cause and fix based on that
understanding.

-- Jaap

Issue #1347, opened 2020-12-27
[Proposal] attachmentbrowser button to copy path to clipboard


jkooij:
## Summary
Add a menu button to the attachment browser to copy the current attachment folder path to the clipboard.
I've already created a patch and will add a pull request.

## Use case
One of my most common use cases for Zim is to collect and organize related document as attachments to a page. E.g. often I get new document versions send via email, or I download pdfs via the browser, and I want to store these together with the relevant page in my notebook. In all these cases, I need to store the documents from some external GUI program, and I want to save them into the directory created by Zim's attachment browser. This requires copying the attachment folder path to the "Save As" popup window of the external program.

## Before proposed change
My usual workflow to save a document to the attachment folder is as follows:
1. Start external program X's option to export the document(s), e.g. press some "Save As" button
2. X opens a "Save As" popup window to select the destination path. Often this popup is a native OS window, and enables the user to navigate to a desired folder, but this takes a lot of clicking and searching. The popup often also has an address bar.
3. Click on Zim's attachment browser's "Open Folder" menu button (and use option to create it if it doesn't exist). This opens the folder in the native file explorer.
4. In the file explorer, select the address bar containing the full directory path
5. press Ctrl-A to select the full path
6. press Ctrl-C to copy to the clipboard
7. Click on address bar of the SaveAs popup from program X
8. press Ctrl-V + Enter to replace the path with the folder from the attachment browser
9. Press Enter or the Save button in the popup window to store the document in the attachment folder
10. Close the opened file explorer window

This workflow is not very efficient. Notice that the file explorer was only opened as a means to get the folder path in the clipboard, and should be closed in the final step to avoid cluttering my work environment.

## After proposed change
I propose to add a third button "Copy path" to the attachment browser. This would **replace steps 4, 5, 6 and 10, and the need to wait for a file explorer window to open and close.**

1. Start external program X's option to export the document(s), e.g. press some "Save As" button
2. X opens a "Save As" popup window to select the destination path. Often this popup is a native OS window, and enables the user to navigate to a desired folder, but this takes a lot of clicking and searching. The popup often also has an address bar.
3. Click on Zim's attachment browser's "Copy Path" menu button (and use option to create it if it doesn't exist). This copies the folder path to the clipboard.
4. Click on address bar of the SaveAs popup from program X
5. press Ctrl-V + Enter to replace the path with the folder from the attachment browser
[...]

PR #1348, opened 2020-12-27
Add menu item to attachment browser to copy path to clipboard

jkooij:
Implement a copy path button in the attachment browser, see issue #1347


jaap-karssenberg:
See the issue for comments

PR #1349, opened 2020-12-28
souceblock: Feature - add highlightjs compatibility

lggomez:
I wanted to implement code highlightning so after checking the codebase I found that at some point there was an attempt to provide compatibility with highlight.js as an alternative for class generation on the html code blocks so the formatter can pick them. The decoupling between the template engine and the plugin itself makes this difficult to implement it in one pass because having highlighted blocks requires specific codes in the </head> section of the implementers only (namely, the highlight.js imports and load code)

Also, another issue (which is solved in the PR) is the mapping between the GTKSource and highlightjs language ids. An intermediate mapping is done between both to solve this

In this PR I propose the following

- Add a checkbox to decide whether to use highlight.js classes on the HTML generated code or not
- Add a mapping GTKSource and highlightjs language ids to solve the internal inconsistencies

#### Open questions / TODOs 

- Since the template has to explicitly have the link refs, does this call for a fork of ZeroFiveEight with said functionality as a default template or is there another (cleaner) solution?
- Can we default to plaintext for HTML generation if no highlighting is selected? to my understanding there is no code formatter that uses the GTK ids, although the correct matches may be several for common language ids like c and javascript

For a live demo you can see the following:
- https://lggomez.github.io/Software_Development/Misc/Test.html
- Fork of ZeroFiveEight with the relevant changes: https://github.com/lggomez/zim-templates-gh/blob/main/templates/html/ZeroFiveEight_luisgg.html#L13-L24

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,7 @@ ENV/
 # debian pybuild
 .pybuild/
 debian/python-module-stampdir/
+
+# IDEs, editors
+.idea/
+.vscode/
--- a/tests/sourceview.py
+++ b/tests/sourceview.py
@@ -135,6 +135,10 @@ class TestSourceViewObject(tests.TestCase):
 		item.activate()
 		self.assertEqual(widget.buffer.get_language().get_name(), 'Python')
 
+		item = tests.gtk_get_menu_item(menu, _('Use highlightjs classes'))
+		item.activate()
+		self.assertEqual(widget.buffer.get_use_highlightjs(), True)
+
 	def testDumpHtml(self):
 		xml = '''\
 <?xml version='1.0' encoding='utf-8'?>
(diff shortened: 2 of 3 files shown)


lggomez:
I could use some help with the following test assertion: https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/1349/files#diff-5d930fdc3dd4ef6742c9102d8c4b44c64e8e2a4f7bc5ac74cf33b3aaabb97cfdR138-R140

For some reason I don't get (I'm not familiar with python or GTK) the assertion helper can't find my ui control as registered:

> AssertionError: Menu item "Use highlightjs classes" not found, we got:
> - _Undo 
> - _Redo 
> -  
> - Syntax 
> - Show Line Numbers 
> -  
> -  
> - C_hange Case

jaap-karssenberg:
Thanks for proposing this feature - makes sense to me. On the implementation, I'm doubtful a user would want to see an attribute for this per code block. Rather I expect it to be a template property. Linking it together, there should be a template property for a choice of using the highlightjs language codes (default=yes ?) and this property is than passed to "object.dump()" in order to get the code block serialized with this property.

PR #1367, opened 2021-01-17
Added new context menu entries in attachmentbrowser for renaming and deleting files

datenwurm:
Adds two new context menu entries in attachmentbrowser for renaming and deleting files.

unnamed

## Rename
The rename context menu entry will open a dialog in which a new name can be entered. 

There are a few validation rules in place so not any arbitrary name can be choosen:
* new name should not be empty
* new name should not already exist
* new name should not contain any path information

When validation fails the error is shown in the dialog.

image

## Delete
When clicking the delete context menu entry a confirmation dialog is shown. Since it is not always clear which file is actually selected, the name of the file is shown. 

image

### DIFF
--- a/zim/plugins/attachmentbrowser/__init__.py
+++ b/zim/plugins/attachmentbrowser/__init__.py
@@ -7,6 +7,7 @@
 # !! NOTE: when changing this plugin, do test performance on a folder with lots of photos!
 #
 # ChangeLog
+# 2021-01-16 Added rename and delete functionality to file browsers context menu (Thomas Engel)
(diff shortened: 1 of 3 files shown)


jaap-karssenberg:
Thanks for the patch - actually looks quite good already !

Had a quick look at the patch
- please consider basing your code on the "newfs" module - will refactor
out the old "fs" soon
- have you considered using trash instead of hard remove ?
- looks like the "path_leaf" function does the same as "file.basename" or
do I miss something ?

datenwurm:
Thanks for your comment! I've added the changes as suggested. If there's anything else please let me know!

Issue #953, opened 2019-10-14
Table Editor: Cannot insert newline without editing source


oturpe:
Zim version: 0.72.0
Operating system: Fedora 30
Language: fi-FI

Table Editor documentation page [Superior Table Operations](https://zim-wiki.org/manual/Plugins/Table_Editor/Superior_Table_Operations.html) says a line break can be added to a table cell by writing `\\n` in the textbox. However, this does actually result in text _\\\\n_ shown, with no line breaks. Similarly, writing `\n` results in text _\\n_. Checking the generated page source, these are written as `\\\\n` and `\\n`. 

Editing the source directly and writing simple `\n` actually does produce a line break. This is the only workaround I have been able to find.

Expected behavior: There is a way to add a line break when using Table Editor and it is correctly documented. Since the documentation already says the way is `\\n`, I guess that would be it, but also `\n` would do as well.

I would be great to be able to enter a line break with Ctrl+Enter, like many other editors do in context of a table cell. But that is a somewhat different story.

This issue seems to be related to closed issue #467.

Nick222:
Zim 0.72
Xubuntu 19.04

arigit:
Same issue, zim 0.73.3 / MacOS

SirBigoo:
Same issue, zim 0.72.0 / Ubuntu 20.04

mohd2126:
Same issue
Zim 0.77
Windows 10

PR #1370, opened 2021-01-18
fix(tableeditor): multiline text in table cell

SirBigoo-zz:
Fix issue #953 : now, zim detects line return on table cell. `\n` is used in cell value to handle multiline

### DIFF
--- a/tests/tableeditor.py
+++ b/tests/tableeditor.py
@@ -34,7 +34,7 @@ TABLE_WIKI_TEXT = '''\
 |        H1       <|         H2 h2 | H3                    <|
 |:----------------:|--------------:|:-----------------------|
 |    Column A1     |     Column A2 | a \\\\name               |
-| a very long cell | **bold text** | two\\nlines             |
+| a very long cell | **bold text** | two<br/>lines          |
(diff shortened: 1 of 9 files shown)


jaap-karssenberg:
Looking at the changeset I don't believe this is a complete working patch - no parser code changed ?

As to the solution direction, I don't think it is necessary to add an "<br>" code to the wiki format, rather the parsing of the "\\n" needs to be fixed

SirBigoo:
Thanks for feedback ! 
I searched to modify parser at first but I didn't find when it's triggered. Not sure to understand all structure and how it works.

On table editor, the only occurrence to parser that I found is `WikiParser`. It calls `parse` function but it seems it never triggers when editing page or exporting (HTML).

If you have any clue about this, I can go further and fix this :)

jaap-karssenberg:
Yes the "parse_table" method in the WikiParser is where a table get read
from wiki format. It is indeed not called while editing, only when reading
the page. With export it is called to read as well, but only if the page
was not already loaded in memory.

For dealing with newlines check specific the use of "unescape_string" -
that methods translates "\\n" into a proper newline character.

Regards,

Jaap

PR #1374, opened 2021-01-22
sourceblock: Feature - Combobox in dialog with autocomplete 

datenwurm:
The current implementation of sourceview opens up a dialog when the context menu entry "Insert > Code Block" is triggered.
This dialog has a combobox where the syntax/language can be selected. The combobox items are grouped by the first letter of the syntax/language. When a syntax/language is selected and the ok-button is pressed the dialog closes and the code block is added to the current page. If the dialog is opened a second time, the previous entry is automatically selected. 

This implementation adds three new features:
* the combobox has now a autocomplete feature which allows to select entries by typing part of their name. The previous behaviour (grouping by first letter) is still present.
* the ok button is now the default button. When the combobox has focus (default) the ok-button can be activated by pressing enter.
* the ok button gets automatically disabled when an invalid syntax is selected.

Standard Behaviour:
image

Autocomplete Feature:
image

I would love to see this change merged into the main branch :)

### DIFF
--- a/zim/plugins/sourceview.py
+++ b/zim/plugins/sourceview.py
@@ -282,49 +282,77 @@ class InsertCodeBlockDialog(Dialog):
 		self.result = (None, None)
 		self.uistate.define(lang=String(None))
 		self.uistate.define(line_numbers=Boolean(True))
-		defaultlang = self.uistate['lang']
-
-		menu = {}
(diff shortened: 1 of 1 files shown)


m8:
waiting for that feature

PR #1376, opened 2021-01-23
Plugin: Dash

datenwurm:
This is my proposal for adding the Dash plugin.

The Dash plugin opens a search dialog via a keybinding to allow quickly executing menu entries by typing part of their name.

image

When enter is being pressed and a valid entry is selected the dialog is closed and the entry is executed.
When reopening the Dash dialog the last entry used is selected by default.  

The dialog can be opened by pressing alt+x on the keyboard. However, the keybinding can be changed via "Edit > Preferences > Keybindings".

By default the Dash dialog shows the associated keybindings to each entry. However, this can be disabled via the plugin options.

image


pfalcon:
Why this niche, rarely-to-be-used feature should be a part of the main distro? (name alone is confusing, to the best of my knowledge, "dash" is an implementation of unix console shell). Who's going to support it? (Starting with answering questions "wtf it's named dash?").
 
> Transforms tabs to spaces

Zim codestyle exactly uses spaces instead of tabs, in proud violation of PEP8. (https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/master/CONTRIBUTING.md#getting-started-with-the-code)

datenwurm:
Hi pfalcon, sorry for the confusion. The dash-plugin was already discussed in the zim mailing list. This merge request is just a result of @jaap-karssenberg stating that he would love to include this plugin in the main branch. Regarding the name - it is based on the Unity Dash (see https://askubuntu.com/questions/299295/what-is-the-dash). Regarding "Transforms tabs to spaces". As you will see when taking a look at the code - i transformed from spaces to tabs. So it's just the other way around and i just stated it wrong in the commit message which should have been saying "Transforms spaces to tabs".

pfalcon:
Sounds good then, please consider the above just as an opinion of a random Zim user who's concerned about wading thru the long list of builtin plugins, some of which aren't descriptively named and some potentially buggy (so will take resources from the maintainer to support, away from the core, which has its bugs too) ;-).

datenwurm:
Previously the keyboard shortcut names were not shown correctly (e.g. "<Primary>c" instead of "Ctrl+c").
I also removed the plugin option to show/hide keyboard shortcuts since the behaviour now looks like i intended.

image

datenwurm:
@jaap-karssenberg: What do you think about merging this plugin into the main branch?

jaap-karssenberg:
@datenwurm  Thanks for submitting this, it's a feature I had on my personal whishlist for quite a while :)  and apologies for the long timeout - will add this to my merge list for the next upcoming release

One question: I see the keybinding for the dash is Alt-x -- is there "prior art" to support this choice ?  I'm biased myself by using the Atom editor for code, which has a feature like this under Ctrl-Shift-P

Atom calls the the "Command Palette" which might be more explanatory than "dash", and maybe we can think of a even better descriptive name ?

jaap-karssenberg:
Did a quick internet search for "command palette" and this turns up quite a
few results for similar features in common code editors. All/most of these
seem to have adopted the Ctr-Shift-P keybinding. So will move ahead merging
this plugin using that name and keybinding.

Hope this is OK with you.

Regards,

Jaap

datenwurm:
Hi Jaap, happy to hear this getting merged into the upcoming release! Regarding the name. Your right, "Command Palette" seems to be a much better name and i also think ctrl-shift-p will be a good fit as a keybinding (as long as the keybinding can be changed. i've just fallen into the habit of pressing alt-x all the time :D)

jaap-karssenberg:
The test case is failing because there is no manual page for this plugin.

Ideally we should also have a specific test case for the plugin itself. Just run the dialog in a test and check it runs without errors would be a good start.

jaap-karssenberg:
P.S. Since the action is "accelonly" and does not show in any menu, would be good to mention the default keybinding in the info text of the plugin in the plugin list as well as in the manual page - else a user may never find out how to open the command palette

datenwurm:
Hi Jaap, thanks for the feedback. I'm quite busy this weekend but i'll provide an update regarding the missing manual page and the specific test case for the plugin in the upcoming week.

jaap-karssenberg:
No hurry, release is still few weeks out at least. And I can also touch the
manual page when merging.

Op vr 23 apr. 2021 23:43 schreef datenwurm ***@***.***>:

datenwurm:
Hi Jaap, Unfortunately I've found a bug in the commandpalette plugin and i'm currently not sure how to fix it.

1. Open Zim
2. Select Text (multiple lines with each line containing multiple words)
3. Open Command Palette plugin
4. Enter "Format strong" and press Ok

Only the first word in the first line is rendered bold. I'm not quite sure why this is happening since the plugin only triggers the Gtk.MenuItem.activate -- so there should not be any difference between manually selecting an Gtk.MenuItem via the menu.

Do you have any idea what might cause this issue?

jaap-karssenberg:
No idea directly, but I'll have a look when I merge the patch

jaap-karssenberg:
@datenwurm merging the plugin now on top of the development branch, but not able to reproduce the bug you mention

PR #1390, opened 2021-02-07
Update README for XDG paths when using Python virtual environment

tai271828:
By explicitly elaborating the XDG path setup for people launching their
zim in a Python virtual environment, we will help them to launch their
zim easier without confusion.

Closes #1381

PR #1393, opened 2021-02-07
Add a new test case of TestXDGDirs

tai271828:
This commit is an enhancement of test case testAllValid by checking
system XDG_DATA_DIRS directly. The new test case will help users catch
issues like #1390 .

### DIFF
--- a/tests/config.py
+++ b/tests/config.py
@@ -180,6 +180,7 @@ class TestXDGDirs(tests.TestCase):
 				self.assertEqual(getattr(zim.config.basedirs, k), list(map(Dir, v.split(':'))))
 
 	@tests.skipIf(os.name == 'nt', 'No standard defaults for windows')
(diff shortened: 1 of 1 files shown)


tai271828:
This pull request wants to check the current running XDG environment, so the failure is expected for the lack of XDG environment in travis CI working space. Let me think about how to fix the false alarm.

tai271828:
This pull request is updated to pass CI and ready for reviewing.

Please note this pull request correlates another pull request https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/1390 .

jaap-karssenberg:
Not sure whether I get how this works. If I run the test before installing zim, doesn't it always fail?

tai271828:
> Not sure whether I get how this works. If I run the test before installing zim, doesn't it always fail?

Nice catch @jaap-karssenberg ! The test case indeed assumes Zim is installed. I will make it also work for the scenario when Zim is not installed yet. Maybe just skipping the test if these is no zim launcher in PATH. Let me think about it.

PR #1420, opened 2021-03-13
Add buttons to traverse through the Journal forwards and backwards

bitmuster:
Hi, here is a small enhancement to add little buttons to traverse through the Journal forwards and backwards easily.
I hope you like it:
image

BTW I see some failing tests, but they don't seem to be related...

### DIFF
--- a/tests/applications.py
+++ b/tests/applications.py
@@ -272,7 +272,10 @@ class TestApplicationManager(tests.TestCase):
 	def testSupportDesktopMimeappsList(self):
 		orig_desktop = os.environ.get('XDG_CURRENT_DESKTOP')
 		def restore_desktop():
-			os.environ['XDG_CURRENT_DESKTOP'] = orig_desktop
+			if orig_desktop:
+				os.environ['XDG_CURRENT_DESKTOP'] = orig_desktop
+			else:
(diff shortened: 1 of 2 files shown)


bitmuster:
Sorry, one commit slipped in while rebasing from master to develop. It's reverted now

jaap-karssenberg:
Hi Michael, I like the idea, 2 suggestions for improvement

1) instead of using buttons with text "<" and ">" it looks better with buttons containing an icon, try `Gtk.Button.new_from_icon_name()` with "pan-start-symbolic" and "pan-end-symbolic" 

2) when the journal is configured to use weeks instead of days, it would be nice if the buttons skipped a whole week rather than a day at a time - you can check the "granularity" property which can be day/week/month/year and use that as a timedelta

Regards,

Jaap

abelikt:
Hello Jaap, thanks a lot for the feedback and the suggestions. I will come up with a solution soon

bitmuster:
Hello Jaap, I implemented your proposals and updated the pull request. Please have a look if it is good now.
In case that one selects years or months as granularity, we would use days again. There are specific buttons for that granularity, so I'm not sure if we should fall back to days here.

PR #1430, opened 2021-03-26
Fix warning about not having called gi.require_version

dehesselle:
This PR fixes the following warning:
> WARNING: Zim.app/Contents/Resources/lib/python3.8/site-packages/zim/plugins/osx_menubar.py:31: PyGIWarning: GtkosxApplication was imported without specifying a version first. Use gi.require_version('GtkosxApplication', '1.0') before import to ensure that the right version gets loaded.
> from gi.repository import GtkosxApplication

### DIFF
--- a/zim/plugins/osx_menubar.py
+++ b/zim/plugins/osx_menubar.py
@@ -25,8 +25,10 @@ from zim.gui.mainwindow import MainWindowExtension
 from zim.plugins import PluginClass
 
 try:
+	import gi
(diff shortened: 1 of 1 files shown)


dehesselle:
I have noticed the checks are failing, but I'm inclined to say "works as designed", as to what other outcome can be expected when testing macOS-only plugins on Linux. If there is a way to make this work, tell me.

dehesselle:
I figured it out. Kind of "not seeing the trees in the forest" problem.

PR #1453, opened 2021-04-25
Hotfix numbered bullets replacing existing bullets

### LINKED ISSUE #1452: Regression in autoformat in 0.73: Bullets and checkboxes are replaced with letters and numbers
#### Bullets and Latin

1. Type `* a. ophthalmica` (shorthand for "artery").

Expected result: `* a. ophthalmica`
Actual result: `a. ophthalmica`

#### Checkboxes and (non-list) numbers

1. Type `[] 1. palkinto` ("first prize" in Finnish).

Expected result: `[ ] 1. palkinto`
Actual result: `1. palkinto`

The same results is achieved with all letters (regardless of case) and (non-negative) numbers regardless of indentation or surrounding lines. This behavior is unaffected by the "Reformat wiki markup on the fly" option in Preferences > Editing. The behavior was not present in version 0.72.0 (installed from Ubuntu repositories), where the resulting wiki markup was as expected.

### Affected versions

New behavior introduced with commit 0108edc5. Tested with version 0.73.5 built from commit ec8dc363.

### Suggested fix

#1453 and #1454 - don't replace existing bullets with numbered bullets.

### Related issues

[Possibility to disable numbered lists](https://bugs.launchpad.net/zim/+bug/1239324)

introt:
Fixes #1452

### DIFF
--- a/tests/pageview.py
+++ b/tests/pageview.py
@@ -2437,6 +2437,13 @@ class TestDoEndOfWord(tests.TestCase, TextBufferTestCaseMixin):
 		self.buffer.place_cursor(iter)
 		self.assertTyping('* ', '<li bullet="*" indent="0"> Test</li>')
 
+	def testNoAutoFormatNumberedWithinList(self):
+		self.set_buffer(self.buffer, '<li bullet="unchecked-box" indent="0"> Test</li>')
+		iter = self.buffer.get_start_iter()
+		iter.forward_chars(2) # put it behind the checkbox
+		self.buffer.place_cursor(iter)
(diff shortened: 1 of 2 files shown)


introt:
This keeps the new behavior where it makes sense eg. with bullets and checkboxes replacing numbered bullets but not the other way around, making it possible to [format multiple lines with dates or names as a bullet list](https://bugs.launchpad.net/zim/+bug/1239324) just like before the change.

introt:
Oops, no need for more code when the fix is taking out the other side of that `or`. Will redo against develop branch first and then fix this one. Sorry!

PR #1456, opened 2021-05-01
Remove trailing **** from first paragraph

introt:
- added in 13d8c46b (I thought it might be a footnote or similar, but there isn't one)
- not present in the develop branch where 13d8c46b isn't merged

Two questions for maintainers:
- For the record, are this kind of small fixes worthy of PRs or should I gather up a few? I consider this one to be due to its visibility.
- As I skimmed the README for other stray markup, I noticed data/{urls,symbols}.list aren't listed under files included from other sources. Should they be?


jaap-karssenberg:
> For the record, are this kind of small fixes worthy of PRs or should I gather up a few? I consider this one to be due to its visibility.

Sure, I have no problem with small PR's - they are easy to review and merge. Especially when it is about documentation - keep them coming ;)

PR #1473, opened 2021-05-27
Add TRANSMIGRATED_BOX to where it's missing from

introt:
I left `tests/data/notebook-wiki.xml` alone as changing it didn't seem to impact the tests and I wasn't sure how to handle `<` being interpreted as a start of an XML tag - should it be `html.escape()`d or is there another way?

### DIFF
--- a/data/menubar.xml
+++ b/data/menubar.xml
@@ -39,6 +39,7 @@
 				<menuitem action='toggle_checkbox'/>
 				<menuitem action='xtoggle_checkbox'/>
 				<menuitem action='migrate_checkbox'/>
+				<menuitem action='transmigrate_checkbox'/>
 			</menu>
 			<menuitem action='edit_object'/>
 			<menuitem action='remove_link'/>
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -2411,6 +2411,7 @@ class TextBuffer(Gtk.TextBuffer):
 				CHECKED_BOX
 				XCHECKED_BOX
 				MIGRATED_BOX
+				TRANSMIGRATED_BOX
 		or a numbered list bullet (test with L{is_numbered_bullet_re})
 		'''
 		iter = self.get_iter_at_line(line)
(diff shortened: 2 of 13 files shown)


zim/gui/pageview/__init__.py
```diff
@@ -6162,11 +6165,13 @@ def do_mark_set(self, buffer, iter, mark):
 			self.actiongroup.get_action('toggle_checkbox').set_sensitive(True)
 			self.actiongroup.get_action('xtoggle_checkbox').set_sensitive(True)
 			self.actiongroup.get_action('migrate_checkbox').set_sensitive(True)
+			self.actiongroup.get_action('transmigrate_checkbox').set_sensitive(True)
```
introt, line 6168: This doesn't fix it not showing up in the menu, but doesn't hurt either

zim/gui/pageview/__init__.py
```diff
@@ -6476,8 +6481,8 @@ def _checkbox_do_populate_popup(self, menu, buffer, iter):
 		menu.prepend(Gtk.SeparatorMenuItem())
 
 		for bullet, label in (
```
introt, line 6483: The popup menu is populated in reverse order, which thus corresponds to the toggle order after this change:

```
[ ] 0
[*] 1
[x] 2
[>] 3
[<] 4
```

introt (commented):
.

zim/gui/pageview/__init__.py
introt: See https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/1332#issuecomment-849820609 for context

data/menubar.xml
```diff
@@ -39,6 +39,7 @@
 				<menuitem action='toggle_checkbox'/>
 				<menuitem action='xtoggle_checkbox'/>
 				<menuitem action='migrate_checkbox'/>
+				<menuitem action='transmigrate_checkbox'/>
```
introt, line 42: There it is, the missing link! Sometimes you've just gotta `grep` harder

introt: I typo'd the commit msg, but the code is ok & works

introt:
Will rework this mess into something nicer & easier to review.

Issue #1463, opened 2021-05-13
Tables (table plugin) are always 100% wide


vboehm:
If I want to create a small table
| employee | payment |
| -------- | --------: |
| Miller | 2500 |
| Jones | 3300 |

the resulting table always uses the entire width
~~~
employee                                                                               payment
Miller                                                                                    2500 
Jones                                                                                     3300
~~~

but it should only use as much width as it needs
~~~
employee  payment
Miller       2500 
Jones        3300
~~~

Could you use this natural way as the default or at least provide a choice (per table) if 100% are requested?

As you can see, this wiki (and nearly every other) uses the correct way :-)

PR #1474, opened 2021-05-28
Fix 100% wide table issue (#1463)

VivekThazhathattil:
modified:   zim/plugins/tableeditor.py
- By default, InsertedObjectWidget's expand flag is set to True, and
  that causes the abnormal table width. Changed the flag to False
  from within the tableViewWidget class to avoid that issue.

- set toolbar to adjust to its natural size than the existing size of
  (300, -1) as whenever the toolbar comes into view, last column sizes
  are annoyingly readjusted.

### DIFF
--- a/zim/plugins/tableeditor.py
+++ b/zim/plugins/tableeditor.py
@@ -323,6 +323,7 @@ class TableViewWidget(InsertedObjectWidget):
 
 	def __init__(self, model):
 		InsertedObjectWidget.__init__(self)
+		InsertedObjectWidget.expand = False;
 		self.textarea_width = 0
 		self.model = model
 
@@ -476,7 +477,7 @@ class TableViewWidget(InsertedObjectWidget):
 				button.set_tooltip_text(tooltip)
 				toolbar.insert(button, pos)
 
-		toolbar.set_size_request(300, -1)
+		toolbar.set_size_request(-1, -1)
 		toolbar.set_icon_size(Gtk.IconSize.MENU)
 
 		return toolbar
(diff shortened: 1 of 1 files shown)


zim/plugins/tableeditor.py
```diff
@@ -323,6 +323,7 @@ class TableViewWidget(InsertedObjectWidget):
 
 	def __init__(self, model):
 		InsertedObjectWidget.__init__(self)
+		InsertedObjectWidget.expand = False;
```
introt, line 326: No need to call InsertedObjectWidget, `self.expand = False` works

zim/plugins/tableeditor.py
```diff
@@ -506,6 +507,8 @@ def create_treeview(self, model):
 			cell = Gtk.CellRendererText()
 			tview_column = Gtk.TreeViewColumn(headcol, cell)
 			tview_column.set_sizing(Gtk.TreeViewColumnSizing.AUTOSIZE)  # allow column shrinks
+			tview_column.set_resizable(True);
+			tview_column.set_reorderable(True);
```
introt, line 511: Reordering and resizing are lost on page reload. Remove the semicolons for consistency; they're not needed

introt (changes requested):
Great work! I'd lose the `set_resizable` and `set_reorderable` as their changes aren't saved and submit them in another PR after working that out.

introt:
One more change request: remove the semicolon on line 326. You might also want to edit out "enabled columns to resize and reorder" from the PR for clarity. Thanks for your cooperation!

VivekThazhathattil:
oops, sorry for that!

introt (approved):
All good, thanks for your contribution!

PR #1475, opened 2021-05-28
Add missing backarrow code

introt:
See commit messages.

I left tests/data/notebook-wiki.xml alone as changing it didn't seem to impact the tests and I wasn't sure how to handle < being interpreted as a start of an XML tag - should it be html.escape()d or is there another way?


introt:
Cherry-picked to https://github.com/RonnyPfannschmidt/zim-desktop-wiki/pull/1#issue-657509800

PR #1476, opened 2021-05-28
BackLinks_Pane: Options to show count, full paths

introt:
1. **Show BackLink count in title** determines whether the number of pages linking to the current page is shown in the widget's pane title. (heavily inspired by attachmentbrowser)
2. **Show full Page Names** determines which are shown: the pages' full names or just their basenames. (naming convention matches BookmarkBar)

Both are gloriously updated on preference changes.


introt:
Redoing...

PR #1479, opened 2021-05-29
BackLinks Pane: Add options to show count and full paths

introt:
- **Show BackLink count in title** determines whether the number of pages linking to the current page is shown in the widget's pane title.
- **Show full Page Names** determines which are shown: the pages' full names or their shorter relative names.
- Both are gloriously updated on preference changes.
- Updated manual to reflect the loss of the old status bar & linked to the plugin as an alternative.


introt:
Thanks. I agree True is a better default; I made it configurable to avoid an unlikely [xkcd 1172](https://xkcd.com/1172/) situation given that the title length affects side panes' width. Given that the commits have made their way into develop together with the default change, I presume I won't need to fix the resulting conflict :)

sojusnik:
Thanks for those improvements, love them!

Is it possible to shorten the tab title in the sidepane from f.i. "0 BackLinks" to "0 BL", in order to have the ability to make the sidpane narrower? Or maybe decrease the distance (padding) between those tabs a bit or add "…" at the end of a tab title, when narrowing the sidepane?

I would prefer to keep my sidepane narrow, but every new entry (read tab) in the sidepane makes the minimum width bigger.

introt:
You're welcome @sojusnik !

I can add an option to abbreviate BackLinks to BL, no problem. However, the width of "0 BL" and "100 BL" is noticeably different, and since the font is not monospace one can't (completely) prevent that by padding with spaces/zeroes. Hopefully you can deal with that!

The padding between the tabs and other sidepane behavior is not controllable by the plugins themselves (AFAIK) and thus must be handled by Zim itself. I like the idea of shorter minimum tab widths, though I'd personally go for fading instead of ellipsis. Feel free to open a new issue about that!

jaap-karssenberg:
Let's not go with abbreviations, that breaks the interface design. Could
add shortening with "..." On all titles by default.

Op za 5 jun. 2021 10:44 schreef introt ***@***.***>:

introt:
Affirmative, my Tallest!

sojusnik:
> Let's not go with abbreviations, that breaks the interface design. Could add shortening with "..." On all titles by default.

Yes, I also think that this is the most elegant solution.

jaap-karssenberg:
Had a look at adding "ellipsize" (the "..." option) to the labels in the
side pane, but doesn't play nice with the notebook tabs I'm afraid :/

sojusnik:
What about "…" in the middle of a word, like already done in the path bar?

grafik

jaap-karssenberg:
Yes, that is exactly what I mean - it doesn't play nice with the notebook
tab widget

The "..." is a feature of the label widget and calculated automatically
based on size allocation of the widget. Doesn't work when the lable is
inside the notebook tab.

-- Jaap

sojusnik:
Instead ellipsizing those entries, we could use generic icons for those. Those tabs then could have a minimal and maximal width, thus allowing to have 3-4 tabs without getting the sidebar too big.

This is the only solution that comes to my mind, allowing to keep the sidebar relatively narrow with more than 2 tabs in there.

jaap-karssenberg:
Alternatieve is to use a drop down menu button to switch views instead of
tabs. I've seen this in other applications. However it takes an additional
click to switch.

Op za 19 jun. 2021 23:15 schreef sojusnik ***@***.***>:

sojusnik:
> Alternatieve is to use a drop down menu button to switch views instead of tabs.

Then we would lose the benefit of seeing the count of the current backlinks and attachments at a glance, if I understand your suggestion correctly. With generic icons we still can have those numbers beside the icons and would save an additional click in comparison to the dropdown solution.

PR #1493, opened 2021-06-09
Fix typo in FAQ manual page

StefRe:
and make link captions consistent throughout the page (name only without path)

### DIFF
--- a/data/manual/FAQ.txt
+++ b/data/manual/FAQ.txt
@@ -21,7 +21,7 @@ These are pages that are linked by other pages but do not (yet) contain text. Yo
 ===== Why do some pages not disappear from the index after deleting them? =====
 The index keeps pages that are linked by other pages even if you delete them. To completely remove them you also need to change any page linking them.
 
-===== I pasted a shall script / C code / ... and it looks all mangled up - is this a bug ? =====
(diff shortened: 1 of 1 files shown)


data/manual/FAQ.txt
```diff
@@ -70,7 +70,7 @@ Zim is written as a "single user" program, so it is not intended for multiple pe
 ===== How do I change the Gtk theme? =====
 On a Linux desktop, you can probably do this via your settings manager.
 
-Windows users and users please see the section "Gtk configuration" in [[Help:Config Files]]
+Windows users and users please see the section "Gtk configuration" in [[Help:Config Files|Config Files]]
```
introt, line 73: Should this say "Windows and $mac_os users"? (no idea what MacOSX is called nowadays)

StefRe: Most probably yes (the manual refers to Mac as _macOS_), although I'm not familiar with macOS. The text was added in a66477e136e7, so in case of doubt we should ask @jaap-karssenberg .

jaap-karssenberg: Yes, agree it should be windows and Mac users

PR #1494, opened 2021-06-12
Make tests work in non-English locales

StefRe:
I had two failing tests when running `test.py` in my German locale due to comparing localized with non-localized texts. A straightforward workaround would be of course to use `LC_MESSAGES=en_US.utf8 ./test.py` but I felt it would be better to just fix the two tests.
#### TestUIActionsRealFile
```
testOpenDocumentRootNotDefined (tests.uiactions.TestUIActionsRealFile) ... ERROR: Für dieses Notizbuch ist kein Wurzelverzeichnis definiert
```
This is straightforward: just use the full localized message as expected message.

#### TestPluginClasses
```
FAIL: runTest (tests.plugins.TestPluginClasses)
  File "/home/steffen/Dev/zim-desktop-wiki/tests/plugins.py", line 73, in runTest
    self.assertIn(label, manual, 'Preference "%s" for %s plugin not documented in manual page' % (label, name))
AssertionError: 'Version automatisch speichern, wenn das Notizbuch geschlossen wird' not found in '...If the option **Autosave version when the notebook is closed** is enabled ...' : Preference "Version automatisch speichern, wenn das Notizbuch geschlossen wird" for versioncontrol plugin not documented in manual page
```
This is just the other way round: as there are no localized versions of the manual to compare the localized plugin preference strings against we have to temporarily disable `gettext` when getting the plugin messages by setting `_` to a no-op and restoring its previous value afterwards.

### DIFF
--- a/tests/plugins.py
+++ b/tests/plugins.py
@@ -3,6 +3,7 @@
 
 import tests
 
+import builtins
 import os
 
 from zim.plugins import *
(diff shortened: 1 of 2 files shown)


StefRe:
Unfortunately, the fix for `TestPluginClasses` doesn't work when run after other tests, obviously due to `gettext` caching. So I'm closing it for the time being.

Issue #155, opened 2017-07-13
Adopt sphinx for code documentation


jaap-karssenberg:
Used to use epidoc for code documentation, but generating html docs broke some time ago, move to spinx instead - see http://www.sphinx-doc.org

- Setup meta data needed to compile API docs
- Update inline documentation to use sphinx formatting
- Integrate docs that are in now "HACKING"

jaap-karssenberg:
See also
* https://www.tomaz.me/2013/09/28/migrating-from-epydoc-to-sphinx-style-docstrings-using-sed-and-some-command-line-fu.html
* https://gisellezeno.com/blog/sphinx-for-python-documentation/

introt:
I've started work on this. See

- Migration script Gist with some todos: https://gist.github.com/introt/58107bb59ad9d19c17ea0a21693c15fd
- sphinx-branch with minimal config: https://github.com/introt/zim-desktop-wiki/tree/sphinx
- commit msg with commands to get up to speed: https://github.com/introt/zim-desktop-wiki/commit/bee3fee3883ba8079a3a3c05fbb63680fd9eebfe

jaap-karssenberg:
Cool!

introt:
Update:

1. Thomaz's script/find-and-replace is too limited as it can't add missing whitespace
2. Adding missing whitespace leads to uneven line lengths and taking shortcuts leads to ugly spaghetti code
3. The real solution is to parse the existing epydoc strings into a tree and rewriting the docstrings with beautiful, perfect sphinx equivalents. I need to learn about (abstract|hierarchy|parse) trees for another project anyway so I can take a shot at it later, but right now I'm busy with other stuff so no promises.

Here's what the output of my last attempt (using a loop and a few rules for adding whitespace) looks like: https://introt.github.io/zim-desktop-wiki/docs/build/html/py-modindex.html

introt:
Before taking on the challenge I looked for existing tools and came across [pydoctor](https://pypi.org/project/pydoctor/), an API documentation generator that supports both epydoc's epytext as well as rst. Below are some reasons to consider adopting pydoctor right away:

- Supports the current inline epytext formatting.
  - Allows us to have up-to-date documentation available right now (almost; see next bullet) without modifications, as can be seen here: https://introt.github.io/zim-desktop-wiki/docs/api/
  - Exposes syntax errors in existing documentation
[...]

PR #1498, opened 2021-06-14
Update insert_line documentation

introt:
The new docstring follows the style of the other menu actions' docstrings (and doesn't confuse my tab-native docstring parser by indenting with spaces).

@jaap-karssenberg a few questions about a possible tool related to #155 :

1. `"""` vs. `'''`? PEP 257 says to use the former, while zim-desktop-wiki has a few thousand `'''`s with a couple hundred `"""`s mixed in.
    - Should one be standardized?
2. Line width; PEP 257 says 72, while Zim's docstrings max out at well over a hundred chars. I've prototyped a solution using TextWrapper that errs on the side of caution to not break code blocks etc, but judging whether the first line should stand alone is next to impossible due to incorrect formatting.
    - Should all docstrings be rewrapped and to what width?
3. Indentation inside docstrings; sphinx doesn't care about the amount or type of indentation as long as it's constant; thus a single space would suffice to fix field lists and would sidestep number 2.
    - What indentation should be used inside docstrings? A tab, a single space, or multiple spaces?
4. Since such a tool would modify all source files anyway, would you consider changing from tabs to four spaces? ;)

Thanks for your time!

### DIFF
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -6934,10 +6934,7 @@ class PageView(GSignalEmitterMixin, Gtk.VBox):
 
 	@action(_('Horizontal _Line'), menuhints='insert') # T: Menu item for Insert menu
 	def insert_line(self):
-		'''
-                This function is called from menu action.
-                Insert a line at the cursor position.
-		'''
+		'''Menu action to insert a line at the cursor position'''
 		buffer = self.textview.get_buffer()
 		with buffer.user_action:
 			buffer.insert_objectanchor_at_cursor(LineSeparatorAnchor())


jaap-karssenberg:
To be honest I don't really care about these pep style conventions - as long as the resulting text. I would be careful with re-wrapping because comments where hand typed for readability; re-wrapping them might cause more headaches than is justified by the gain of pleasing a pep.

( About those 72 characters - still using some old paper punch cards as bookmarks in paper books - luckily my own fortran coding was all digital - but I brought the paper cards to college when to explain my professor why his compiler borked at 72 ...  )

However I do care about the tabs, anyone has yet to convince me why I should make my `<Tab>` key insert 4 spaces when there is a perfectly fine `Tab` character to represent indentation.

introt:
Thanks for your input (and your punch card story)!

Rewrapping isn't limited to pleasing a PEP; converting epytext to rst will make the line endings not line up (as seen [here](https://github.com/introt/zim-desktop-wiki/commit/5aa3dc2f3fb1368bb70a3131b784904af184c3d0#diff-becd9017610318fb5cba536c9dc0b495d10ca67e43be3fc068211a97eb554384L173-R182)) due to differences in markup width and missing indentation in field lists etc. The effect can be limited by indenting with a single space instead of a tab, though the result will still occasionally look a bit wonky. 

Mixing spaces and tabs doesn't seem to affect Sphinx, so I guess I'll try implementing that to wrap up (pun intended) an alpha for a demo.

Issue #1423, opened 2021-03-18
FR: Image resizing


quesada:
Right now, pasting an image keeps the original size, which could be inappropriate for a notetaking program

jiewangjs:
strongly agree. Especially in HiDPI machine, this is a very unconvient thing. please give a resize drag feature soon.

jEsuSdA:
I think an usable approach will be to fit image with to zim width (and of course, the height should be the one to avoid image deformation).

Maybe a configuration value to fill the percentaje of zim with to scale images will be incredible good solution. So user can force a 100% width or 50% width (to have two images for row), etc.

As a more accuracy solution, launching the image viewer at clicking image will be a good point.

What do you think about?

fdt-dev:
Right-clicking on an image will bring up a dialogue window where the picture can be resized. In the same dialogue window a link can be added. It's all very simple and a lot of the ideas already are implemented. Have you tried that?

It's just that Zim can't do automatic resizing, as far as I understand how Zim works. I guess that must have to do with the fact that the "page" doesn't have a fixed length or width. Zim isn't a word-processor.
I just want pitch in here because I am pretty happy with how Zim currently deals with pictures. I think it's great how it works, resizing is easy and accurate and adding links to pictures (so you can click them to open them in the standard image viewer) is also perfectly straightforward. Just voicing my opinion.

neinseg:
> Right-clicking on an image will bring up a dialogue window where the picture can be resized. In the same dialogue window a link can be added. It's all very simple and a lot of the ideas already are implemented. Have you tried that?

I had the same issue as OP just now, and I had to search through this issue tracker to find out that these options are hidden under "edit properties" in the context menu. Discoverability of this feature is poor, and it breaks with the UI convention used with other text editing programs that allow images to be inserted.

jEsuSdA:
> Right-clicking on an image will bring up a dialogue window where the picture can be resized. In the same dialogue window a link can be added. It's all very simple and a lot of the ideas already are implemented. Have you tried that?
> 

I just discover that. Thanks a lot, @fdt-dev !!! ;)

directentis1:
Can we add diagram's property to resize diagram images  generated inside Zim?
Currently it has only "edit" option just to  edit code snippets and regenerate those diagrams.

![](https://i.imgur.com/tUbSpdn.png)

LSeelig:
[...]

Issue #324, opened 2018-02-19
Fit images inside viewport


synchr:
This is a feature request to make images fit inside the viewport of a displayed page.

It should behave the following way:
* If the image fits within the current viewport, it is displayed in its native size
* If the image doesn't fit inside the viewport, it is scaled down so it just fits inside the viewport

This behaviour should be preference so it can be disabled if necessary. If this feature is implemented it should in my opinion be enabled by default.

Ideally it should also work when a user resizes the Zim window. So the scaling should change according to the Zim window changes.

mtill:
Hi!
I would be really happy to see such a feature implemented in ZIM as I'm using ZIM for taking notes during MS Teams calls. inserting larger screenshots into a wiki page makes it somehow unhandy to use. From my perspective it would make sense to dynamically scale them down to the viewport size as proposed by seihtam.
I tried to find a solution for that and checked the source code, but I don't know where to start.
Any ideas how we could solve that?
BR
mtill

introt:
Long-term solution is the refactoring of image objects; in the short term, something like https://github.com/zim-desktop-wiki/zim-desktop-wiki/discussions/1937#discussioncomment-2205475 is more actionable

PR #1505, opened 2021-06-20
scale down pasted images to a max width, enabling the user to see the full pasted picture that can then more easily be integrated into a document

mtill:
currently, images pasted from clipboard are inserted as-is, even if they do not fit to window/screen width.

this behavior makes it pretty hard to use ZIM as a notebook for documenting online meetings, as capturing screenshots is a common task that often helps to caputure screen content presented in online sessions.

this commit scales down pasted images to a max width, enabling the user to see the full pasted picture that can then more easily be integrated into a document.

I would like to add an option in the notebook settings dialog to let the user define the desired setting for the max width value.
Any hint on how to get this done in the source code is appreciated.

related issues:
https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/1423
https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/324

### DIFF
--- a/zim/gui/clipboard.py
+++ b/zim/gui/clipboard.py
@@ -14,16 +14,13 @@ from gi.repository import GdkPixbuf
 
 import logging
 
-from functools import partial
-
 from zim.fs import File, Dir, FS
-from zim.newfs import LocalFolder, LocalFile, SEP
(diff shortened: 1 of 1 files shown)


mtill:
wrong version

PR #1506, opened 2021-06-20
scale down pasted images to a max width, enabling the user to see the full pasted picture that can then more easily be integrated into a document

mtill:
currently, images pasted from clipboard are inserted as-is, even if they do not fit to window/screen width.

this behavior makes it pretty hard to use ZIM as a notebook for documenting online meetings, as capturing and documenting screenshots is a common task usually done in many online sessions and web calls.

this commit scales down pasted images to a max width, enabling the user to see the full pasted picture that can then more easily be integrated into a document.

I would like to add an option in the notebook settings dialog to let the user define the desired setting for the max width value.
Any hint on how to get this done in the source code is appreciated.

### DIFF
--- a/zim/gui/clipboard.py
+++ b/zim/gui/clipboard.py
@@ -154,7 +154,10 @@ def deserialize_parse_tree(register_buf, content_buf, iter, data, length, create
 def deserialize_urilist(register_buf, content_buf, iter, data, length, create_tags, user_data):
 	notebook, path = user_data
 	links = unpack_urilist(data)
-	tree = _link_tree(links, notebook, path)
+	linkDicts = {}
+	for k in links:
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
I recognize the issue, but will not accept this fix for it.

The plan is to make implement images in the same way as inline objects and let them automatically adjust their width.

mtill:
thanks for the quick feedback, your idea on how to tackle this problem wrt too large images absolutely makes sense.
is there any rough time schedule when this feature would possibly be implemented?

--
i just wrote a small external script for zim that i will now use as a quick fix/hack for being able to use zim also as a note taking app during online meetings.

thank you very much for zim / its great!

PR #1515, opened 2021-06-30
improved backlinks plugin: added popup menu option to stick the plugin to a certain page, making it much easier for the user to iterate through a certain list of backlinks

mtill:
improved backlinks plugin:
1. added popup menu option to stick the plugin to a certain page, making it much easier for the user to iterate through a certain list of backlinks for a specific page instead of instantly switching to the backlinks of the newly opened page
2. fixed "Open in new page" popup entry

### DIFF
--- a/zim/plugins/backlinkpane.py
+++ b/zim/plugins/backlinkpane.py
@@ -83,40 +83,69 @@ class BackLinksWidget(Gtk.ScrolledWindow, WindowSidePaneWidget):
 		self.treeview.connect('row-activated', self.on_link_activated)
 		self.treeview.connect('populate-popup', self.on_populate_popup)
 
-	def set_page(self, notebook, page):
+		self.linked_to_pageview = True
+		self.currentNotebook = None
+		self.currentPage = None
+		self.update_title()
+
+	def updateModel(self):
 		model = self.treeview.get_model()
 		model.clear()
 
-		try:
-			backlinks = notebook.links.list_links(page, LINK_DIR_BACKWARD)
-				# XXX allow access through page object
-		except IndexNotFoundError:
-			backlinks = []
+		if self.currentNotebook is not None and self.currentPage is not None:
+			try:
+				backlinks = self.currentNotebook.links.list_links(self.currentPage, LINK_DIR_BACKWARD)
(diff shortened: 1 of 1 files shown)


mtill:
Hi!
please let me know if anything else would be needed in order to get this into the official release -- I'm willing to help wherever I can!
Best regards,
mtill

mtill:
here is a screenshot of the newly added context menu, enabling the user to "stick"/"decouple" the backlinks pane from the currently shown page:

image

jaap-karssenberg:
Sorry for the slow reply. I'm mostly offline this month.

I understand the use case for this change. However I worry the interface
may be unclear.

One improvement I can think of is to "pop out" the backlink pane and show a
dialog that is not linked to the pageview rather than "freezing" it in
place. I think that would be more obvious conceptually for the user.

Implementation wise this can easily be achieved by reusing the search
dialog with a query for backlinks. Or with a custom dialog of course.

Regards,

Jaap

Op ma 26 jul. 2021 21:39 schreef mtill ***@***.***>:

introt:
> One improvement I can think of is to "pop out" the backlink pane and show a
dialog that is not linked to the pageview rather than "freezing" it in
place.

> Implementation wise this can easily be achieved by reusing the search
dialog with a query for backlinks. Or with a custom dialog of course.

Isn't that what "Search -> Search Backlinks..." is?

Another way to iterate through a page's BackLinks is to open the page in a New Window, as clicking the BackLink pane links opens them in the main window.

> fixed "Open in new page" popup entry

Nice! I want that merged yesterday :+1:

mtill:
Hi!
Thank you very much for the feedback!

- I found the "stick to file" feature quite useful in daily life. it is quite common in many IDEs (for example in eclipse, see screenshot), I was not aware of the backlinks dialog, thanks for pointing that out -- I think that will do!
- "fixed "Open in new page" popup entry"
  --> I propose to get this fixed (there's simply one line that was missing here, namely: "menu.show_all()" ) as this is clearly a bug in ZIM


eclipse

mtill:
i will open a new pull request for fixing "Open in new page" popup entry

PR #1520, opened 2021-07-01
Make MoveTextDialog also move attachments if they are selected

### LINKED ISSUE #1519: Move text action does not move attachments if attachments are selected
As described.

I also have a fix coming up in a MR in a minute. Stay tuned.

theRealSuperMario:
Fix #1519


theRealSuperMario:
Disclaimer: I am extremely unfamiliar with the code base and will most like have screwed completely in how it should be implemented.

If you throw me some buzzwords and tags, I will try to quickly adapt the implementation so that it fits nicely into the rest of the code.

Cheers

jaap-karssenberg:
Hi Sandro,

Thanks for the proposal, makes sense. Also reflects similar behavior I
implemented for the clipboard in c28cde55159ae848ad5b6e586b742231cd717b9a
and e248aa714a75c7a3ad4393048c5b142699e2db54.

Suggestions for improvement:
- Use the "parsetree" to find images and file links in the content instead
of parsing the wiki text yourself, this will make it robust for changes in
the format
- Try to detect both images and attachments that are linked, equations pose
a special case -- see the clipboard code referred above how these are
handled there
- Check whether the image / attachments appears more often in the page, if
so, consider using "copy" instead of "move"
- Maybe above is easiest achieved by actually re-using code from the
clipboard and refactor the "move text" to use a "copy paste" action to
accomplish the intended result
- Add a test case :)

Thanks!

Jaap

PR #1534, opened 2021-08-03
foreach_line_in_selection: ignore empty lines

introt:
Status: https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/1534#issuecomment-986004627

---

Usually I don't want my empty lines to become empty list members, eg.

```
this is a list where items are separated by an empty line

let's make it into a checklist!
```

would turn into

```
[ ] this is a list where items are separated by an empty line
[ ] 
[ ] let's make it into a checklist!
```

instead of (imo) the more sane

```
[ ] this is a list where items are separated by an empty line

[ ] let's make it into a checklist!
```

Thus, this change. Q&A:

* Does it break some other functionality?
  * Maybe. It didn't break any tests though could do with some of its own.
* Shouldn't the empty lines be removed?
  * No, some people might be into that sort of stuff and it's not up to us to judge them.
* Does it fix my use case?
  * Yes.

### DIFF
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -2343,7 +2343,9 @@ class TextBuffer(Gtk.TextBuffer):
 				# because line is not visually part of selection
 				end.backward_char()
 			for line in range(start.get_line(), end.get_line() + 1):
-				func(line, *args, **kwarg)
+				# don't call function on empty lines
+				if not self.get_line_is_empty(line):
+					func(line, *args, **kwarg)
 			return True
 		else:
 			return False
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -2344,11 +2344,18 @@ class TextBuffer(Gtk.TextBuffer):
 				# exclude last line if selection ends at newline
 				# because line is not visually part of selection
 				end.backward_char()
-			for line in range(start.get_line(), end.get_line() + 1):
-				# don't call function on empty lines
-				if skip_empty_lines and self.get_line_is_empty(line):
-					continue
-				func(line, *args, **kwarg)
+			if skip_empty_lines:
+				all_lines_are_empty = True
+				for line in range(start.get_line(), end.get_line() + 1):
+					if not self.get_line_is_empty(line):
+						all_lines_are_empty = False
+						func(line, *args, **kwarg)
+				if all_lines_are_empty:
+					# the user wanted to do something to these empty lines, so we do
+					self.foreach_line_in_selection(func, *args, skip_empty_lines=False, **kwarg)
+			else:
+				for line in range(start.get_line(), end.get_line() + 1):
+					func(line, *args, **kwarg)
 			return True
 		else:
 			return False
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -4528,14 +4528,14 @@ class TextView(Gtk.TextView):
 			elif keyval in KEYVALS_ASTERISK + (KEYVAL_POUND,):
 				def toggle_bullet(line, newbullet):
 					bullet = buffer.get_bullet(line)
-					if not bullet and not buffer.get_line_is_empty(line):
+					if not bullet:
 						buffer.set_bullet(line, newbullet)
 					elif bullet == newbullet: # FIXME broken for numbered list
 						buffer.set_bullet(line, None)
 				if keyval == KEYVAL_POUND:
-					buffer.foreach_line_in_selection(toggle_bullet, NUMBER_BULLET)
+					buffer.foreach_line_in_selection(toggle_bullet, NUMBER_BULLET, skip_empty_lines=True)
 				else:
-					buffer.foreach_line_in_selection(toggle_bullet, BULLET)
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
I agree with you on the wanted behavior. However I think changing `foreach_line_in_selection()` is not a clean way to fix it. 

Indenting and unindenting functions also use this method and there it is less clear that this is what the user would want - e.g. in verbatim blocks unindented empty lines could break the block into two blocks.

A corner case to think about is what happens when the user only selects the empty line and then toggles a list type. Arguably in that case they do want to change the line.

Probably cleanest way out is to use a `kwarg` indeed, e.g. "skip_empty_lines" but make sure this only takes effect if there is at least one line in the selection that is not empty. Agree this makes it much more complicated but think it would be more robust in the long run -- at least with proper test cases.

introt:
Thanks for the quick reply, Jaap!

I agree with the kwarg, but let's argue on the specifics a bit :)

> A corner case to think about is what happens when the user only selects the empty line and then toggles a list type. Arguably in that case they do want to change the line.

Zim doesn't allow empty bullets (they're deleted on reload) so I think that's a moot point. If you mean an empty selection/having the cursor on an empty line should produce a bullet, I agree, and that's handled by _apply_bullet separately (if bounds --> else).

> Probably cleanest way out is to use a kwarg indeed, e.g. "skip_empty_lines" but make sure this only takes effect if there is at least one line in the selection that is not empty.

What's the use case for this? Zim doesn't support empty bullets and creating new lines generates new properly indented bullets anyway.

I'll commit the kwarg change when I get a sufficient break from life stuff.

introt:
> make sure this only takes effect if there is at least one line in the selection that is not empty.

7011af6 implements this. It's slightly spaghetti, but performant: it loops over the lines just once most of the time, twice only if the selection is completely empty.

However, unlike Format -> Bullet List, selecting empty lines and pressing * doesn't do anything. I've yet to investigate that further, but it's hardly a deal breaker.

> at least with proper test cases.

c293aca tests that skipping works, but testing for "not skipping fully-empty selections" isn't as straight-forward, as Zim deletes empty formatting. The test below fails with ['\n', '\n', '\n'] (which shouldn't be due to the above thing since this calls the same func as the GUI).

```
	def testApplyBulletDoesEmptyLinesIfWholeSelectionIsEmpty(self):
		pageview = setUpPageView(self.setUpNotebook(), '\n\n\n')
		buffer = pageview.textview.get_buffer()
		begin = buffer.get_iter_at_offset(0)
		end = buffer.get_iter_at_offset(3)
		buffer.select_range(begin, end)
		pageview.apply_format_bullet_list()
		self.assertEqual(pageview.page.dump('wiki'), ['* \n', '* \n', '* \n'])
```

introt:
Turns out `_do_key_press_event_selection` defines its own `toggle_bullet` which skips empty lines:

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/022bff6ebe91152192ac342d7d2456600b0ddc6b/zim/gui/pageview/__init__.py#L4519

I now realize that you might have originally meant the non-emptiness checking to be done by the function given to `foreach_line_in_selection()` instead of `foreach_line_in_selection()` itself- but then checking that there's at least one non-empty line gets really messy, so I'm still in favor of the kwarg implementation. Thus, I'm gonna remove emptiness-checking from toggle_bullet

jaap-karssenberg:
@introt is this PR ready for merge?

zim/gui/pageview/__init__.py
```diff
@@ -2342,8 +2344,18 @@ def foreach_line_in_selection(self, func, *args, **kwarg):
...
 				end.backward_char()
-			for line in range(start.get_line(), end.get_line() + 1):
-				func(line, *args, **kwarg)
+			if skip_empty_lines:
+				all_lines_are_empty = True
+				for line in range(start.get_line(), end.get_line() + 1):
+					if not self.get_line_is_empty(line):
+						all_lines_are_empty = False
+						func(line, *args, **kwarg)
+				if all_lines_are_empty:
```
introt, line 2353: I feel like a second kwarg that decides whether to call the func on empty lines iff all_lines_are_empty would be a bit excessive. On the other hand, it's the cleanest solution to support both the behavior you described (the user called format on empty lines, so we format the empty lines) as well as the previous behavior of `toggle_bullet`

zim/gui/pageview/__init__.py
```diff
@@ -4516,14 +4528,14 @@ def multi_line_indent(start, end):
 			elif keyval in KEYVALS_ASTERISK + (KEYVAL_POUND,):
 				def toggle_bullet(line, newbullet):
 					bullet = buffer.get_bullet(line)
-					if not bullet and not buffer.get_line_is_empty(line):
```
introt, line 4519: Previously, `toggle_bullet` handled checking for empty lines, resulting in different behavior between `apply_format_bullet_list`/using the Format menu and pressing `*` on a selection

introt:
@jaap-karssenberg It's ready for your review. I'm not sure whether we need to bullet empty selections as pressing `*` hasn't done that before, but if you feel like that should be done, the code is there. Can be merged as-is, or we could add a second kwarg as described in the review comments above

PR #1536, opened 2021-08-06
fixing backlinks plugin: open backlink in new window

mtill:
backlinks plugin: fixed "Open in new page" popup entry that is currently not shown when trying to open the context menu


mtill:
I would recommend to have this small fix in the next release -- as it is clearly a bug.
Please let me know what else I could provide for getting it ready to be merged

Thanks and kind regards
mtill

PR #1543, opened 2021-08-11
Feature: Option to always create non-existing folders without prompting the user

### LINKED ISSUE #1541: Attachments: Option to always create non-existing folders without prompting
Current behavior: Trying to open non-existing attachment folder prompts the user to create one
Wanted behavior: An option to always create a non-existing folder without further user interaction

Code responsible for current behavior:

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/ea04e9fd52de102e95f653adab803d5f16c648f9/zim/gui/applications.py#L576-L591

~~There's also this bit that could (should?) share the implementation:~~ Included in PR

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/ea04e9fd52de102e95f653adab803d5f16c648f9/zim/gui/pageview/__init__.py#L7190-L7211

introt:
* make pageview use zim.gui.applications.open_folder_prompt_create
* add preference to always create missing directories without prompting
* Fixes #1541

### DIFF
--- a/zim/gui/applications.py
+++ b/zim/gui/applications.py
@@ -24,7 +24,7 @@ from gi.repository import GdkPixbuf
 import zim.fs
 from zim.fs import File, Dir, TmpFile, cleanup_filename
 from zim.config import XDG_CONFIG_HOME, XDG_CONFIG_DIRS, XDG_DATA_HOME, XDG_DATA_DIRS, \
-	data_dirs, SectionedConfigDict, INIConfigFile
+	data_dirs, SectionedConfigDict, INIConfigFile, Boolean, ConfigManager
 from zim.parsing import split_quoted_strings, uri_scheme
 from zim.applications import Application, WebBrowser, StartFile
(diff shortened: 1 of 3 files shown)


introt:
Todo:

- [x] make pageview use open_folder_prompt_create
- [x] add a preference for always creating missing folders
- [ ] make open_folder_prompt_create check said preference

The preference in the dialog saves state, next up is making applications.py read it. I need some zzz:s before trying to figure out that one.

introt:
So- turned out preferences should start with a capital. Thought that was just because the other ones were named after classes. Must say that the [PyDoctor docs](https://introt.github.io/zim-desktop-wiki/docs/api/zim.config.dicts.ConfigDict.html#setdefault) were really useful (and available without conversion to sphinx format) – though I'm not sure if it has support for custom fields (sphinx does via plugins, though they seem to all be BSD-licensed).

PR #1546, opened 2021-08-13
Feature: Add "Cut Link" to context menu

introt:
The placement follows that of normal cut and copy, while the mnemonic is 'k' to avoid collisions.

Use case: Reorganizing link lists

Examples:

1. Turning links with "and" in them into two separate links
2. Keeping the link text while moving the link to a sub-item

### DIFF
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -6707,7 +6707,7 @@ class PageView(GSignalEmitterMixin, Gtk.VBox):
 
 		item = [
 				Gtk.MenuItem.new_with_mnemonic(_('Copy _Link')), # T: context menu item
-				Gtk.MenuItem.new_with_mnemonic(_('Cu_t Link')), # T: context menu item
+				Gtk.MenuItem.new_with_mnemonic(_('Cut Lin_k')), # T: context menu item
                 ]
 
 		if type == 'page':
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -6686,7 +6686,7 @@ class PageView(GSignalEmitterMixin, Gtk.VBox):
 		item.set_sensitive(not self.readonly)
 		menu.prepend(item)
 
-		# copy
+		# copy & cut
 		def set_pagelink(o, path, anchor):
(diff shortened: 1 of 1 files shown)


pfalcon:
Why can't you "Copy link" and "Remove link" instead? Why desire to make context menu to scroll off the screen by overloading it with features 99% of people will never need (and even won't know what they do, I for one would never guess what "Cut link" does).

Btw, "Remove link" could be achieved with the standard "Clear formatting" shortcut of Ctrl+9. Except that it's buggy (given e.g. a URL, it will "clear" it piece by piece instead of the entire link). And that's another problem - Zim is full of buggy features already, and people want to add even more features to make it completely buggy.

introt:
Greetings, fellow human! I appreciate your continued interest in Zim's conquest. Allow me to alleviate your concerns.

* Copy + remove is twice the clicks; this makes some workflows two times faster.
* Context menus getting rendered off-screen is another issue and irrelevant to the matter at hand.
  * Btw is "Copy link to this location" visible when clicking links or selecting plain text for a reason?
* I've already found this quite useful; it enables a non-cumbersome way to reorganize links where the text doesn't match the page name.
  * Which is "most of the time" in the case of languages with [inflection](https://en.wiktionary.org/wiki/taivutus), like mine.
* As usual, cutting things removes them from the selection and puts them to the clipboard. Given the context (Copy & Remove Link) and the fact that the user can just try it out and Ctrl+Z, I wouldn't worry too much about discoverability.
* Ctrl+9 requires either uncomfortable stretching or having my right hand on the keyboard, while the copying is done with the mouse. Doing both in one go without letting go of the rodent sounds awfully convenient to me.
  * Also, clear formatting clears all formatting, not just the link- this doesn't _cut it_ (pun very much intended, haha) as text can nowadays have multiple formatting options applied at once.
  * In my humble opinion it does what it says on the tin- clears the formatting of the selected text, leaving its surroundings alone. Hopefully I've read it right, as I depend on this behavior in some of my workflows.
* The implementation is as simple as chaining "Copy link" and "Remove link", there's no new logic to maintain or introduce new bugs.
  * It does define a new mnemonic, which overlaps with the regular cut- thanks for bringing that to my attention!

Issue #1040, opened 2020-03-31
Feature Request: Moving Items in lists up & down


joemaro:
I'd like to be able to simply move the currently selected part of a list up or down. In case of various items being selected they'd all move accordingly.

To be precise, here's a situation:

* item one
* item two (cursor is in this row)
* item three

after pressing Alt+Up (for example) the situation would look like this:

* item two (cursor is in this row)
* item one
* item three

If it's a numbered list, the numbers would automatically adapt and keep in order.

sojusnik:
Already possible, see [Line Sorter](https://zim-wiki.org/manual/Plugins/Line_Sorter.html) plugin. AFAIR, this was asked several times before, presumably, because it's not mentioned in the description. Anyone wants to fix that?

nelas:
That's already possible using `Ctrl+Up` or `Ctrl+Down` to move a line up or down. If you selected more than one line and ctrl+up, they will move together. I don't think you need any plugin. However, in a numbered list the numbers will not be automatically updated.

joemaro:
> That's already possible using `Ctrl+Up` or `Ctrl+Down` to move a line up or down. If you selected more than one line and ctrl+up, they will move together. I don't think you need any plugin. However, in a numbered list the numbers will not be automatically updated.

I tried that before requesting this and for me they don't work and are not listed in the keybindings section of the help-document.

Version is 0.72.1 which is the latest, and i'm using it on manjaro linux. 

Apart of that: thanks for an amazing program devs!!!

PR #1549, opened 2021-08-15
Update description of Line Sorter plugin

timbze:
As recommended [here](https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/1040#issuecomment-607375906)


introt:
Hi and thanks for your PR! Could you also update the manual? Thanks!

PR #1550, opened 2021-08-16
Modify Keybindings help text to reflect actual context

shulist:
In the Key Bindings section of the manual the function:

`<Ctrl>D       Inserts timestamp`

is a bit incomplete and should correctly state

`<Ctrl>D       Insert Date and Time`

At least on the Windows & Linux version that is what the pop-up is titled. (I cannot check Mac).


introt:
Hi shulist!

I agree the term "timestamp" is a bit technical, but would rather elaborate than replace - maybe `Insert a timestamp with the "date and time" dialog` could work ("edit link" isn't capitalized above)?

By the way, it looks like you've accidentally opened a PR from the develop branch to the master branch- happens to the best of us. The cleanest way to fix that is to close this PR and open a new one.

shulist:
@introt so as a GitHub newbie I am not sure what branch(es) I should be using to open this PR. Suggestions?

PR #1551, opened 2021-08-19
Fix tag prefix @@

ArmstrongW:
lots of language use @, example:  java annotation / python sugar / perl list ..., so i suggest modify tag prefix to @@, please.


introt:
You can press Ctrl+Z after typing `@text` or clear its formatting to have it not be a tag. There's also Code Blocks (with syntax highlighting) and Format --> Verbatim. Thus I see no need for this breaking change.

ArmstrongW:
ok, expect the new version  :+1:

PR #1552, opened 2021-08-19
Pop from history when leaving the unmodified page quickly

e3rd:
I often navigate through the notebook using <kbd>Ctrl</kbd>+<kbd>PgUp/Down</kbd>. The pages change according to the position in the contents sidebar. However, when I need to navigate further than a page, all the pages end up in the history. Which is not intented, I wanted just switch from `A` to `D`, opening `B` and `C` was just a side effect. When I want to go back in history by <kbd>Alt</kbd>+<kbd>Left</kbd>, I'd prefer to end up directly back on `A`, not to cycle through `C`, `B`, `A` respectively.

The proposed solution waits 1 s. If page is being left within 1 s, the page that ended up at the top of the history stack gets deleted. Would you please give me few advices? This is just proof of concept, I understand it has to be expanded. Is the solution good or did I take a wrong way? Should I create a method `history.py/History.pop_last`?

### DIFF
--- a/zim/gui/mainwindow.py
+++ b/zim/gui/mainwindow.py
@@ -262,6 +262,7 @@ class MainWindow(WindowBaseMixin, Window):
 		"C{WxH+X+Y}", if C{None} the previous state is restored
 		'''
 		Window.__init__(self)
+		self.last_change = 0
 		self.notebook = notebook
 		self.page = None # will be set later by open_page
 		self.navigation = NavigationModel(self)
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
I don't feel this kind of logic makes sense. After all, your opened the page, so it should be shown in the history.

One alternative I can think of is to follow the logic of the pathbar plugin and have an option to only add modified pages to the history to separate browsing and editing work flow.

e3rd:
> After all, your opened the page

I've opened it, yet it was my desire, I just wanted to navigate to the page below. If I wanted to separate editing work flow, I would have chosen View / Pathbar / Recently Changed page.

But if that does not make sense to you, I can do nothing, have a nice time and thanks for having replied

Issue #86, opened 2017-05-31
Make linespacing in wrapped lines configurable


sojusnik:
I'm referring to [this](https://bugs.launchpad.net/zim/+bug/1537876) and [this](https://answers.launchpad.net/zim/+question/264064).

jaap-karssenberg:
See http://zim-wiki.org/manual/Help/Config_Files.html - does "linespacing" do what you are looking for ?

sojusnik:
No, see quote from the link above:

> The 'linespacing' option in style.conf only dictates the spacing from one paragraph to the next, which in editorial context is called "space after" or "space below". I honestly think it should be called "paragraph spacing", in both contexts.
> 
> Zim could support a per line spacing, improving the readability of text, as well as allowing us to use alternative fonts that are quite packed. Linux Biolinum, for instance, is too crowded without line spacing, but is very readable when the lines are spread.

jaap-karssenberg:
Implement similar to "linespacing" but use gtk.TextView.set_pixels_inside_wrap

Or make "linespacing" affect both and add a new "paragraphspacing" to set set_pixels_below_lines realtive to linespacing

PR #1555, opened 2021-08-22
Make linespacing affect wrapped lines

kawapa:
The parameter "linespacing" now affects also wrapped lines.
Issue #86

### DIFF
--- a/data/style.conf
+++ b/data/style.conf
@@ -4,6 +4,7 @@ tabs =
 font =
 justify =
 linespacing = 3
+wrapped-lines-linespacing = 3
 
 [Tag strong]
 weight = PANGO_WEIGHT_BOLD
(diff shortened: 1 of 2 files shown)


sojusnik:
Thanks for this PR! So `linespacing` now affects the space between paragraphs **and** lines. Is there a new value like `paragraphspacing` that alters only the space between paragraphs? The latter is needed, because with the new `linespacing` you don't have a visual distinction between paragraphs any more. The distance between lines and paragraphs is the same now.

kawapa:
@sojusnik
My above commit does what you want.

jaap-karssenberg:
Sorry, just committed an alternative fix that reverts the behavior of
"linespacing" to avoid regression and introduces "line-height" (borrowed
name from CSS) for the intended effect on wrapped lines.

jaap-karssenberg:
Decided to go with your version anyway - "line-height" doesn't work well with a value of zero - it is a spacing on top of the height really. Will merge both patches

Issue #1563, opened 2021-09-04
Animated GIF resize doesn't work


xuthus:
This is zim 0.73.5
Platform: Windows 10, Macos Catalina
Locale: en_US UTF-8, ru_RU
Python: 3.9.6
PixbufVersion: 2.42.6

1. Insert GIF image into the page.
2. Right-click it, choose `Edit Properties`
3. Change `Width` or `Height`, click `OK`

Expected result:
* the image changed its size to values entered

Actual result:
* warning icon is shown instead of the image

![warning icon](https://user-images.githubusercontent.com/6282044/132103358-d988039f-682d-4734-a755-e3a4f1accb15.png)

Reason:
1. Add a breakpoint to a line pageview.py:1198
2. Open Edit Properties dialog, click OK button
3. Debugger stops the application on the breakpoint
4. Call of the method `GdkPixbuf.Pixbuf.new_from_file_at_size(file.path, w, h)` produces exception with a message
> gdk-pixbuf-error-quark: Not all frames of the GIF image were loaded.

Quick googling hints that the root cause in GIF codec used by gdk-pixbuf.
[Here for example the problem is fixed for `thumbnailer`](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/commit/6418d6da6a56519aa1585526067d891b5b2a87e0). But for many other places the issue still active, for example for Python wrappings.

As a quick fix I suggest loading pixbuf using `new_from_file` and scale image in zim code.
If sometimes ZIM contributors decide to add other options for images (additionally to resize) new options will be implemented in ZIM, so my suggestion will not hinder it.

xvusrmqj:
gif show, but still can not play on windows zim 0.75.0

PR #1564, opened 2021-09-04
#1563 - GIF resize doesn't work

xuthus:
fixed animated GIF resizing - `new_from_file_at_size` replaced with `new_from_file.scale_simple`


jaap-karssenberg:
Will merge this in develop after next "point" release. Could you please add a test case to make sure I don't break it again later when I forgot all about this issue? Have a look at "`TestWebPImageSupport`" in "`tests/pageview.py`" for a example test case.

jaap-karssenberg:
Merged via new commit - let to many conflicting commits in between - my fault

PR #1567, opened 2021-09-08
FindWidget: re.escape only selected text; fixes #1566

### LINKED ISSUE #1566: Find and Replace: needless escaping when opening in regex mode
If the `Regular expression` box is ticked, the regex characters in an earlier search string get escaped every time the dialog is opened. Escaping should only be applied to the selected text, not to the saved earlier string.

Affected versions: 0.73.5 and latest commit in develop branch

Steps to reproduce:

1. Open the dialog with Ctrl+H
2. Find what: `.*`
3. Make sure "Regular expression" box is ticked.
3. Close the dialog.
4. Make sure no text is selected and reopen the dialog with Ctrl+H
5. (Repeat steps 3 to 4 ad infinitum)

Expected result: `.*` in the search field
Actual result: `\.\*` in the search field after step 4 and an increasing number of backslashes if repeated: `\\\.\\\*`, `\\\\\\\.\\\\\\\*`, `\\\\\\\\\\\\\\\.\\\\\\\\\\\\\\\*`...

Somewhat related issue: #1158

introt:
Changed the scope of the check, now only escapes selected text

### DIFF
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -8175,8 +8175,8 @@ class FindWidget(object):
 		if bounds:
 			start, end = bounds
 			string = start.get_slice(end)
-		if flags & FIND_REGEX:
-			string = re.escape(string)
+			if flags & FIND_REGEX:
+				string = re.escape(string)
 		self.find(string, flags, highlight)
 
 	def on_find_entry_changed(self):


introt:
@jaap-karssenberg ping, really wish you'd pick this up before the point release! Fixes #1566 (unintended re.escaping)

PR #1604, opened 2021-09-21
New syntax (anchors) → bump WIKI_FORMAT_VERSION

Self-Perfection:
Release 0.74 introduced changes to syntax format that should be reflected in format version.

### DIFF
--- a/zim/formats/wiki.py
+++ b/zim/formats/wiki.py
@@ -19,9 +19,10 @@ from zim.formats.plain import Dumper as TextDumper
 old_url_re = old_url_re.p
 
 
-WIKI_FORMAT_VERSION = 'zim 0.6'
+WIKI_FORMAT_VERSION = 'zim 0.7'
 assert WIKI_FORMAT_VERSION != 'zim 0.26' # skip number for historic reasons
 # History:
+# - 0.7  links to section within note
 # - 0.6  new url parsing logic
 # - 0.5  support nested formatting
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
We discussed and chosen not to do so since a) this syntax was reserved
already for object usage and b) no upgrade available for edge cases where
there might be a conflict.

Op di 21 sep. 2021 12:04 schreef Alexander Meshcheryakov <
***@***.***>:

Self-Perfection:
I don't quite understand this reasoning. What upgrade you might expect for conflicting cases? What is it, how should it behave?

Anyway if this have been discussed and decision have been made then feel free to reject this PR. I honestly assumed that this was forgotten.

I just want to highlight that new syntax does not handled nicely with older zim versions, namely 0.73.2. Link `Some:Note#Section' leads to page `Some:NoteSection`.

jaap-karssenberg:
Yes, I understand the conflict. My reasoning is that a version bump only
helps if there is also code using the version to make the conflicts go
away. In this case there is no such compatibility.

Self-Perfection:
Even if Zim does not have special code to handle this format change other tools and apps might have such special handling.

Or let me put it this way: seeing `Wiki-Format: zim 0.6` at the top of text file provided particular expectations before. And now one does not know what to expect exactly. Value '0.6' was deprived of its meaning.

PR #1625, opened 2021-09-26
Make zeitgeist plugin work again

### LINKED ISSUE #787: Cannot switch-on zeitgeist plugin
Xubuntu 18.10
Zim 0.70 snapshot 11.04.2019
libzeitgeist
python-zeitgeist

But I cannot switch-on zeitgeist plugin in plugins-list in options.

I see:
• libzeitgeist - Error (Дополнительно)

mstimberg:
Some minimal changes to make the zeitgeist plugin work again. Note that this now requires the `python3-zeitgeist` package on Debian/Ubuntu.

Fixes #787

### DIFF
--- a/data/manual/Plugins/Log_events_with_Zeitgeist.txt
+++ b/data/manual/Plugins/Log_events_with_Zeitgeist.txt
@@ -6,6 +6,6 @@ Creation-Date: 2012-06-05T20:52:03+02:00
 
 When this plugin is activated, information about accessing, creating and modifying Zim Wiki pages is logged using the Zeitgeist framework. This allows applications like the Unity Dash to display recently used pages or the Activity Journal to give a timeline overview of your activities.
 
-**Dependencies:** This plugin requires the Python bindings for Zeitgeist. For Debian and Ubuntu, these bindings are available in the ''python-zeitgeist'' package.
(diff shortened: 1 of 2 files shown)


introt (approved):
Great work, thank you!

I can confirm that the plugin now loads without error and sends events on page change as intended.

~However, while there is an event sent for page deletion, there doesn't seem to be one for pages being renamed, which results in Activity Journal listing the file with both the old and new names and triggering the Crash reporter when trying to access the nonexistent file.~ Nevermind that, the same crash happens after deleting a page, and isn't limited to Zim: I was able to reproduce it with other programs by simply renaming and/or deleting files via the shell while the AJ was open.

The crash is limited to Activity Journal and the nonexistent files/duplicates are cleared from the UI on its restart. To me, this sounds like an upstream issue with Activity Journal and not an issue with the plugin.

image
_Exhibit A: "page" was renamed to "test" via Zim!_

image
_Exhibit B: The error as shown in the "Crash report" tool._

Supporting page renames might prove challenging due to having to send events for each subpage which may or may not exist on Zeitgeist's side etc; regardless, I would consider that to be a separate feature and out-of-scope for this PR, as it restores the feature set the plugin has previously shipped with and doesn't break Zeitgeist.

Nick222:
Open Zim - edit opened note without switching to another note - save and close Zim - cannot see any entry in Zeitgeist journal.

Open Zim - switch to another note without editing - close Zim - see new entry in Zeitgeist journal.

Is it right?

Xubuntu 21.04
Zim 0.74.1

PR #1626, opened 2021-09-26
Fix #1560: add linking related UX in ToC pane

### LINKED ISSUE #1560: ToC pane missing Copy Location & Drag'n'drop for links
Given #1293 the ToC pane should support the same linking-related UX as the other TreeView panes, as in being able to copy links to headings that aren't currently visible in the editor.

Edit: "Open in New Window" would also be useful when reworking the structure of longer notes

roland-ruedenauer:
The context menu for ToC entries gets extended by offering now in addition
- "Copy link to heading"
- "Open in New Window"

The PR also fixes the wrong displaying of headings with an explicit anchor in the ToC pane (filtering for a tag was added bc of this).

### DIFF
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -6526,7 +6526,6 @@ class PageView(GSignalEmitterMixin, Gtk.VBox):
 
 	def _do_activate_link(self, link, hints):
 		type = link_type(link)
-
 		if type == 'page':
 			href = HRef.new_from_wiki_link(link)
 			path = self.notebook.pages.resolve_link(self.page, href)
@@ -6536,17 +6535,20 @@ class PageView(GSignalEmitterMixin, Gtk.VBox):
 			open_file(self, path)
 		elif type == 'notebook':
 			from zim.main import ZIM_APPLICATION
-
 			if link.startswith('zim+'):
-				uri, pagename = link[4:], None
+				uri, pagename, anchor = link[4:], None, None
 				if '?' in uri:
 					uri, pagename = uri.split('?', 1)
-
-				ZIM_APPLICATION.run('--gui', uri, pagename)
-
+				args = [uri]
+				if pagename and '#' in pagename:
+					pagename, anchor = pagename.split('#', 1)
+				if pagename:
+					args.append(pagename)
+				if anchor:
+					args.append(anchor)
 			else:
-				ZIM_APPLICATION.run('--gui', FilePath(link).uri)
-
+				args = [FilePath(link).uri]
+			ZIM_APPLICATION.run('--gui', *args)
 		else:
 			if type == 'mailto' and not link.startswith('mailto:'):
 				link = 'mailto:' + link  # Enforce proper URI form
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -1387,10 +1387,10 @@ class TextBuffer(Gtk.TextBuffer):
 			return None
 
 	def _get_implict_anchor_if_heading(self, iter):
-		text = self._get_heading_text(iter)
+		text = self.get_heading_text(iter)
 		return heading_to_anchor(text) if text else None
 
-	def _get_heading_text(self, iter):
+	def get_heading_text(self, iter):
 		line_start = iter.copy() if iter.starts_line() else self.get_iter_at_line(iter.get_line())
 		is_heading = any(filter(_is_heading_tag, line_start.get_tags()))
 		if not is_heading:
@@ -1398,7 +1398,7 @@ class TextBuffer(Gtk.TextBuffer):
 
 		line_end = line_start.copy()
 		line_end.forward_line()
-		return line_start.get_text(line_end)
+		return line_start.get_text(line_end).strip()
 
 	#endregion
 
@@ -6665,7 +6665,7 @@ class PageView(GSignalEmitterMixin, Gtk.VBox):
 		item = Gtk.MenuItem.new_with_mnemonic(_('Copy _link to this location')) # T: menu item to copy link to achor location in page
 		anchor = buffer.get_anchor_for_location(iter)
 		if anchor:
-			heading_text = buffer._get_heading_text(iter) # can be None if not a heading
+			heading_text = buffer.get_heading_text(iter) # can be None if not a heading
 			item.connect('activate', _copy_link_to_anchor, anchor, heading_text)
 		else:
 			item.set_sensitive(False)
--- a/zim/gui/server.py
+++ b/zim/gui/server.py
@@ -132,7 +132,7 @@ class ServerWindow(Gtk.Window):
 		try:
 			uri = self.notebookcombobox.get_notebook()
 			if uri:
-				notebook, x = build_notebook(NotebookInfo(uri))
+				notebook, _ = build_notebook(NotebookInfo(uri))
 				if not notebook:
 					return
 			else:
--- a/zim/main/__init__.py
+++ b/zim/main/__init__.py
@@ -28,8 +28,7 @@ from zim import __version__
 
 from zim.utils import get_module, lookup_subclass
 from zim.errors import Error
-from zim.notebook import Notebook, Path, \
-	get_notebook_list, resolve_notebook, build_notebook
+from zim.notebook import Notebook, Path, get_notebook_list, resolve_notebook, build_notebook
 from zim.formats import get_format
 
 from zim.config import ConfigManager
@@ -39,12 +38,14 @@ from .command import Command, GtkCommand, UsageError, GetoptError
 from .ipc import dispatch as _ipc_dispatch
 from .ipc import start_listening as _ipc_start_listening
 
+from typing import Optional, Tuple
+
 
 class HelpCommand(Command):
 	'''Class implementing the C{--help} command'''
 
 	usagehelp = '''\
-usage: zim [OPTIONS] [NOTEBOOK [PAGE]]
+usage: zim [OPTIONS] [NOTEBOOK [PAGE [ANCHOR]]]
    or: zim --server [OPTIONS] [NOTEBOOK]
    or: zim --export [OPTIONS] NOTEBOOK [PAGE]
    or: zim --search NOTEBOOK QUERY
@@ -137,9 +138,9 @@ class NotebookCommand(Command):
 
 	def get_notebook_argument(self):
 		'''Get the notebook and page arguments for this command
-		@returns: a 2-tuple of an L{NotebookInfo} object and an
-		optional L{Path} or C{(None, None)} if the notebook
-		argument is optional and not given
+		@returns: a 3-tuple of an L{NotebookInfo} object, an
+		optional L{Path} and an optional anchor or C{(None, None, None)}
+		if the notebook argument is optional and not given
 		@raises NotebookLookupError: if the notebook is mandatory and
 		not given, or if it is given but could not be resolved
 		'''
@@ -147,25 +148,32 @@ class NotebookCommand(Command):
 		args = self.get_arguments()
 		notebook = args[0]
 
+		notebookinfo = None
+		path = None
+		anchor = None
+
 		if notebook is None:
 			if self.arguments[0] == 'NOTEBOOK': # not optional
 				raise NotebookLookupError(_('Please specify a notebook'))
 					# T: Error when looking up a notebook
-			else:
-				return None, None
-
-		notebookinfo = resolve_notebook(notebook, pwd=self.pwd)
-		if not notebookinfo:
-			raise NotebookLookupError(_('Could not find notebook: %s') % notebook)
-				# T: error message
-
-		if len(self.arguments) > 1 \
-		and self.arguments[1] in ('PAGE', '[PAGE]') \
-		and args[1] is not None:
-			pagename = Path.makeValidPageName(args[1])
-			return notebookinfo, Path(pagename)
 		else:
-			return notebookinfo, None
+			notebookinfo = resolve_notebook(notebook, pwd=self.pwd)
+			if not notebookinfo:
+				raise NotebookLookupError(_('Could not find notebook: %s') % notebook)
+					# T: error message
+
+			if len(self.arguments) > 1 \
+			and self.arguments[1] in ('PAGE', '[PAGE]') \
+			and args[1] is not None:
+				pagename = Path.makeValidPageName(args[1])
+				path = Path(pagename)
+
+			if len(self.arguments) > 2 \
+			and self.arguments[2] in ('ANCHOR', '[ANCHOR]') \
+			and args[2] is not None:
+				anchor = args[2]
+
+		return notebookinfo, path, anchor
 
 	def build_notebook(self, ensure_uptodate=True):
 		'''Get the L{Notebook} object for this command
@@ -173,7 +181,7 @@ class NotebookCommand(Command):
 		@param ensure_uptodate: if C{True} index is updated when needed.
 		Only set to C{False} when index update is handled explicitly
 		(e.g. in the main gui).
-		@returns: a L{Notebook} object and a L{Path} object or C{None}
+		@returns: a L{Notebook} object, a L{Path} object and an anchor or C{None}
 		@raises NotebookLookupError: if the notebook could not be
 		resolved or is not given
 		@raises FileNotFoundError: if the notebook location does not
@@ -181,9 +189,10 @@ class NotebookCommand(Command):
 		'''
 		# Explicit page argument has priority over implicit from uri
 		# mounting is attempted by zim.notebook.build_notebook()
-		notebookinfo, page = self.get_notebook_argument() 	# can raise NotebookLookupError
+		notebookinfo, page, anchor = self.get_notebook_argument() 	# can raise NotebookLookupError
 		if not notebookinfo:
 			raise NotebookLookupError(_('Please specify a notebook'))
+
 		notebook, uripage = build_notebook(notebookinfo) # can raise FileNotFoundError
 
 		if ensure_uptodate and not notebook.index.is_uptodate:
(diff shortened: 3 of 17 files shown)


Tryme69 (approved):

introt:
Nice work!

* On the first run I had an issue with the menu items being off-by-one which I couldn't reproduce.
* If you have the time, it'd imo make sense to disable pro-/demoting of horizontal bars
- [ ] Pasting a link to a page title heading broke with the error shown below. This works on latest `develop` commit when copied from inside the wiki page, but is broken here regardless of whether copying from the page or the plugin.

```
DEBUG: set_pagelink <Notebook: Test> <Page: test:subtest:subsub> anchor=subsub text=subsub
DEBUG: Action: paste
DEBUG: Targets available for paste: [Gdk.Atom.intern("TIMESTAMP", False), Gdk.Atom.intern("TARGETS", False), Gdk.Atom.intern("MULTIPLE", False), Gdk.Atom.intern("SAVE_TARGETS", False), Gdk.Atom.intern("UTF8_STRING", False), Gdk.Atom.intern("COMPOUND_TEXT", False), Gdk.Atom.intern("TEXT", False), Gdk.Atom.intern("STRING", False), Gdk.Atom.intern("text/plain;charset=utf-8", False), Gdk.Atom.intern("text/plain", False)], we want parsetree
DEBUG: Requesting data for ('text/x-zim-parsetree', 0, 1) -- using set_with_data() workaround
  File "/home/introt/repos/zim-desktop-wiki/zim/gui/pageview/__init__.py", line 4141, in do_paste_clipboard
    self.get_buffer().paste_clipboard(Clipboard, None, self.get_editable(), text_format=format)
  File "/home/introt/repos/zim-desktop-wiki/zim/gui/pageview/__init__.py", line 3376, in paste_clipboard
    parsetree = clipboard.get_parsetree(self.notebook, self.page, text_format)
  File "/home/introt/repos/zim-desktop-wiki/zim/gui/clipboard.py", line 641, in get_parsetree
    return parsetree_from_selectiondata(selectiondata, notebook, path, text_format)
  File "/home/introt/repos/zim-desktop-wiki/zim/gui/clipboard.py", line 228, in parsetree_from_selectiondata
    replace_parsetree_links_and_copy_images(parsetree, notebook, path)
  File "/home/introt/repos/zim-desktop-wiki/zim/notebook/content_updater.py", line 94, in replace_parsetree_links_and_copy_images
    parsetree.replace(
  File "/home/introt/repos/zim-desktop-wiki/zim/formats/__init__.py", line 668, in replace
    self._replace(self._etree.getroot(), tags, func)
  File "/home/introt/repos/zim-desktop-wiki/zim/formats/__init__.py", line 679, in _replace
    replacement = func(Element.new_from_etree(child))
  File "/home/introt/repos/zim-desktop-wiki/zim/notebook/content_updater.py", line 168, in _replace_links_to_page_and_copy_images
    return notebook._update_link_tag(node, new_path, target, oldhref)
  File "/home/introt/repos/zim-desktop-wiki/zim/notebook/notebook.py", line 842, in _update_link_tag
    short = newhref.parts()[-1]
IndexError: list index out of range
```

roland-ruedenauer:
@introt thanks for testing the changes so quickly. You surprised me.
Unfortunately I wasn't able to reproduce the problem you ran into in my local testing. I tested with having horizontal bars displayed in the ToC pane and without. Can you maybe have a look what's contained in new_href when the exception is raised or provide your test page content?
I also wonder how I could have broken copy & paste of links to headings from the page view directly.
Did you copy from the ToC pane immediately after changing the page content? In this case I could imagine we might run into problems, because there's a significant delay before the ToC get's updated after you modify the page content.

roland-ruedenauer:
While testing my changes with the new release 74.1 I observed a few minor issues
* ToC should always be above other TextView sibling windows. Having a source-view item on the page the ToC pane on first display get's overlapped by source-view content.
* Moving the mouse inside the ToC pane causes unexpected display of link info details (Go to "...") in the underlying pageview when there's a link on the page at the same relative y-pos of the mouse pointer.

introt:
No problem! Really excited for this feature, as it speeds up my current workflow quite a bit.

I'm in the middle of my Monday here, but was able to get some insight as well as a logging error for good measure :D

Here's my quick'n'dirty logging setup, stripped down to essentials:

```python3
    def _update_link_tag(self, elt, source, target, oldhref):
        if oldhref.rel == HREF_REL_ABSOLUTE: # prefer to keep absolute links
            newhref = HRef(HREF_REL_ABSOLUTE, target.name)
        elif source == target and oldhref.anchor:
            newhref = HRef(HREF_REL_FLOATING, '', oldhref.anchor)
        else:
            newhref = self.pages.create_link(source, target)
        logger.debug("newhref:", newhref.parts(), newhref.to_wiki_link())

        newhref.anchor = oldhref.anchor

        link = newhref.to_wiki_link()

        logger.debug("elt:", elt.gettext(), "oldref:", oldhref.to_wiki_link(), oldhref.parts())
        if elt.gettext() == elt.get('href'):
            elt[:] = [link]
            logger.debug("gettext==href")
        elif elt.gettext() == oldhref.parts()[-1] and len(elt) == 1:
            # We are using short links and the link text was short link
            # and there were no sub-node (like bold text) that would be cancelled.
            # Related to 'short_links' but not checking the property here.
            logger.debug("gettext==last part")
            short = newhref.parts()[-1]
            if newhref.anchor:
                logger.debug("newhref.anchor")
                short += '#' + newhref.anchor
            elt[:] = [short]  # 'Journal:2020:01:20' -> '20'
        else:
            logger.debug("neither")

        elt.set('href', link)

        return elt
```

Relevant log snippets with PR:
```python3
    logger.debug("newhref:", newhref.parts(), newhref.to_wiki_link())
Message: 'newhref:'
Arguments: ([], '#subsub')
--- Logging error ---
```

```python3
    logger.debug("elt:", elt.gettext(), "oldref:", oldhref.to_wiki_link(), oldhref.parts())
Message: 'elt:'
Arguments: ('subsub', 'oldref:', 'test:subtest:subsub#subsub', ['test', 'subtest', 'subsub'])
DEBUG: gettext==last part
```

The same log parts on `develop`:

```python3
    logger.debug("newhref:", newhref.parts(), newhref.to_wiki_link())
Message: 'newhref:'
Arguments: ([], '#subsub')
--- Logging error ---
```

```python3
    logger.debug("elt:", elt.gettext(), "oldref:", oldhref.to_wiki_link(), oldhref.parts())
Message: 'elt:'
Arguments: ('subsub\n', 'oldref:', 'test:subtest:subsub#subsub', ['test', 'subtest', 'subsub'])
DEBUG: neither
```

See the difference? I may not have figured it out if it weren't for the logging error, as I started by logging `elt`, not `elt.gettext()` and the line wrapped very similarly. I haven't the slightest clue about where `elt` comes from, but looking through your changes, one of them seems especially _sus_, as the kids would say.

zim/gui/pageview/__init__.py
```diff
@@ -1390,18 +1390,18 @@ def _get_close_anchor_or_object_id(self, iter):
...
+	def get_heading_text(self, iter):
 		line_start = iter.copy() if iter.starts_line() else self.get_iter_at_line(iter.get_line())
 		is_heading = any(filter(_is_heading_tag, line_start.get_tags()))
 		if not is_heading:
 			return None
 
 		line_end = line_start.copy()
 		line_end.forward_line()
-		return line_start.get_text(line_end)
+		return line_start.get_text(line_end).strip()
```
introt, line 1404: This is the breaking change, as confirmed by the issue getting resolved by taking it out. I agree that stripping should totally be applied here, but as we've seen, it does expose at least one bug elsewhere.

So: do we leave it out, or brace ourselves for a bunch of issues that may creep up sooner or later? I reckon it's up to Jaap to decide. Either way, changes need to be made in one direction or the other.

introt (changes requested):

zim/gui/pageview/__init__.py
roland-ruedenauer: Hi,
please check if  the change I commited solves the issue you ran into.

zim/notebook/notebook.py
```diff
@@ -835,7 +835,7 @@ def _update_link_tag(self, elt, source, target, oldhref):
 
 		if elt.gettext() == elt.get('href'):
 			elt[:] = [link]
-		elif elt.gettext() == oldhref.parts()[-1] and len(elt) == 1:
+		elif oldhref.parts() and elt.gettext() == oldhref.parts()[-1] and len(elt) == 1:
```
introt, line 838: I think you meant to type `newhref` instead of `oldhref` here - that fixes the issue.

introt (changes requested):

introt:
Forgot to ask about this yesterday - did you look into implementing drag'n'drop? Would speed up the flow further by not having to to open the menu and triggering paste separately. Don't think we're in a hurry here - new features w/ possible side effects may need to wait till 0.74 is "stable" before getting merged into develop for inclusion in 0.75, if I've read Jaap correctly.

zim/notebook/notebook.py
roland-ruedenauer: I added a check for `newhref` being empty too (the case you ran into). While testing yesterday I had an issue related to `oldhref` which was obviously a different problem.
This would also mean I was not able to reproduce your initial problem. Anyway, the parts list maybe empty and we have to check for this case.

roland-ruedenauer:
I did not have a look into also supporting drag'n drop for now. I can have a look into this too but it could take a while until I find some time.

zim/notebook/notebook.py
```diff
@@ -835,11 +835,11 @@ def _update_link_tag(self, elt, source, target, oldhref):
...
 			# We are using short links and the link text was short link
 			# and there were no sub-node (like bold text) that would be cancelled.
 			# Related to 'short_links' but not checking the property here.
-			short = newhref.parts()[-1]
+			short = newhref.parts()[-1] if newhref.parts() else ''
```
introt, line 842: This makes pasting the heading produce a link like `#this` instead of `this` like all the other headings do. Imo they should all behave the same way; if that way is `this`, it can be achieved with `elif oldhref.parts() and newhref.parts() and ...` on line 838

introt:
> I did not have a look into also supporting drag'n drop for now. I can have a look into this too but it could take a while until I find some time.

Jaap seems to be making destructive changes in develop again (eg. migrating to `newfs`) so it might now make sense to get this wrapped up && merged before looking into d'n'd. Up to you, I really appreciate your efforts either way!

roland-ruedenauer:
Hi @introt, I rebased the PR after the many changes applied lately and fixed the failing tests.
Can you check if you copy&paste problem is still present?

Note also the fix Jaap commit in 88eef741356c9e0e47eb698d58302ac3dc853c79.

I also had a look into implementing drag-and-drop in the meantime. It's working for me principally - meaning I get the expected link created at the drop location. What's not ok, is the visual feedback indicating the drop position and the mouse pointer changing when entering/leaving the text area. It works for a short time but then all a sudden get's not updated correctly anymore.
This might be caused by zim itself as I see it working in a tiny test application. I don't see these issues resolved soon and excluded my DnD changes from the PR. It's better to tackle this separately.

introt (approved):
Thanks for the great work @roland-ruedenauer! Copying headers now works as expected both between pages and on the same page. I can also confirm that your changes fix  #1730 and #1769, including Move Selected Text. Code looks good to me based on a cursory scroll, I'll have to leave the API changes for Jaap to judge.

Re: drag'n'drop, could you leave the code up somewhere ~for me to use~ to aid with hunting down the issue? Thanks!

zim/gui/server.py
```diff
@@ -132,7 +132,7 @@ def start(self):
 		try:
 			uri = self.notebookcombobox.get_notebook()
 			if uri:
-				notebook, x = build_notebook(NotebookInfo(uri))
+				notebook, _ = build_notebook(NotebookInfo(uri))
```
jaap-karssenberg, line 135: idem dito - please avoid these style changes

zim/main/__init__.py
```diff
@@ -39,12 +38,14 @@
...
 
+from typing import Optional, Tuple
+
 
 class HelpCommand(Command):
 	'''Class implementing the C{--help} command'''
 
 	usagehelp = '''\
-usage: zim [OPTIONS] [NOTEBOOK [PAGE]]
+usage: zim [OPTIONS] [NOTEBOOK [PAGE [ANCHOR]]]
```
jaap-karssenberg, line 48: I would strongly prefer the ANCHOR argument to be provided via an option. I would suggest "--id" as the user interface also refers to these as "id" rather than "anchor".  

I'm OK with allowing the anchor as part of the PAGE argument, since the "#" is reserved and not allowed in page names.

zim/main/__init__.py
```diff
@@ -137,75 +138,82 @@ def get_default_or_only_notebook(self):
...
 
-		return notebook, page or uripage
-
+		return notebook, page or uripage, anchor
 
 class GuiCommand(NotebookCommand, GtkCommand):
 	'''Class implementing the C{--gui} command and run the gtk interface'''
 
-	arguments = ('[NOTEBOOK]', '[PAGE]')
+	arguments = ('[NOTEBOOK]', '[PAGE]', '[ANCHOR]')
```
jaap-karssenberg, line 208: Idem dito, prefer "--id" for the anchor argument

zim/main/__init__.py
```diff
@@ -658,6 +669,7 @@ def __init__(self):
 		self._log_started = False
 		self._standalone = False
 		self._windows = set()
+		self._arguments = None
```
jaap-karssenberg, line 672: What is the purpose of this attribute ? Can't find back where it is used.

zim/notebook/page.py
```diff
@@ -372,6 +372,10 @@ def __str__(self):
 	def parts(self):
 		return self.names.split(':') if self.names else []
 
+	def short_name(self):
```
jaap-karssenberg, line 375: Good addition! Cleans up a bit of jargon in different places.

zim/plugins/tableofcontents.py
```diff
@@ -76,7 +75,10 @@ def get_headings(parsetree, include_hr):
 		if t[0] == HEADING:
 			level = int(t[1]['level'])
 			text = tokens_to_text(
-						collect_untill_end_token(tokens, HEADING) ).strip()
+				filter_token(
```
jaap-karssenberg, line 78: This fix is already solved another way - please see commit 5b87aa73e3c96772c81dab692a33ef1acbc2c42e

zim/tokenparser.py
```diff
@@ -21,9 +21,16 @@ class EndOfTokenListError(AssertionError):
 	pass
 
 
-def collect_untill_end_token(token_iter, tag):
+def collect_until_end_token(token_iter, tag):
```
jaap-karssenberg, line 24: Good catch on the typo in the function name.

Would prefer to revers this function to the implementation without a generator. Function is small and low level so called often in parsing so rather keep it a single function call. In this better to have a bit of redundancy in functions

zim/tokenparser.py
```diff
@@ -32,12 +39,26 @@ def collect_untill_end_token(token_iter, tag):
...
 				break
-
-		tokens.append(t)
+		yield t
 	else:
 		raise EndOfTokenListError('Did not find "%s" closing tag' % tag)
 
-	return tokens
+
+def filter_token(token_iter, token):
```
jaap-karssenberg, line 47: Good addition to the toolkit. Please leave this function in, even if the patch doesn't need it anymore (due to fix in 5b87aa73e3c96772c81dab692a33ef1acbc2c42e).

Thanks!

zim/plugins/tableofcontents.py
```diff
@@ -396,8 +424,30 @@ def on_populate_popup(self, treeview, menu):
 			else:
 				item.set_sensitive(False)
 
+		sep = Gtk.SeparatorMenuItem()
+		menu.append(sep)
+
+		item = Gtk.MenuItem.new_with_mnemonic(_("Copy _link to this location"))
```
jaap-karssenberg, line 430: Minor: for translation strings, please add a comment like "# T: purpose of string" this is picked up as info for translators

zim/plugins/tableofcontents.py
```diff
@@ -396,8 +424,30 @@ def on_populate_popup(self, treeview, menu):
...
+		menu.append(item)
+		link_details = self.can_copy_link_to_anchor(paths)
+		if link_details:
+			item.connect('activate', self.on_copy_link_to_anchor, *link_details)
+		else:
+			item.set_sensitive(False)
+
+		# open in new window
+		item = Gtk.MenuItem.new_with_mnemonic(_('Open in New _Window'))
```
jaap-karssenberg, line 439: Idem dito: please add translation comment

jaap-karssenberg (changes requested):
Hi Roland!

Sorry it took me some time to get around this review. Due to the whole covid thing lot of spare time got sucked away into child care and work ..

Added a bunch of comments, but overall I very much like the improvements you are making 👍 

The main blocking item for me is the commandline syntax, but open for discussion. You might want to file separate merge request for the ToC and the commandline arguments. That allows to discuss separate. But also OK to keep them together as long as it are two commits.

Thanks !

Jaap

introt:
I've split token parsing and HRef.short_name into new PRs #1947 and #1948; these include the requested changes (as well as extra fixes & related tests in the latter).

Besides those, this PR includes the following which should be split off before closing:

- [ ] linking related UX in ToC pane (in case @roland-ruedenauer stays MIA)
  - [ ] the very welcome addition of `strip()` to `get_heading_text`
- [ ] the `anchor` arg, which imo should just be included as part of the `page` (which itself should be renamed to `url` or `link` or something)
  * in addition to 458688c2042602ba4a0c94dab8b1e11f9008911d, this would also benefit from having something "proper check link syntax, including achor part" --> Discuss at #1942

introt:
Keeping this open for the anchor stuff, ToC functionality is now in #1952

introt:
Aforementioned anchor stuff handled by f2f6565b90f01f30978771d8d2f1b85628d475ce

PR #1627, opened 2021-09-26
Fix html:// to https:// in Link manual page

maclomhair:
Currently, the docs refer to a html:// page, probably http:// or https:// was ment.


introt (approved):
Yes, that makes a lot more sense- `html://` isn't an URL protocol. Good catch, thank you!

PR #1732, opened 2021-11-07
Use LC_TIME locale to determine first day of week

layus:
As per https://docs.python.org/3/library/locale.html `setlocale` needs to be called to initialize getlocale() from the user environment. Then we can lookup the right locale setting. `LC_TIME` is the proper one here, not the default locale (`LANG` or `LC_ALL`)

:warning: not tested, submitted directly from github edit form

See also https://stackoverflow.com/questions/48270191/python-does-not-use-locale-from-environment.

### DIFF
--- a/zim/datetimetz.py
+++ b/zim/datetimetz.py
@@ -15,6 +15,7 @@ with weeknumbers correctly.
 
 import re
 import locale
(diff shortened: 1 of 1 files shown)


introt:
Please test this and reopen against `develop`. Thanks!

PR #1744, opened 2021-11-15
Equation Plugin: Configurable Image DPI

egrath:
One thing i always found annoying was that the output DPI of dvipng was left at it's default, which is 96 DPI. Depending on the Font (-size) set for the Wiki Pages, Equations can look way to small.

This PR makes the DPI configurable in the preferences dialog of the plugin.


jaap-karssenberg:
I'm open to add this option. However please change the implementation such that it does not rely on the sorting order of the "`dvipngcmd`" hash. Not entirely sure, but I think hash sorting behavior might not be guaranteed over python releases.

egrath:
Made the handling of external program execution more robust by simplifying it. Also changed the latex command execution to match the mechnism used by dvipng for consistency reasons.

jaap-karssenberg:
Thanks!

Issue #1486, opened 2021-06-06
Ability to open multiple notebooks within same instance and move between them


itsnotsaved:
Currently each notebooks open in separate instance but when open many notebooks then it's quite difficult to move between them within dock or taskbar.

Add option to open and move between multiple notebooks within same instance.

When open new notebook then provide option to choose for open within the instance or separate instance

jaap-karssenberg:
What version are you looking at?  Latest versions by default open all
notebooks in a single process.

itsnotsaved:
It seems i have asked the question incorrectly. Zim opens all notebooks in single process.
I concerned about each notebooks open in separate window (it also suitable; i have misunderstood once again)

I mean about onenote's "Group" feature, i'm not sure if there any existing issues for it
https://support.microsoft.com/en-us/office/create-a-section-group-in-onenote-47c33260-fdb8-4ae3-9c6b-39407bf1cba3

Each group can have each index (side panel) then if there are related contents/projects etc that difficult to maintain within single index so we can keep them all as groups without create completely separate notebooks.

The problem, i have so many main pages list with subpages; and i can't keep them in single notebook even if they're related to same project. Then i created separate notebooks although only related to one project while i have other projects notebooks as well then finally i have lot of opened notebook windows so i can't even recognize easily or find necessary notebook to maximize. 

I see onenote address this problem by offering "group" feature then it may be possible to implement for zim software.

Zim 0.73.5-t202105070815

jaap-karssenberg:
Maybe I fail to understand the use case, but why not just add another level
of nesting the pages in your main notebook?  You can create as many levels
of sub-pages as you need to be organized.

-- Jaap

itsnotsaved:
>You can create as many levels of sub-pages as you need to be organized.

yes i know it^^

I use "building" as example.

e.g. **[A Building]** let's say i have 40+ main pages with multiple subpages (e.g.5-10) for each. when has around 30 folders then it fills the left panel index screen in 15inch laptop screen (i know its possible to keep creating 60-70) but when has such amount in one index, it make complicate to switch to other pages or find them quick when working with multiple main pages (subpages in each etc) often. Furthermore when keep expanded several main pages then it makes super complicated

![Screenshot from 2021-06-19 03-40-04](https://user-images.githubusercontent.com/73447325/122653978-a6cf1c00-d172-11eb-9136-13cc9099569a.png)


[...]

PR #1750, opened 2021-11-18
Journal plugin: added option to show all files from the notebook in the Journal tab

mtill:
I would prefer an option to include all the files and folders on the Journal page instead of just showing those that are directly related to the Journal itself. This makes it easier to switch between the files as the user does not need to switch between the journal pane and the pageindex pane.

### DIFF
--- a/zim/plugins/journal.py
+++ b/zim/plugins/journal.py
@@ -120,6 +120,7 @@ Also adds a calendar widget to access these pages.
 	plugin_notebook_properties = (
 		('namespace', 'namespace', _('Section'), Path(':Journal')), # T: input label
 		('granularity', 'choice', _('Use a page for each'), DAY, (DAY, WEEK, MONTH, YEAR)), # T: preferences option, values will be "Day", "Month", ...
+		('show_all_files', 'bool', _('Show all files from notebook'), False), # T: preferences option, show or hide all other files not related to the Journal
 	)
 
 	def path_from_date(self, notebook, date):
@@ -233,6 +234,8 @@ class JournalNotebookViewExtension(NotebookViewExtension):
 		self.notebook = pageview.notebook
 		self.calendar_widget = CalendarWidget(plugin, self.notebook, self.navigation)
 		self.connectto(pageview, 'page-changed', lambda o, p: self.calendar_widget.set_page(p))
+		if pageview.page is not None:
+			self.calendar_widget.set_page(pageview.page)
 
 		properties = self.plugin.notebook_properties(self.notebook)
 		self.connectto(properties, 'changed', self.on_properties_changed)
@@ -243,7 +246,7 @@ class JournalNotebookViewExtension(NotebookViewExtension):
 		self.disconnect_from(old_model)
 
 		index = self.notebook.index
-		namespace = properties['namespace']
+		namespace = None if properties['show_all_files'] else properties['namespace']
 		new_model = PageTreeStore(index, root=namespace, reverse=True)
 		self.calendar_widget.treeview.set_model(new_model)
 


jaap-karssenberg:
Please explain a bit more about the issue you are trying to fix. From my point of view, the whole point of the journal pane is to show only the journal and focus on that organized by date. What do you gain by adding the whole index in there ?

mtill:
Thank you for your kind reply.

I'm using ZIM on a daily basis, both for 1) taking notes from phone calls and meetings, and for 2) maintaining some kind of knowledge base. For 1), I'm using the Journal plugin, for 2), I'm switching to the PageIndex plugin. Switching between both plugins/panes is a bit inconvenient. E.g., when I switch to my knowledge base after taking some note after a phone call (i.e., switching from Journal to PageIndex), the Journal page is selected in PageIndex.

Instead, I would prefer to see all my notes from my knowledge base (which is not part of the Journal) directly after switching to PageIndex pane. This pull request implements an additional feature for the Journal plugin for showing all wiki pages in the Journal. This makes it easier to switch between journal and knowledge base when using, e.g., the bookmark plugin (selecting a date from journal calendar jumps to journal page, whereas clicking on bookmark jumps to knowledge base entry page, for instance.

Another option would be, maybe, to have some keyboard shortcut for showing some external calendar dialog coming up but as far as I know that is currently not implemented.

Best regards,
mtill

jaap-karssenberg:
If I translate what you describe, what you really are looking for is a
combination of the calendar widget with the regular page index, correct ?

Does that mean that you use the calendar widget to navigate in your journal
a lot, but not the journal index ?

-- Jaap

mtill:
> If I translate what you describe, what you really are looking for is a
combination of the calendar widget with the regular page index, correct ?

yes!

> Does that mean that you use the calendar widget to navigate in your journal
a lot, but not the journal index ?

yes, I'm not using the Journal Index.

I think in earlier versions the journal plugin was  implemented a bit differently and as far as i understood, something like an additional calendar dialog was also implemented ?  
https://www.schiessle.org/img/articles/keep-yourself-organized-zim.png

BR
mtill

jaap-karssenberg:
Yes in the past the journal only had a calendar widget and it was packed
above the index. However this was changed because we got more and more
plugins and that interface did not scale. So now each plugin had its own
pane.

Would like to keep the index view and the journal view separated. Would it
help for you to e.g. have a key binding to switch between the panes? Or
e.g. the option that Ctrl-D triggers a dialog with a calendar instead of
immediately going to the last date? Or maybe when you go to a journal page
switch to the calendar pane automatic and switch back when you navigate
away?

As a clarifying question: when using the calendar widget, do you often
navigate to older journal pages, or do you use it to get an overview of the
current week /month while working? Or to ask differently: do you use the
calendar for navigation or for information?

Op vr 3 dec. 2021 19:28 schreef mtill ***@***.***>:

mtill:
Hi!

I think this would be the ideal solution:
"option that Ctrl-D triggers a dialog with a calendar instead of
immediately going to the last date" 

> Or maybe when you go to a journal page
switch to the calendar pane automatic and switch back when you navigate
away?

what i think is a bit annoying is that the journal page is selected/activated in the PageIndex. 

> As a clarifying question: when using the calendar widget, do you often
navigate to older journal pages, or do you use it to get an overview of the
current week /month while working? Or to ask differently: do you use the
calendar for navigation or for information?

I'm using the calendar pane in order to navigate between dates - e.g. to switch to the journal page for last monday where i kept all the notes/meeting minutes for that day etc.

mtill:
Jaap, just another thought:

how about implementing it this way:
- whenever a Journal page is opened, the Journal plugin is activated.
- whenever some other page is openend, the PageIndex is activated.

this behavior should be implemented as an option.

So whenever you press ALT-D in order to jump to the current day, ZIM then directly jumps to the journal page in the journal plugin. When pressing ALT-HOME for jumping to the "Home" Page, the PageIndex is opened.

BR
mtill

introt:
> what i think is a bit annoying is that the journal page is selected/activated in the PageIndex.

(related: #1486 (as in having a separate "jounal section" from the rest of the notebook); esp. https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/1486#issuecomment-929564220)

mtill:
Hi!

based on our discussion yesterday, I thought  about all the different options here.

I provided a patch for the journal plugin to show pages from an arbitrary namespace that can be (optionally) specified by the user:
image

Meaning that if this option is set by the user, the index from the journal plugin will now show all pages from the specified index. Which makes things - at least from my point of view - so much easier as I can now use ZIM as a Journal and also see all my notes from my "knowledge base" on the very same page (without jumping between Journal pages and "knowledge base" pages in the PageIndex).

What do you think?
Kind regards
mtill

mtill:
after providing a patch for the pageindex, allowing the user to specify a namespace, i would propose to implement the mentioned solution in some of the following releases:

https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/1750#issuecomment-986005611

PR #1796, opened 2021-12-13
autoselect: don't select word starting from cursor; fixes #1768

### LINKED ISSUE #1768: toggle_format w/ autoselect doesn't toggle formatting when cursor is at the beginning of a word
**Describe the bug**
> |hello there

Pressing Ctrl+B bolds 'hello'; pressing it another time doesn't unbold it. Other formats like italics and emphasis can similarly be applied on top, but not removed.

**To Reproduce**
Here's how the test setup looks like in the beginning:
image

Here's the view after pressing Ctrl+B once; the "hello" goes bold, but bolding doesn't get toggled on:
image

After a second press the bold formatting is toggled on, but the "hello" stays bold (my tiny window also grew for some reason):
image

Subsequent presses just toggle the active formatting without affecting the "hello", eg. the view after the third press is identical to the one after the first one, as is the one after the fifth and so on.

Here's what it looks like after pressing Ctrl+Z (a separate issue):
image

**Expected behavior**
According to its docstring (and name), toggle_format should behave identically regardless of whether the cursor is at the beginning of a word or inside it.

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/4c68661c758457a4cbb08a99806c3958f2b5ea96/zim/gui/pageview/__init__.py#L7284-L7286

 - Version 0.74.2
 - OS: linux

**Additional context**
Discussed here, with other users able to reproduce: https://github.com/zim-desktop-wiki/zim-desktop-wiki/discussions/1731#discussioncomment-1602531

In other editors, an "autoselection" occurs only when the cursor is inside a word, not at its beginning; this seems to be a saner choice, as it allows prepending lines with formatted text without having to add a space before writing said text. **Please consider changing the autoselect behavior.**

introt:
Fixes #1768 by changing autoselect behavior to not select words starting from cursor; this is the default behavior of eg. LibreOffice Writer and Thunderbird.

### DIFF
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -3005,8 +3005,10 @@ class TextBuffer(Gtk.TextBuffer):
 		self.select_range(start, end)
 		return True
 
-	def select_word(self):
+	def select_word(self, selectstarting=True):
 		'''Selects the current word, if any
+
+		@param selectstarting: whether to select word that starts from cursor
 
 		@returns: C{True} when succcessful
 		'''
@@ -3015,7 +3017,11 @@ class TextBuffer(Gtk.TextBuffer):
 			return False
 
 		bound = insert.copy()
-		if not insert.starts_word():
+		if insert.starts_word():
+			if not selectstarting:
+				# do not select word starting from cursor
+				return False
+		else:
 			insert.backward_word_start()
 		if not bound.ends_word():
 			bound.forward_word_end()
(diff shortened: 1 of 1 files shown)


introt:
If this behavior is acceptable, need to 
- [ ] update docs
- [ ] add/modify tests

pfalcon:
>  autoselect: don't select word starting from cursor; fixes #1768
> Fixes #1768 by changing autoselect behavior to not select words starting from cursor; this is the default behavior of eg. LibreOffice Writer and Thunderbird.

Can you please make the description self-sufficient and understandable by other people as is? From the description above, it's unclear what you changed, and how it works now. ("don't select words starting from cursor" - ok, does it select words from the end of page now, or something?)

introt:
It augments `autoselect`'s behavior by introducing a new kwarg[0] to `select_word`[1] which makes it not select words that start from the cursor's current position, eg. whose first letter is to the right of the cursor (assuming LTR); I agree that this could be made clearer by the docs, but the linked issue and discussion should cover that for the time being.

0. `selectstarting: whether to select word that starts from cursor`
1. `Selects the current word, if any`

PR #1811, opened 2021-12-15
[Documentation] Fix #948; s/TRUE/True/g

### LINKED ISSUE #948: Error in documentation
Help/Templates says on "index()":

collapse: if TRUE only branches related to the current page are visible, if FALSE all branches are 
visible - defaults to TRUE

"TRUE" and "FALSE" sould be "True" and "Fasle", otherwise the export failes with an error mesage.

introt:
Fixes #948 and thus prevents many headaches.


introt:
Btw I don't see the correct syntax for function arguments - eg. `index(collapse, False)` - mentioned on the page; am I missing something?

PR #1821, opened 2021-12-18
Use assertEqual instead of assertEquals

cheese:
assertEquals is deprecated in Python 3.2 and removed in 3.11.


introt:
Good catch! Could you reopen this against `develop`? Thanks!

cheese:
> Good catch! Could you reopen this against `develop`? Thanks!

Done

Issue #64, opened 2017-05-24
improve search speed


jaap-karssenberg:
Page search is very slow compared to e.g. unix grep. 

Improve this by actually using grep as subprocess, or similar low-level read function to find "candidates". Only evaluate full query on these "candidates".

Osndok:
I didn't really have time to explore this in depth (as I could have just been connecting it wrong), but I was surprised find that the search callback (that presents the items found so far and allows the search to be cancelled) might actually be slowing down the search considerably. Again, **NOT** 100% sure...

quesada:
Ripgrep seems to be the one to beat. Maybe a plugin that does a system call to rg (instead of grep; or as an option, the same way we have 'import' and 'scrot' for screenshots) and displays the information as is now would work great and be a relatively low hanging fruit.

Self-Perfection:
Prefiltering with tools like grep o ripgrep overcomes CPU bottleneck, but there still might be IO bottleneck on slow storage, like rotary HDDs. Therefore I find myself usually searching in zim notebook with `git grep --cached` on my latptop. It does not need to read couple of thousand txt files from HDD and performs in less than a second. And for comparison grep runs ~20s if notebook is not cached in RAM.

mercurial also has grep tool, though I'm not sure if it can give similar speedup.

introt:
I've [ported](https://github.com/introt/zim-desktop-wiki/commit/28ed20e65c8157baea607587d1f027d23dc5d924) Osndok's patch to the latest version of Zim. It should work on Windows, as long as you've got `grep` in your path; let me know if it doesn't.

It should be noted that this version **replaces** the "native" implementation instead of complementing `_process_content` by prefiltering. This means your search string is used as grep patterns*; queries aren't processed. As an avid grep enjoyer, I'd like to present that as an option in the search dialog.

As for other options, making the grep base command (and thus executable) user editable in addition to plugin-style dependency checking should accommodate both `ripgrep` and `git grep --cached` users.

Not sure what's the next course of action towards a true implementation from here - @jaap-karssenberg should "grep search" be implemented in `zim.search` or as a plugin? Should "regular search" be made extendable, or replaceable by moving it into a plugin?

*multiple patterns are supported as-is - just copy and paste a newline ![image](https://user-images.githubusercontent.com/26069787/147549453-970cbfd9-163c-492f-a667-06471c2fe326.png)

jaap-karssenberg:
[...]

PR #1840, opened 2021-12-28
Fix NameError raised by self.notebook.info

introt:
`NotebookInfo` was not defined in `notebook.py`. It's imported here to avoid circular imports and also because this property doesn't seem to be used by any existing code - I bumped into this while trying to port the patch from #64 to the latest version.

Let me know if you want a test for this - imo this should be covered by a test of a feature which requires this property

### DIFF
--- a/zim/notebook/notebook.py
+++ b/zim/notebook/notebook.py
@@ -352,7 +352,7 @@ class Notebook(ConnectorMixin, SignalEmitter):
 			uri = self.uri
 		except AssertionError:
 			uri = None
-
+		from . import NotebookInfo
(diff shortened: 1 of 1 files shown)


zim/notebook/notebook.py
```diff
@@ -352,7 +352,7 @@ def info(self):
 			uri = self.uri
 		except AssertionError:
 			uri = None
-
+		from . import NotebookInfo
```
jaap-karssenberg, line 355: Please move this import to the top of the file

introt: Can't - it'd be a circular import and fail. Both `__init__` and `info` import stuff from `notebook`

Issue #1797, opened 2021-12-13
[Mac OS] automatic tests fail with segfault for Mac OS


jaap-karssenberg:
Automatic test suite as run on Github fails with segfault for Mac OS. Failure started after merging from master branch. See https://github.com/zim-desktop-wiki/zim-desktop-wiki/actions

Would need confirmation of offline run of the test suite on Mac OS and a `git bisect` to determine the change that causes this.

For info: a segfault is by default a bug in a lower level library or in the interpreter, not an issue raised within python. Therefore it is hard to troubleshoot. Hopefully the bisect narrows done the piece of code triggering it and might suggest a work around to avoid this.

dehesselle:
I've run the test suite a couple of times and I'm not getting a segmentation fault.

```plaintext
Ran 896 tests in 67.033s

FAILED (failures=49, errors=3, skipped=24, expected failures=21)
```

jaap-karssenberg:
Hmm, than it is specific for the GitHub environment - will be hard to
troubleshoot :(

dehesselle:
My current take on the situation: cannot promise a timeframe, but I'll make something from my CI reusable for GitHub so we can have automatic tests.

PR #1844, opened 2021-12-29
Disable automatic tests on Mac OS

introt:
until #1797 is fixed. This makes the check-marks actually useful for catching failing tests in the meantime.


introt:
Since the tests were ran according to the yaml in the PR, the checkmark here validates both the PR and the argument of making it useful again :)

PR #1847, opened 2021-12-30
Path.absname, absolute paths for insert date

introt:
Complements Path.relname and Path.basename. In addition, replaced
instances of "':' + link.name" with this new property and changed
links created by the "Insert Date and Time" dialog absolute.

### DIFF
--- a/zim/export/template.py
+++ b/zim/export/template.py
@@ -310,7 +310,7 @@ class ExportTemplateContext(dict):
 			else:
 				# links to other pages
 				builder.append(LINK,
-					{'type': 'page', 'href': ':' + path.name},
+					{'type': 'page', 'href': path.absname},
 					path.basename)
 			builder.end(LISTITEM)
 
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -7654,7 +7654,7 @@ class InsertDateDialog(Dialog):
 			text = model[0][self.DATE_COL]
 
 		if self.link and self.linkbutton.get_active():
-			self.buffer.insert_link_at_cursor(text, self.link.name)
+			self.buffer.insert_link_at_cursor(text, self.link.absname)
 		else:
 			self.buffer.insert_at_cursor(text)
 
--- a/zim/gui/widgets.py
+++ b/zim/gui/widgets.py
@@ -1773,7 +1773,7 @@ class PageEntry(InputEntry):
 		elif path.isroot:
 			self.set_text('')
 		else:
-			self.set_text(':' + path.name)
+			self.set_text(path.absname)
 
 	def get_path(self):
 		'''Get the path shown in the widget.
--- a/zim/notebook/page.py
+++ b/zim/notebook/page.py
@@ -227,6 +227,12 @@ class Path(object):
 		return self.name[i:]
 
 	@property
+	def absname(self):
+		'''Get the full, absolute path.
+		Returns ':' for the top level namespace.'''
+		return ':' + self.name
+
+	@property
 	def namespace(self):
 		'''Gives the name for the parent page.
 		Returns an empty string for the top level namespace.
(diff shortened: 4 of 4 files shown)


introt:
This addition could also be useful for implementing a "only use absolute links" option (see https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/492#issuecomment-479948692)

introt:
This produces wrong results; I assumed Paths know their absolute paths, which is false

jaap-karssenberg:
The attribute `path.name` is the full name of the path. The syntax `":" + path.name` creates wiki syntax for an absolute link. Probably the most clean way to handle this would be to use a `HRef` object rather than a path object ?

introt:
Looking back, my problem was creating `Path`s from wiki syntax for relative links, against the instructions in the documentation (`Never construct a path directly from user input`).

> Probably the most clean way to handle this would be to use a HRef object rather than a path object ?

Which part? I tried[0] handling `InsertDateDialog` with a `HRef` just now, ended up with `HRef(HREF_REL_ABSOLUTE, self.link.name).to_wiki_link()` which seems excessive for prepending a `:` - `self.link` being a `Path` instead of a `HRef` is a tad bit confusing.

I can't say I've wholly cracked the distinction between `Path` and `HRef`. Why not have `Path` represent absolute paths with eg. `from_relative_link(source: Path, target: str)` and `to_wiki_link(relative_source: Path=None)` to serve the `HRef` needs? Being half-rhetorical here, probs has something to do with either maintainability/performance/something I-can't-quite-articulate regarding relative/floating links or semantics. Though semantically I also find using `rel` in the context of "href" for location-relation instead of resource-relation confusing, but that's likely due to me not grasping the more abstract concepts and relating the terminology to html - I better stop here before I start questioning why we don't have a `Link` class.

0. https://github.com/introt/zim-desktop-wiki/tree/href-date

jaap-karssenberg:
The class "Path" should be renamed to "PageName" that makes it more clear
what it is exactly. It is an (absolute) page name, no more, no less. It
should not have any logic for handling links.

We do have a "Link" class - the index has an "IndexLink" class to represent
a link between two pages as stored in the index.

The class "HRef" represents a link as it appears in the wiki source, and
contains semantic information about the link type but does not contain an
end point. It was introduced to help with the logic of updating links when
moving pages while preserving the link type (relative, floating or
absolute).

To resolve a link you need to first create a HRef and then call
notebook.pages.resolve_link(). Which will just return the end point as a
Path, not a Link as you might expect.

If you are sure you need an absolute link, you can indeed just call ":" +
path.name, however it may not be robust in some far future where an
alternative syntax is supported.

There is also notebook.pages.create_link() if you want the most concise
link format.

Hope this helps,

Jaap

introt:
Thanks, that clears it up a lot!

`PageName` is (at least in this context) a lot more descriptive - and is used in some parameter names already. Are you open to a PR for this?

> If you are sure you need an absolute link, you can indeed just call ":" +
> path.name, however it may not be robust in some far future where an
> alternative syntax is supported.

":" + path.name is already used in a few places. Would the following addition to HRef make sense to you? Pagename shouldn't need another round of validation or resolving, and this allows the future-proof creation of absolute links without importing HREF_REL_ABSOLUTE.

```
@classmethod
def pagename_to_wiki_link(klass, pagename):
    '''Returns href for C{pagename} as text for absolute wiki link.'''
    #return ':' + pagename.name
    return klass(HREF_REL_ABSOLUTE, pagename.name).to_wiki_link()
```

I've opened #1853 for some related documentation improvements.

jaap-karssenberg:
For now the ":" + page name usage is fine. A convenience method on href is
also ok with me.

Open to PR to rename Path. Probably will keep Path as an alias to avoid
breaking 3rd party code for some time.

Op vr 7 jan. 2022 14:27 schreef introt ***@***.***>:

PR #1848, opened 2021-12-31
Upgrade Python version requirement to >= 3.6

introt:
Needed for type hints etc; 3.5 has been EOL for over a year, [the oldest Ubuntu LTS is on 3.6](https://packages.ubuntu.com/search?keywords=python3).

Previously discussed at #1776.

Includes the slightest bit of cleanup of old code, didn't touch eg. the regex patterns that have been held back by Python 2.5 (not a typo!).

Let's start a happy new year with a new(er) Python version! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ First, you should verify you have the dependencies zim needs. To list all depend
 You will at least need the following:
 
 * Gtk+ >= 3.18
-* python3 >= 3.2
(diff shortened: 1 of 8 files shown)


introt:
Might make sense to also adjust the automatic test matrix to include Python 3.6 on at least one platform to avoid using newer syntax like f-strings, but that's another issue/PR

PR #1856, opened 2022-01-10
Use assertEqual for Python 3.11 compatibility

hugovk:
`assertEquals` has been deprecated since Python 3.2 and is removed in 3.11, due out in October. Use `assertEqual` instead.

https://docs.python.org/3.11/whatsnew/3.11.html#removed


introt:
Thank you! We already have the fix in `develop` waiting to be merged into master - see #1821. Sorry for the inconvenience, need to update the contributing guidelines to include the branch priority. Hope to see more PRs from you in the future!

hugovk:
No problem!

It would be helpful to set `develop` as the default branch as well, then GitHub presents that when visiting the site, and for cloning, and most importantly as the target branch for PRs.

introt:
Feel free to drop those lines here: #1860; I'm not in control of that :)

PR #1877, opened 2022-01-21
Optimize testing

introt:
I'm not aware of changes that'd make the in-between versions break, as we're not doing anything version-tied stuff (try-except or __future__ imports etc)


introt:
Saw other people weren't using quotes, but of course they weren't using 3.10 either :P

PR #1878, opened 2022-01-21
Optimize testing

introt:
We're not doing any inbetween-version-spesific stuff, so stick to the oldest and newest versions on dev branch


introt:
I don't remember whether there was some problem with 3.6 workflows, but we'll see whether this PR gets the green tick

PR #1924, opened 2022-02-12
Source View plugin: Don't show hidden languages

### LINKED ISSUE #1921: Crash when editing code "Defaults" code block
**Describe the bug**
Typing anything a code block with syntax highlighting set to "Defaults" immediately crashes the program.

**To Reproduce**
1. Enable the "Source View" plugin if it is not already enabled.
2. Insert -> Code Block
3. Select "Defaults" in the syntax field of the dialog, confirm
4. Place the cursor in the newly created code block
5. Type anything
6. Observe crash

**Environment:**
 - Version 0.74.3, Windows binary installer from website
 - OS: Windows 10
 - Language: en-us

**Additional context**
Nothing is written to the debug log. I assume this is because Zim crashes before anything can be written.

Edit: This bug appears to be very similar to another issue on [Launchpad](https://bugs.launchpad.net/zim/+bug/1478215).

introt:
Fixes #1921

Ref: https://lazka.github.io/pgi-docs/GtkSource-3.0/classes/Language.html#GtkSource.Language.get_hidden

### DIFF
--- a/zim/plugins/sourceview.py
+++ b/zim/plugins/sourceview.py
@@ -44,7 +44,7 @@ if GtkSource:
 	lang_ids = lm.get_language_ids()
 	lang_names = [lm.get_language(i).get_name() for i in lang_ids]
 
-	LANGUAGES = dict((lm.get_language(i).get_name(), i) for i in lang_ids)
+	LANGUAGES = dict((lm.get_language(i).get_name(), i) for i in lang_ids if not lm.get_language(i).get_hidden())
 
 	ssm = GtkSource.StyleSchemeManager()
 


introt:
Too bad we aren't on 3.8, would've loved a walrus here `:=`

introt (commented):
In addition to hiding `Defaults`, we'll lose

```
gtk-doc
Haddock
JavaScript Built-in Values
JavaScript Expressions
JavaScript Functions and Classes
JavaScript Literals
JavaScript Modules
JavaScript Statements
JSDoc
TypeScript JavaScript Expression Additions
TypeScript JavaScript Function and Class Additions
TypeScript JavaScript Literal Additions
TypeScript JavaScript Module Additions
TypeScript JavaScript Statement Additions
TypeScript Type Expressions
TypeScript Type Generics
TypeScript Type Literals
```

I've tried a few of these, and they didn't result in a crash. Should we instead just pop Defaults out of LANGUAGES, or implement some sanity checks (if lang.prop != None)?

dehesselle:
It is ok and intended behavior to not show any hidden languages. We're not going to lose anything. All the `.lang` files with the `hidden` attribute exist for modularization purposes, they are not standalone language definitions. All their content gets included by "real" language files. For example, you'll see that there is a `javascript.lang` which references all the content from the other files.

I've checked all of them:
- gtk-doc is used by `c.lang`
- Haddock is used by `haskell.lang`
- JavaScript* are used by `javascript.lang`
- JSDoc is used by `javascript.lang`
- TypeScript* are used by `typescript.lang`

introt:
Thanks for chiming in, I wasn't sure whether that was the case with the first two!

PR #1925, opened 2022-02-13
Insert Symbol: allow "custom autocorrect"

introt:
Phrases don't show up in the dialog, since they aren't added to the
symbol order list. Lines can be broken using \n, and escapes are
provided to allow literal \n as well as non-comment #.

In my opinion, text replacements should be a core feature, and I see
no point in needlessly duplicating the end_of_word check between
multiple plugins. This solution is quite clean, and could be made
user-friendlier by having a separate dialog and config file for the
phrases while keeping the shared replacement logic.

Closes zim-desktop-wiki/zim-desktop-wiki#846


jaap-karssenberg:
Thanks!

PR #1926, opened 2022-02-13
Fix NameError in _update_new_file_submenu

introt:
Came across this in the logs while testing:

```
DEBUG: Store page in background: Home:Page name with spaces
DEBUG: Index file: /home/introt/Notebooks/Notes/Home/Page_name_with_spaces.txt
  File "/home/introt/repos/zim-desktop-wiki/zim/gui/pageview/__init__.py", line 7204, in _update_new_file_submenu
    name = file.basename
NameError: name 'file' is not defined
```

### DIFF
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -7200,7 +7200,7 @@ class PageView(GSignalEmitterMixin, Gtk.VBox):
 			def handler(menuitem, file):
 				self.insert_new_file(file)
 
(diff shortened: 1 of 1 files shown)


introt:
@datenwurm do you think this might be related to #1822? I've done no digging nor testing, but this is the function you were pondering about calling

datenwurm:
@introt thanks, i'll check it out and ping back when i know more!

PR #1928, opened 2022-02-13
Fix German translation for score

dehesselle:
This PR fixes a mistranslation in the German localization.

"score" as in music, not "score" as in grading.

### DIFF
--- a/translations/de.po
+++ b/translations/de.po
@@ -1810,7 +1810,7 @@ msgstr "Suche auf die aktuelle Seite und Unterseite(n) einschränken"
 #. menu item
 #: zim/gui/searchdialog.py:150 zim/plugins/scoreeditor.py:68
 msgid "Score"
-msgstr "Bewertung"
+msgstr "Noten"
 
 #. Window title
 #: zim/gui/server.py:46


introt (commented):
Thanks! Could you please add the context clue ("`"score" as in music, not "score" as in grading.`") to the plugin source as well?

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/3e9064f2dda99b66e26de8adbf566b6fef1b32c5/zim/plugins/scoreeditor.py#L68

dehesselle:
Isn't this sufficient?
https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/3e9064f2dda99b66e26de8adbf566b6fef1b32c5/zim/plugins/scoreeditor.py#L4-L5

introt:
No, because translators using the web ui only see the following context, albeit surrounded by a fancy gooey:

```
#. menu item
#: zim/gui/searchdialog.py:150 zim/plugins/scoreeditor.py:68
msgid "Score"
```

(which is most likely how we ended up here; I remember another issue being caused by a translator mistakenly translating the context clue instead of the string)

translations/de.po
```diff
@@ -1810,7 +1810,7 @@ msgstr "Suche auf die aktuelle Seite und Unterseite(n) einschränken"
 #. menu item
 #: zim/gui/searchdialog.py:150 zim/plugins/scoreeditor.py:68
```
introt, line 1811: Took me a while to notice, but the translation is correct - for the search result scoring. The plugin needs to be modified.

introt (changes requested):

translations/de.po
introt: I suggest changing the visible string to "Music Score", because it's an item in a menu and should be unambiguous in any language

introt: @dehesselle yes it does:

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/a9917c94a033a2ebf57bcd4ee111f023b14b03a0/translations/de.po#L1809-L1813

jaap-karssenberg:
Fixed the translation key to "Music Score" - you should now be able to set proper German translation via the Weblate interface. Thanks for the input!

PR #1930, opened 2022-02-13
Score editor: add check for convert-ly

introt:
On Windows, users may have the lilypond executable in their path
while lacking convert-ly. This adds a check for the availability
of the latter to prevent activating the plugin while lacking the
required dependencies.

Closes #1923

### DIFF
--- a/data/manual/Plugins/Score_Editor.txt
+++ b/data/manual/Plugins/Score_Editor.txt
@@ -5,9 +5,9 @@ Wiki-Format: zim 0.4
 
 The score editor is a simple dialog that allows you to insert a score into a page using **GNU Lilypond**.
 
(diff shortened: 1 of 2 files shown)


introt:
Re: #1923

> _After proper activation of the plugin (and installed dependencies)_, **I cannot find the way to access this dialog.** The manual does not explain how to insert the musical scores.

Emphasis mine; the italics are handled by this PR while #1928 should address the bold part by adjusting the translation. However, the manual does seem to be written with the presumption that the user will know to look into the `Insert` menu to find the _Insert_ plugin dialogs - which may not be the case when the UI language is not English.

This issue isn't limited to this plugin, and might not be an actual issue if the plugin name and menu item actually match each other, but is something that I feel like we should try to address in the future. Could be as simple as adding actual context (instead of just `#T: menu item`) to the translation strings and ellipsis to all of the Insert Plugin menu items (which they currently lack, which btw breaks UX) to force the translations to be updated.

PR #1957, opened 2022-02-26
Prevent autoformatting of bullets during verbatim

introt:
`/^* /` would produce a bullet while the verbatim format was active.

Source text produced before this PR after selecting the verbatim formatting and typing `* test`:
`* test` (the bullet gets autoformatted and the verbatim format gets ditched)

New behavior:
`''* test''` (text typed in verbatim is displayed verbatim)

### DIFF
--- a/tests/pageview.py
+++ b/tests/pageview.py
@@ -2106,6 +2106,23 @@ foo
 		#~ tree = buffer.get_parsetree(raw=True)
 		#~ self.assertEqual(tree.tostring(), wanted)
 
+		# Test autoformatting of bullets when verbatim format is selected (#1957)
+		press(view, '\n')
+		buffer.toggle_format_tag_by_name('code')
+		press(view, '* test')
+		wanted = '''\
+<?xml version='1.0' encoding='utf-8'?>
+<zim-tree raw="True">aaa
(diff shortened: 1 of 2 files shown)


introt:
Apparently this change induced time travel:

```
FAIL: testFileInfo (tests.newfs.TestLocalFS)
  File "/home/introt/repos/zim-desktop-wiki/tests/newfs.py", line 249, in testFileInfo
    self.assertGreaterEqual(f.mtime(), f.ctime())
AssertionError: 1645884489.4548044 not greater than or equal to 1645884489.4558043
```

And broke copying of the current line:

```
FAIL: testCopyCurrentLine (tests.pageview.TestPageViewActions)
  File "/home/introt/repos/zim-desktop-wiki/tests/pageview.py", line 3630, in testCopyCurrentLine
    self.assertEqual(get_text(buffer2), 'test 456\n')
AssertionError: '' != 'test 456\n'
+ test 456
```

I'll look into this later - was quicker to draft a PR than to submit the original issue :)

introt:
...and now it passes all the tests on both my machine as well as the GitHub ones. Weird. Will add test to prevent regression before readying for review.

PR #1959, opened 2022-02-26
Added mnemonics to collapse all entries in PageIndex

mtill:
Added mnemonic to collapse all entries in PageIndex (without the need to select PageIndex pane beforehand).

I am aware that you can press "\" in PageIndex pane, but for that to work, one first needs to select the PageIndex pane which I found is inconvenient in daily use.

Any feedback welcome!

### DIFF
--- a/zim/plugins/pageindex/__init__.py
+++ b/zim/plugins/pageindex/__init__.py
@@ -1,6 +1,7 @@
 
 # Copyright 2008-2018 Jaap Karssenberg <jaap.karssenberg@gmail.com>
 
+from nturl2path import pathname2url
 from gi.repository import GObject
 from gi.repository import Gtk
 from gi.repository import Gdk
@@ -17,7 +18,7 @@ from zim.notebook.index.base import TreeModelMixinBase
 from zim.notebook.index.pages import PagesTreeModelMixin, PageIndexRecord, IndexNotFoundError, IS_PAGE
 
 from zim.plugins import PluginClass
-from zim.actions import PRIMARY_MODIFIER_MASK
+from zim.actions import PRIMARY_MODIFIER_MASK, action
 
 from zim.gui.notebookview import NotebookViewExtension
 from zim.gui.widgets import BrowserTreeView, ScrolledWindow, \
@@ -125,6 +126,10 @@ class PageIndexNotebookViewExtension(NotebookViewExtension):
 		self.treeview.disconnect_index()
 		model = PageTreeStore(self.pageview.notebook.index)
 		self.treeview.set_model(model)
+
+	@action(_('Collapse All'), accelerator='<Alt>K', menuhints='view') # T: menu item
+	def collapse(self):
+		self.treeview.collapse_all()
 
 
 class PageIndexWidget(Gtk.VBox, WindowSidePaneWidget):
--- a/zim/plugins/pageindex/__init__.py
+++ b/zim/plugins/pageindex/__init__.py
@@ -131,6 +131,17 @@ class PageIndexNotebookViewExtension(NotebookViewExtension):
 	def collapse(self):
 		self.treeview.collapse_all()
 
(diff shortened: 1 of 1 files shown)


mtill:
Added mnemoric to unfold/select current page in PageIndex (useful e.g. when not using the autoexpand feature of the PageIndex Plugin)

zim/plugins/pageindex/__init__.py
```diff
@@ -1,6 +1,7 @@
 
 # Copyright 2008-2018 Jaap Karssenberg <jaap.karssenberg@gmail.com>
 
+from nturl2path import pathname2url
```
jaap-karssenberg, line 4: What is this import used for?

zim/plugins/pageindex/__init__.py
```diff
@@ -126,6 +127,21 @@ def reload_model(self):
 		model = PageTreeStore(self.pageview.notebook.index)
 		self.treeview.set_model(model)
 
+	@action(_('Collapse All'), accelerator='<Alt>K', menuhints='view') # T: menu item
```
jaap-karssenberg, line 130: Be aware that if the label is shown in the generic "view" menu, it won't be clear that "collapse all" relates to the index.

More general, afraid this won't scale - can't have a window-wide accelerator like this for each side pane. So probably should be a keybinding that is activated only when the index pane has focus.

zim/plugins/pageindex/__init__.py
mtill: fixed - thanks!

zim/plugins/pageindex/__init__.py
mtill: I see two options:
1) one option would be to remove it completely from "View" Menu (and only keep it as a shortcut)
2) another option would be to simply rename it ("Collapse Page Index") - I just changed that in the latest commit

I would really vote for an option that works independently from the current focus - as this should, from my point of view, work while one is typing in the ZIM notebook/editor as one does not need to change focus first to Pageindex.

BR

mtill: PS: do you agree with the other feature that is part of the pull request:
"Added Mnemonics to unfold/select current page in PageIndex (useful e.g. when not using the autoexpand feature of the PageIndex Plugin)" ?
Thank you for your feedback & best regards

jaap-karssenberg:
For both I agree on having key bindings, just hesitant to make it window
generic "accelerators" - believe it should be keybindings on the widget.

Btw, collapse all should already work with "\" as the keybinding if I'm not
mistaken.  Maybe easiest thing is to just expand on that and let "/" expand
to current page.

I can imaging the use of an accelerator that does "switch focus to side
pane" + "expand to current page" - but then it may need to be shared
between plugins - e.g. for a journal page I would want to focus the journal
page. If each plugin has it's own accelerator, we run out of available keys
pretty quickly I'm afraid.

Maybe look into the "switch to side pane" key binding (Ctrl-Space by
default) and tie a plugin specific action to this ?

-- Jaap

mtill:
Tried with "Ctrl-Space" + "\" but unfortunately, this does not seem to work reliably in case both PageIndex and Journal are activated in parallel (PageIndex shown on left side, Journal on the right).
Also, it would be nice to just have one keyboard shortcut for this action, not two.
Can we maybe implement it as proposed and not assign anything by default ? So the user can decide whether he/she wants to use this feature and whether to assign a keyboard shortcut or not. This way, it would also scale as it is left to the user to decide which feature to use

jaap-karssenberg:
Agree, a action without default keybinding is a good compromise for now -
might break it later though if I come up with something that can easier
generalize.

-- Jaap

mtill:
great! looking forward to have it merged into the upcoming release

jaap-karssenberg:
Can you change the PR to set default accelerator to None ?

Thanks,

Jaap

mtill:
sure - i just did: default accelerator was set to None
Best regards
mtill

jaap-karssenberg:
Merged

PR #1961, opened 2022-02-28
Use LC_TIME locale to determine first day of week

rolfkleef:
As per https://docs.python.org/3/library/locale.html `setlocale` needs to be called to initialize getlocale() from the user environment. Then we can lookup the right locale setting. `LC_TIME` is the proper one here, not the default locale (`LANG` or `LC_ALL`)

:warning: not tested, submitted directly from github edit form


rolfkleef:
As [per discussion](https://github.com/zim-desktop-wiki/zim-desktop-wiki/discussions/1954#discussioncomment-2245571): this replaces #1732 (now targeted against `develop`).

It seems to have worked in my case, but I'm not sure this change did that. If I revert the change, and run `develop` again before this patch, it still works as expected. Trying to reproduce it...

rolfkleef:
Ok, when reviewing my testing I realised I installed the Babel library "to be sure", but apparently didn't have it installed before.

There is a note about this optional library in the Journal plugin manual page, to help with weekday conventions.

I have turned this into a very different code suggestion:

* The datetime function would not log a debug message if there is an import error for the Babel library: I think making that explicit would help.
* It would be helpful to have the same remark about Babel in the information dialog for the Journal plugin, and to perhaps also show whether the import can be done, as dependency.

PR #1971, opened 2022-03-02
Further test matrix optimization

introt:
We use msys2's Python on Windows, so no point in setting up nor running regular Python there.


introt:
Bad quotes.

PR #1480, opened 2021-05-29
LinkMap: add refresh button, opening links in new windows & preferences for zoom and link opening

introt:
See the manual for more information.

PR #1972, opened 2022-03-02
Optimize GitHub Workflows testing matrix

introt:
We use msys2's Python on Windows, so no point in setting up nor running regular Python there.

Separate windows-latest to simplify setup, as it shares no steps with the unix-like OSes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,21 +8,21 @@ on:
       - develop
 
 jobs:
-  build:
+  unix-build:
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


introt:
In addition to reducing our carbon footprint, this should reduce developer frustration due to Windows just being Windows and failing for no reason: https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/1480#issuecomment-1053575182, https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/1480#issuecomment-1053576166

introt:
I'll redo this on top of #1978 once it gets merged

PR #1977, opened 2022-03-06
LinkMap plugin: Add more dynamic navigation

20kdc:
Adds some configuration options for more dynamic navigation of the link map.
In particular, adds a `depth` option (to control link search depth) and a `refresh_on_move` option (to allow navigating the link map dynamically).


20kdc:
So apparently, what I managed to somehow miss is that someone had already opened a PR that covers these features (#1955).
Sorry about that.

PR #1978, opened 2022-03-07
Run tests on macOS using custom dependencies

dehesselle:
This PR adds testing for macOS based on the same (JHBuild based) dependencies that I use to create our app.  
Fixes https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/1797

- Due to its very different nature I decided to create a dedicated job instead of complicating the neat matrix setup (I'm open to other opinions).

- The job is named "Python 3.x" as the Python version cannot be chosen from the outside and is unknown from the job's perspective.

- The tests need to be run on `macos-10.15`. `macos-latest` (i.e. `macos-11`) causes problems.

- I renamed the job from `build` to `test` as I assume that is a copy-paste mistake. While I usually refrain from making changes outside of my feature, I did this for uniformity as I wasn't going to name my new test job `build_macos`. 

- This PR is against the `develop` branch according to https://github.com/zim-desktop-wiki/zim-desktop-wiki/discussions/1860.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,13 +8,11 @@ on:
       - develop
 
 jobs:
-  build:
+  test:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        # disable macos until #1797 is fixed
-        #os: [macos-latest, ubuntu-latest, windows-latest]
         os: [ubuntu-latest, windows-latest]
         python-version: ["3.6", "3.10"]
 
@@ -27,11 +25,6 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           architecture: x64
-
-      - name: Install MacOS system dependencies
-        run: |
-          brew install gtk+3 pygobject3
-        if: matrix.os == 'macos-latest'
 
       - name: Install Ubuntu system dependencies
         run: |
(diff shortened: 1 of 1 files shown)


introt:
The Windows build also uses custom dependencies via MSYS2, so it'd make sense to break that one off from Ubuntu as well. Since (and correct me if I'm wrong) we don't use the system's Python with neither Windows nor macOS builds, it doesn't make much sense to test on unused versions since those are covered by the Ubuntu test; Linux is the only platform where we depend on the distributor's Python anyway.

I closed #1972 in favor of this one, will redo the Windows stuff following this one's stylistic choices. Imo it makes sense to keep the matrix for the time being, and maybe test `master`/`release` against some more and less stable distributions as well.

dehesselle:
> Since (and correct me if I'm wrong) we don't use the system's Python with neither Windows nor macOS builds, it doesn't make much sense to test on unused versions since those are covered by the Ubuntu test; Linux is the only platform where we depend on the distributor's Python anyway.

@introt Correct. As it happens I'm still using Python 3.9.x on macOS but it's not a big deal to update that to 3.10.x. Should I do that so it is more aligned with what is tested on other platforms?

introt:
We aren't doing any version-specific stuff, so testing against multiple versions is mostly there to ensure we aren't using too new (3.6+) nor deprecated features.

There might be some efficiency improvements in the newer versions, but other than that I see no point in upgrading from non-EOL versions.

PR #1982, opened 2022-03-10
Path existance is made agnostic of whether path is directory or file

Mikael-Lovqvist:
The context menu for an anchor in the page view will use `zim.fs.File` which is derived from `zim.fs.FilePath` to represent the target no matter if the target is a file or a directory. Since `file.exists()` is used to populate the context menu no `Open with...` or even `Open` will be available for anchors to local directories (for instance, another zim).
By using the `zim.fs.FilePath.exists()` function instead these options are available both for files and directories.

Another solution could be to use the `FilePath` type for all paths that has not yet been determined whether they are directories or not in the application logic.


jaap-karssenberg:
Please base your patch on the "Develop" branch instead. The "zim.fs" module
has been retired in favor of "zim.newfs".

Regards,

Jaap

PR #1986, opened 2022-03-18
Plugin for detecting and merging changes on disk

joukhaha:
Hello,
this intends to provide the functionality of an earlier pull request #178 in the form of a plugin that can be enabled and controlled through plugin settings.
Needs testing, but overall the current implementation is much simpler than previously, to the point where it might more comfortable that this gets the job done in a more reasonable way that aligns better with zim's code. Feel free to comment and provide pointers if this is not the case...

If the users of the earlier modifications are willing to give it a test and comment (I will do more tests later as well), it should be possible to iron out possible issues. According to zim manuals, the installation of the plugin should also be possible simply by copying the plugin directory (zim/plugins/diskmerger) into zim plugin directory (but didn't try this yet). If @jaap-karssenberg considers better to keep this as a custom plugin living outside zim, that should also be a possibility.

One feature of the plugin now is that if it's configured to compare the changes on disk and buffer, it will always report differences in the header. This seems to appear because zim for some reason parses the header on disk and on buffer differently in array elements. Ideally this might be fixed in zim. A major improvement with respect to the older version is also that this one should not screw up tables (as easily implemented by adapting some functions that were already built in).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@ Jaap Karssenberg <jaap.karssenberg@gmail.com>
 This branch is the Python rewrite and starts with version 0.42.
 Earlier version numbers for zim correspond to the Perl branch.
 
+##  0.74.3 - Tue 07 Dec 2021
+* Fix issue with broken links on move page
+* Fix bug in copy-paste
+* Fix bug on move page for existing textbuffer
+* Fix bug with saving of notebook editable state
+* Fix for error while loading plugins
+* Fix issue with stray unicode character
+* Fix issue to allow relative document root
+* Fix hack for rendering sourceview blocks
+* Add mnemonic keys to open sidepane tabs
+* Add links and screenshot to macOS app & update README information
+* Improve inline rendering of equations & make image DPI confingurable
+* Add box drawing characters to symbols list
+* Scroll on moving lines with linesorter plugin
+* Move floating table of contents to overlay layer
+
 ##  0.74.2 - Mon 11 Oct 2021
 * Let Toolbar plugin also remove window decoration
 * Fix execution of custom tools from Toolbar
--- a/README.md
+++ b/README.md
@@ -42,7 +42,8 @@ You will at least need the following:
 * python3 >= 3.2
 * python3-gi (also known as pygobject, but make sure to have the "gi" based version)
 * python3-xdg (optional, but recommended)
-* xdg-utils (optional, but recommended)
+* xdg-utils (optional, but recommended on linux)
+* python3-pillow (optional, to support more image formats like ".webp")
 
 To verify that zim is working properly on your system, you can run the test suite using `./test.py`. Failures do not have to be critical, but in principle, all tests should pass.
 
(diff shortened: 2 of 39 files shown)


dehesselle:
Can you change this to target the `develop` branch? The project has moved to the strategy of "master is production" and all development needs to happen in `develop`. Apologies for the inconvenience.

joukhaha:
Certainly, if/when I find straightforward means to do that... But based on a quick check it would appear that it would be easiest on the upstream repo (in the pull requests there should be a direct button to change base branch):
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request.
Or is there a better way?

introt:
If the drop-down isn't appearing for you after clicking "Edit", it'd probably be easiest if you'd close this PR and open a new one. Us underlings can't do that for others' PRs, sorry!

jaap-karssenberg:
Sounds like great functionality to have, but really need to have good test
coverage in the test suite before we can merge it. Could you add test cases
for this plugin ?

joukhaha:
(I assume the plugin succesfully transferred to development branch now.. sorry for the late notice and thanks for the tip, Dehesselle).

Figuring out how to add test coverage for the plugin is not really something that is obvious to me, unfortunately. In principle I would not mind trying to learn more about it, but can't say if/when I might have time for it. Some suggestions and pointers on how to approach the testing problem would, however, be quite useful if I get to that point...

introt:
> Figuring out how to add test coverage for the plugin is not really something that is obvious to me, unfortunately. In principle I would not mind trying to learn more about it, but can't say if/when I might have time for it. Some suggestions and pointers on how to approach the testing problem would, however, be quite useful if I get to that point...

https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/1376 can be used for reference:

> Ideally we should also have a specific test case for the plugin itself. Just run the dialog in a test and check it runs without errors would be a good start

Here's the corresponding code: https://github.com/datenwurm/zim-desktop-wiki/blob/c0874ccc4dd4ed5e854be05ffc36b05c7e40adbc/tests/command_palette.py#L4-L14 (~10 lines)

kanehekili:
I've been using Tomboy for almost 15 years, and it simply copied an "old" version and marked it so. I did not expect a "modern" application of not being able to handle working on different devices. (which implies merge conflicts)

joukhaha:
Ok, long time without anything happening, but now I added the simple test case used as an example above. Running test.py diskmerger seems to pass (no errors, at least). I also made some fixes/hacks to dealing with file permissions, as they seemed to be non-functional if different users edited the files. Largely stuff should still be considered untested, though. I guess I'll need to remake the pull request somehow to keep this on the list of possible future additions....

kanehekili:
My use case: I am a single user with more than one device. On each device  your software is installed. They "sync" with a cloud instance. My expectation: "Push" a note from device A. Later  "pull" from device B. If device B has changed that note as well I 'd expect a "merge conflict" by having two versions of the same note. Tomboy solves that with a new note that is called "xxx.old". So it's not a multi user problem, rather a multi device problem (..from my point of view). By (software) design it should make no difference if two users or two devices produce that situation.

PR #2011, opened 2022-04-17
Fix #2007 - sane Gtk-related defaults for Windows and macOS

### LINKED ISSUE #2007: Changing defaults on Windows to work around upstream Gtk window issues
**Is your feature request related to a problem? Please describe.**
[Window hotkeys & snapping don't work OOTB on Windows 11](https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/2004#issuecomment-1097707397). Switching virtual desktops on Windows 10 also affects tiling (#2004). Changing the defaults seem to also fix other issues experienced by users (see #1639).

**Describe the solution you'd like**
Changing the defaults on new Windows installs to those listed by @fmos, as suggested by @quctex:

> - Enable the Toolbar plugin
> - In the plugin configuration disable/uncheck the checkbox on "Show controls in the window decoration"
> - Make sure the environment variable GTK_CSD is not set (e.g. running set on the command prompt and looking for it).

**Describe alternatives you've considered**
_Only_ documenting the problem in the manual and on the download page. Should be mentioned in the manual in any case.

> [It wasn't clear from the tool bar plugin options that disabling the "show controls in window decoration..." would have this effect.](https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/1639#issuecomment-1000342691)

Alternatively, porting to Gtk4 :P

**Additional context**
See the linked issues.

introt:
See commit messages.

Draft - doesn't address `GTK_CSD` being 1 by default on non-portable Windows installs.

One way would be to just override it here, but that does prevent people from enjoying the beauty of the broken decorations:

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/1b9d3c14b2a2c7184a28ec5cb93fccac2f321b95/zim/__init__.py#L185-L187

### DIFF
--- a/zim/plugins/toolbar.py
+++ b/zim/plugins/toolbar.py
@@ -36,6 +36,8 @@ from zim.gui.customtools import CustomToolManager
 
 import zim.errors
 
+import os  # see issue #2007 - Gtk bug on other platforms
+DEFAULT_DECOR = os.uname().sysname == 'Linux'
 
 STYLES = (
 	('ICONS', _('Icons')), # T: toolbar style
@@ -73,7 +75,7 @@ configuring the window decoration.
 
 	plugin_preferences = (
 		# key, type, label, default
-		('show_headerbar', 'bool', _('Show controls in the window decoration') + '\n' + _('This option requires restart of the application'), True),
+		('show_headerbar', 'bool', _('Show controls in the window decoration') + '\n' + _('This option requires restart of the application'), DEFAULT_DECOR),
 			# T: option for plugin preferences
 		('show_toolbar', 'bool', _('Show toolbar'), True),
 			# T: option for plugin preferences
--- a/zim/plugins/toolbar.py
+++ b/zim/plugins/toolbar.py
@@ -36,8 +36,10 @@ from zim.gui.customtools import CustomToolManager
 
 import zim.errors
 
-import os  # see issue #2007 - Gtk bug on other platforms
-DEFAULT_DECOR = os.uname().sysname == 'Linux'
+import os  # see issue #2007 - Gtk bug on Windows and macOS
+DEFAULT_DECOR = not (os.name == 'nt' or \
+	(hasattr(os, 'uname') and os.uname().sysname == 'Darwin')
+)
 
 STYLES = (
 	('ICONS', _('Icons')), # T: toolbar style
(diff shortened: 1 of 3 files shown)


zim/plugins/toolbar.py
```diff
@@ -73,7 +75,7 @@ class ToolBarPlugin(PluginClass):
 
 	plugin_preferences = (
 		# key, type, label, default
-		('show_headerbar', 'bool', _('Show controls in the window decoration') + '\n' + _('This option requires restart of the application'), True),
+		('show_headerbar', 'bool', _('Show controls in the window decoration') + '\n' + _('This option requires restart of the application'), DEFAULT_DECOR),
```
introt, line 78: The manual states that the controls in the window decoration "is the default without the plugin", so this likely won't cut it.

zim/plugins/toolbar.py
```diff
@@ -36,6 +36,8 @@
 
 import zim.errors
 
+import os  # see issue #2007 - Gtk bug on other platforms
+DEFAULT_DECOR = os.uname().sysname == 'Linux'
```
introt, line 40: `AttributeError: module 'os' has no attribute 'uname'` on Windows. Since this isn't a one-and-done, we might as well account for the BSDs - they likely support the same stuff as Linux via X11, so the workaround needs not apply.

introt (commented):
I'd request changes, but apparently GitHub doesn't encourage self-management :)

dehesselle:
FYI broken CI on macOS is not your fault, you're just having the bad luck of catching me right in between changes (introducing multi-arch). I'll have a fix ready later today.

dehesselle:
Just completed testing, your changes have the desired effect on macOS. Nice!

dehesselle:
Jaap just merged my fix, you can rebase to unbreak macOS CI. Apologies for the inconvenience.

introt:
@dehesselle thanks for testing & fixing CI - was also good to hear Jaap's alright - did you try what happens when the plugin is disabled?

I'm wondering whether we should just document the need to keep it enabled or dig deeper to keep the decors off even without the plugin.

dehesselle:
> did you try what happens when the plugin is disabled?
Same behavior as before, the plugin needs to be enabled for the option to take effect. (This doesn't surprise me, this is "works as designed" since it was introduced.)

> I'm wondering whether we should just document the need to keep it enabled
I have documented the steps on how to enable "macOS look & feel" in the readme which is linked to from our download page (and it will become even more verbose with my next update). Future versions will already have these set by default thanks to your changes. IMHO that should be sufficient.

dehesselle:
> did you try what happens when the plugin is disabled?

Same behavior as before, the plugin needs to be enabled for the option to take effect. (This doesn't surprise me, this is "works as designed" since it was introduced.)

> I'm wondering whether we should just document the need to keep it enabled

I have documented the steps on how to enable "macOS look & feel" in the readme which is linked to from our download page (and it will become even more verbose with my next update). Future versions will already have these set by default thanks to your changes. IMHO that should be sufficient.

introt:
Thanks for your input, I think that settles the need for digging deeper into the window decoration code.

Your macOS docs are great by the way; sadly, judging by the quote in #2007, their Windows equivalents can't be held in the same regard - hopefully we can get them sorted out before the next release.

introt:
Closed by https://github.com/zim-desktop-wiki/zim-desktop-wiki/commit/1550842c2363b92b711b0489d5c4d52066c7765a

PR #2028, opened 2022-05-04
Fix KDE drag-n-drop bug

alexivkin:
On KDE page drag and drop is broken because, like #390 data absent, but unlike it, it's an empty string, not a None

### DIFF
--- a/zim/plugins/pageindex/__init__.py
+++ b/zim/plugins/pageindex/__init__.py
@@ -539,7 +539,7 @@ class PageTreeView(BrowserTreeView):
 		assert selectiondata.get_target().name() == PAGELIST_TARGET_NAME
 		data = selectiondata.get_data()
 		logger.debug('Drag data recieved: %r', data)
-		if data is None:
+		if data is None or len(data)==0:
 			data = zim.gui.clipboard._internal_selection_data # HACK issue #390
 			zim.gui.clipboard._internal_selection_data = None
 			logger.debug('Got data via workaround: %s', data)


zim/plugins/pageindex/__init__.py
```diff
@@ -539,7 +539,7 @@ def do_drag_data_received(self, dragcontext, x, y, selectiondata, info, time):
 		assert selectiondata.get_target().name() == PAGELIST_TARGET_NAME
 		data = selectiondata.get_data()
 		logger.debug('Drag data recieved: %r', data)
-		if data is None:
+		if data is None or len(data)==0:
```
dehesselle, line 542: Please adopt the coding style as it is present in the file, i.e. add spaces. See here for example: https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/26de261b13f1560e315bb6914e0f91c61ed38ac0/zim/plugins/pageindex/__init__.py#L580

alexivkin:
feel free run an autofix for the style

introt (changes requested):
Sorry, I don't think we have a linter configured for Zim's style. Additionally, PRs should be opened against the `develop` branch. I presume you made this edit from browser, so please redo the edit there: https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/85b6d81d31652042b1e9b007b211f9aab435047b/zim/plugins/pageindex/__init__.py#L557

introt:
Leaving this open for tracking purposes

PR #2037, opened 2022-05-09
Corrected variable name.

fhaftmann:
The logging message invocation in the exception handler referenced a non-existing variable and hence made exception handling itself crash.


introt (changes requested):
Good catch, thanks! Could you change the target branch to `develop`? Sorry for the inconvenience!

PR #2040, opened 2022-05-09
Corrected variable name.

fhaftmann:
The logging message invocation in the exception handler referenced a non-existing variable and hence made exception handling itself crash.

This replaces PR https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/2037.


introt (approved):
Looks good, thank you!

PR #2078, opened 2022-05-29
Update macOS version to Big Sur

dehesselle:
This accomodates necessary updates in the build infrastructure.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,12 +47,12 @@ jobs:
 
 
   test_macos:
-    runs-on: macos-10.15
+    runs-on: macos-11
     env:
       # a fixed path is required to use precompiled dependencies
       WRK_DIR: /Users/Shared/work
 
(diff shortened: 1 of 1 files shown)


dehesselle:
There are unexpected complications. This line in the code runs fine on Catalina (10.15) but crashes on Big Sur (11) as well as Monterey (12).
https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/e63526642c1c5cba7fa76a0ca3835a6a783649c5/tests/plugins.py#L147

The only fix I've found so far is raising the target platform from El Capitan (10.11) to Catalina (10.15), as going for High Sierra (10.13) still crashes. (I'm not looking at Mojave as every Mac that can run Mojave can run Catalina as well.)

As this problem appears to only affect the test suite at this time, I'm closing the PR. We'll have to revisit this topic once GitHub retires Catalina runners in CI.

PR #2094, opened 2022-06-28
Support closing an open 'find bar' by pressing ESC

roland-ruedenauer:
I found no way to close the "find bar" opened by a search operation using the keyboard, so I added a small keypress handler a while ago that allows one to close the "find bar" via pressing the ESC key.
So far I've only tested this on Linux.

### DIFF
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -6069,6 +6069,15 @@ class PageView(GSignalEmitterMixin, Gtk.VBox):
 		self._insertedobject_manager = InsertedObjectPageviewManager(self)
 		self.__zim_extension_objects__.append(self._insertedobject_manager) # HACK to make actions discoverable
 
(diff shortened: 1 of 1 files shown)


introt (changes requested):
Hi Roland, nice to see you again - it's been a while!

This functionality already exists, though it does require focusing the find bar (by eg. pressing ctrl+f).

Regardless of whether that's good for UX, there's no need to have duplicate handlers, and the remaining one should return the appropriate value (eg. `True` if the event has been handled).

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/5e82f2d6fd5d2fc1b45cbd396b1fdac0f6465f4c/zim/gui/pageview/__init__.py#L8438-L8444

roland-ruedenauer:
Hi!
yes, one handler is enough. I removed the exising one from class FindBar.

introt (approved):
Looks good to me, thanks!

Issue #2077, opened 2022-05-28
Windows Installer doesn't set DisplayVersion


ajkajk:
The Windows installer sets up fields in `[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\404fbece-3a0a-4f4f-b1f1-82ce46af9696]` e.g. `"DisplayName"="Zim Desktop Wiki"`. However it doesn't currently set up a DisplayVersion string e.g. `"DisplayVersion"="0.74.3"`.

As a result the windows package manager (winget) can't determine the currently installed version.  This means it attempts to update zim as the version available is always more promising than the "unknown" version. The effort in making zim available in the "winget" package source effectively does not support detecting if a new version of zim is available for update.  

Would it be possible to change the installer to set up a `DisplayVersion` string and support use of winget to update zim?

introt:
Thanks for the report!

`grep` returns

    windows/src/zim-installer.nsi:        WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\${UNINSTKEY}" "DisplayName" "${APPNAME}"

so the fix is likely just adding another `WriteRegStr` - could you confirm this to work and open a PR? I don't have a Windows box handy.

ajkajk:
Thanks for your response.

I'll give it a go.  However, I have no experience in any of this, so it will likely take some time to get up to speed....

introt:
You can find the steps here: https://github.com/zim-desktop-wiki/zim-desktop-wiki/tree/master/windows

Instead of `git clone https://github.com/zim-desktop-wiki/zim-desktop-wiki.git`, run `git clone https://github.com/introt/zim-desktop-wiki.git --branch displayversion` to fetch my patched version with the fix.

Thanks for taking initiative!

ajkajk:
I've tried building the installer.  `build.sh` fails for me at line 168 `pip install PyGObject pyinstaller==3.6 # Fix pyinstaller version, latest has bug` i.e. nothing to do with the extra line in `zim-installer.nsi` 
I can run that `pip install` from the command line OK. I tried playing with the line in the script, including leaving it out altogether, but never got the environment set-up to create the installer.  
The error reported is `error: --plat-name must be one of ('win32', 'win-amd64', 'win-arm32', 'win-arm64')` I don't know what value is passed, or where it comes from.
Sorry.
[...]

PR #2097, opened 2022-07-04
Windows deployment changes

fmos:
Fixes the deployment which failed due to upstream changes, and adds DisplayVersion to support updates via winget (see #2077).


introt:
Looks good to me, thanks @fmos! I'm open to merging straight to `master` as this concerns packaging, not the applications itself.

@jaap-karssenberg LGTM, as the cool kids say :)

Issue #1419, opened 2021-03-12
There should be a distinction between em dash and en dash


eyolf:
Currently, if two hyphens are entered surrounded by spaces, this i converted to an em dash. However in the dokuwiki syntax to which it is referred in the wiki, -- produces en en dash, and --- gives an em dash. There are good reasons why this distinction should be made. One is that the en dash is preferred in many typographic traditions, but the main reason is that the en dash should be used to indicate ranges (2–3, London–Paris). In these situations an em dash will always be wrong.
The next question, then, is if the spaces around are needed, but I guess they are, since multiple dashes can be used to make a hr.

nelas:
Hi @eyolf, I agree the default behavior should be  `--` for en dash and `---` for em dash. Perhaps there's a historical reason why it's not, but @jaap-karssenberg can weight in about that.

Otherwise, it's simple to change the default behavior. You probably have the _Insert Symbol_ plugin activated. If you go to `Insert > Symbol... > Edit`, it will open the `simbols.list` config file on your text editor. At the bottom of the file there is a section called `Additional typography` with the following line:

```
--	8212  # em dash aka &mdash; in HTML
```

Change it to three hyphens, and insert a line for handling the en dash like that:

```
--	8211  # en dash aka &ndash; in HTML
---	8212  # em dash aka &mdash; in HTML
```

Save and close the file, and then restart Zim. You should now be able to insert en or em dashes using two or three hyphens as expected.

Also consider that you can insert dashes (and other symbols) by typing `\ndash` or `\mdash`. And that if you type `\ndash;` (note the semicolon), you'll generate an en dash _without_ the trailing whitespace.

Hope this helps!
Cheers

jaap-karssenberg:
No particular historic choice on my side - just copied the default list
from the HTML::Entities perl module, feel free to propose your change as a
merge request and I'll pick it up.

Regards,

Jaap

PR #2107, opened 2022-07-17
Enables distinct shortcuts for en and em dashes

nelas:
Currently, the `Insert Symbol` plugin autoformats `--` to `—` (em dash). This is counter-intuitive since `--` usually represents `–` (en dashes) and `---` represents `—` (em dashes) in other markups such as LaTeX.

This pull request changes the default behavior as discussed in #1419.

### DIFF
--- a/data/symbols.list
+++ b/data/symbols.list
@@ -337,7 +337,8 @@
 <=>	8660
 <==>	8660
 <==	8656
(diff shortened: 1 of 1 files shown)


introt (approved):
Thanks for clearing this out from the backlog!

PR #2108, opened 2022-07-17
Fixes typo and adds example to Template manual

nelas:
Added an example showing how to reference a template resource using the `resource(filename)` function. It wasn't clear to me that the filename should be within quotes. Also fixed a typo and some formatting.

### DIFF
--- a/data/manual/Help/Templates.txt
+++ b/data/manual/Help/Templates.txt
@@ -323,7 +323,7 @@ The following options are supported:
 
 **HTML**
 * ''options.line_breaks'': value should be either "''default''" or "''remove''", controls line breaks (''<br>'' elements) in paragraphs
-* ''options.empty_lines'': value should be either "defualt" or "remove", controls whether empty lines in the source should be inserted as ''<br>'' elements in the HTML output between top level elements like paragraphs, headings etc.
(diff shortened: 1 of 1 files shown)


introt (commented):
While I approve the addition (it's great, thanks!), I'd like to see the existing weird quotation fixed while we're at it.

Given the above example (`[% options.line_breaks = 'remove' %]`), I think all the quotes around formatted text should instead be inside the formatted text, eg. `''"default"''` instead of `"''default''"` - I'd wager that double quotes are used here due to `'''` being wiki markup.

nelas:
Hi @introt, yes, in this case, it works with quotes inside because the example `'remove'` is just above. But for most cases throughout the manual, I think it'd be cleaner to remove the quotes. The verbatim is already doing the job of highlighting the word, adding quotes around is redundant. Changing from "`verbatim`" to `verbatim` is also easier to read. I can revise this in another PR if it makes sense.

introt:
Agreed - just didn't look around enough before commenting, my bad. The combination of verbatim + quotes to emphasize quoted text is indeed redundant, and shouldn't be used.

The special case for verbatim + quotes is when those quotes should be included verbatim (pun not intended), as is the case with the options: the quotes denote that the options are literally [string literals](https://en.wikipedia.org/wiki/String_literal#Bracketed_delimiters) (pun intended).

PR #2109, opened 2022-07-17
Standardize menu items to title case

nelas:
Most menu items are written in Title Case, except for a few. I changed these to Title Case for consistency. Example: `View debug log` to `View Debug Log`.


introt (approved):
Thanks!

Issue #2089, opened 2022-06-21
Reindex error across Master and Develop branches


releve1403:
When switching from "master branch" to "develop branch", I noticed two entries marked "conflicted" in the journal. 

Tried to erase them within Zim but it didn't work; pages were still ther. 

Just erased the files directly in file manager. Then reindexing gave this crash error below.

Maybe it isn't really a bug, as the messages appears to be related to tasklist. And, as there is many differences between Master and Develop about tasklist, this is just kind of normal... Don't know.

This is zim 0.74.3
Platform: posix
Locale: fr_CA UTF-8
FS encoding: utf-8
Python: (3, 8, 10, 'final', 0)
PyGObject: (3, 36, 0)

**Traceback**
```
  File "/usr/lib/python3/dist-packages/zim/actions.py", line 159, in _on_activate
	self.call()
  File "/usr/lib/python3/dist-packages/zim/actions.py", line 130, in call
	return self._action.func(self._instance, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/zim/plugins/tasklist/init.py", line 172, in show_task_list
	self._task_list_window = self._show_task_window(selection_state=None)
  File "/usr/lib/python3/dist-packages/zim/plugins/tasklist/init.py", line 181, in _show_task_window
	window = TaskListWindow(notebook, index, navigation, properties, self.plugin.preferences['show_inbox_next'])
  File "/usr/lib/python3/dist-packages/zim/plugins/init.py", line 111, in _init_wrapper
	orig_init(self, *arg, **kwarg)
  File "/usr/lib/python3/dist-packages/zim/plugins/tasklist/gui.py", line 338, in init
	swindow = self._create_tasklisttreeview(navigation, properties)
  File "/usr/lib/python3/dist-packages/zim/plugins/tasklist/gui.py", line 96, in _create_tasklisttreeview
	self.tasklisttreeview = TaskListTreeView(
  File "/usr/lib/python3/dist-packages/zim/plugins/tasklist/gui.py", line 801, in init
	self.refresh()
  File "/usr/lib/python3/dist-packages/zim/plugins/tasklist/gui.py", line 851, in refresh
	self._append_tasks(self.taskselection.list_tasks(), None)
  File "/usr/lib/python3/dist-packages/zim/plugins/tasklist/gui.py", line 863, in _append_tasks
	for prio_sort_int, row in enumerate(task_iter):
[...]

Issue #1584, opened 2021-09-16
Option to remove attachments that are no longer linked


sojusnik:
As discussed [here](https://bugs.launchpad.net/zim/+bug/416859), it still would be very useful to have a way of clearing all unlinked attachments, like images, automatically or in one batch, f.i. when using "check & update index".

introt:
Anki has a similar option for deleting unused media files. With Zim it's a bit trickier due to indirect links from other pages, but definitely doable. I'd also love an option for the reverse: to remove broken links/objects that point to attachments that don't exist.

jaap-karssenberg:
Pre-requisite is to add logic in zim to add file links to the index. That
way it becomes straight forward to track links / cleanup attachments / fix
links.

I have a partial patch lying around, but needs further fixes to make it
work.

introt:
Would it make sense to also store the attachment dimensions somewhere, including when they haven't been manually set? That'd allow us to calculate the correct view position when opening an anchor link, as currently pages with lots of images tend to just throw you wherever close by.

introt:
> Pre-requisite is to add logic in zim to add file links to the index. That
> way it becomes straight forward to track links / cleanup attachments / **fix**
> **links.**

Related: #48

pityhero233:
That function would definitely be needed, but may I inquire if it's implemented as automatic or toggleable in the future patch you mentioned? Because I find myself pasting images into Zim, then removing the image itself so I don't have to save my screenshots elsewhere on disk while also being able to save screenshots into Zim attachment ...

introt:
Definitely won't delete stuff without asking by default, that much is sure.

Personally I'd go for batch over fully automatic, with a confirmation dialog similar to the one for page deletion.

![image](https://user-images.githubusercontent.com/26069787/163345347-c9fdcb0b-dac0-44cc-8bfc-2036f32ae4d3.png)

oxidworks:
I've been a zim user for many years and this has always bothered me — orphaned attachments piling up over time with no easy way to clean them up. So I finally wrote a plugin for it.

It scans all pages for image/file references (including cross-page and imagegenerator objects), then lists attachment files that aren't referenced anywhere. Selected orphans can be moved to the system trash.

Tested on my own notebook (~3000 pages) — found 55 orphans totaling 7.75 MB.

Submitted as PR: #2943

Thanks to all the developers for this great tool!

PR #2112, opened 2022-07-21
Bettering the handling of missing images re: #2102

### LINKED ISSUE #2102: Error messages not helpful when Diagram image file missing
**Describe the bug**
If you have a Zim page with Graphviz/DOT code where the generated image file was deleted outside of Zim, you get odd and unhelpful error messages about that.

**To Reproduce**
1. Enable the "Diagram" plugin and make sure it's happy.
2. Use "Insert > Diagram" to create a diagram (sample code below).
3. Verify everything works in Zim.
4. Find the diagram file outside of Zim (`find /path/to/Zim/ -iname '*.png'`) and delete it.
5. Refresh the page in Zim.
6. The diagram will disappear.
7. Edit the Graphviz/DOT code but do not make changes, and hit OK.
8. Get a dialog that says:
```
An error occurred while generating the image.
Do you want to save the source text anyway?
[No][Yes]
```

Also, once that happens the included image in the Zim page gets very small (since it doesn't exist), so it's hard to click on it to hit "Edit" to get back into the Graphviz/DOT code editor.

**Work around:**
1. Click "Edit" on the image to get into the Graphviz/DOT code editor.
2. Make a change in the Graphviz/DOT code code, e.g., hit enter and then backspace.
3. Hit OK and it will now work.

Sample Graphviz/DOT code you can play with
```graphviz
digraph {
    deploy     [label="Deploy to production?", shape=box]
    is_friday  [label="Is it \n read-only \n Friday?", shape=diamond]
    no_deploy  [label="Do not deploy!",        shape=box]
    ok_deploy  [label="OK to deploy!",         shape=box]
    tgif       [label="Enjoy your weekend!",   shape=box, style=rounded]

    deploy    -> is_friday
    is_friday -> no_deploy  [label="yes"]
    is_friday -> ok_deploy  [label="no"]
    no_deploy -> tgif
    ok_deploy -> tgif
}
```

**Expected behavior**
1. Either Zim should provide a better error message,
2. Or, BETTER, it should notice a missing image file and just re-create it, ideally at index or page-open time, but at least at edit-code time.

- zim-0.74.3-t202205241835~ubuntu20.04.1 from PPA
- OS: Mint-20.3
- graphviz-2.42.2-3build2
- xdot-1.1-2

**Additional context**
This happens because of #793 & #1584.  Until I started using "Diagram" I had zero image files in my Zim dir, so when I accidentally paste an image (at least once every couple of weeks), I just run `find /path/to/Zim/ -iname '*.png' -delete` to nuke all images.  Obviosly that nukes the Diagram image too (might be better to output SVG instead of PNG anyway?).  No big deal, I thought, Zim or I will just re-create it when needed.

Except...there is no clue what is wrong, so other folks may be lost if this happens.  Even though I caused the root problem and knew what it was, I found the errors confusing and it took me some thinking to find the (trivially simple) workaround.

When run `zim --debug` errors are not helpful:
```
DEBUG: BackgroundCheck started
DEBUG: BackgroundCheck finished
DEBUG: Opening dialog "Edit Diagram"
DEBUG: Dialog response OK
DEBUG: Running QuestionDialog
DEBUG: Q: An error occurred while generating the image.
Do you want to save the source text anyway?
DEBUG: A: True
DEBUG: Closed dialog "Edit Diagram"
INFO: Remove file: /tmp/zim-0xlmyvcm/diagram.dot
INFO: Remove file: /tmp/zim-0xlmyvcm/diagram.png

(zim:239576): Gtk-WARNING **: 15:51:31.974: gtk_widget_size_allocate(): attempt to allocate widget with width -2 and height -2

(zim:239576): Gtk-WARNING **: 15:51:31.974: Negative content width -1 (allocation 1, extents 1x1) while allocating gadget (node box, owner GtkVBox)

(zim:239576): Gtk-WARNING **: 15:51:31.974: Negative content height -1 (allocation 1, extents 1x1) while allocating gadget (node box, owner GtkVBox)

(zim:239576): Gtk-WARNING **: 15:51:31.974: gtk_widget_size_allocate(): attempt to allocate widget with width -1 and height 0
```

introt:
Regenerates missing images on page open/reload and when opening or accepting the `ImageGeneratorDialog`. Outdated images are also regenerated on page open/reload (ie. on the initialization of `BackwardImageGeneratorModel`) based on `LocalFile.mtime()`, allowing tighter workflows using external editors (eg. `vimdot` for diagrams, or any Lilypond-compatible editor for sheet music etc).

Due to the existing implementation writing the script file **after** generating the image, `LocalFile._set_mtime()` is called and thus
depended upon.

Neither missing nor outdated images are regenerated on export.

Partially fixes #2102.

Also includes a nicer fallback for missing images (should both the source file and the render go missing):

image

### DIFF
--- a/zim/plugins/base/imagegenerator.py
+++ b/zim/plugins/base/imagegenerator.py
@@ -183,6 +183,18 @@ class BackwardImageGeneratorModel(ImageGeneratorModelBase):
 			self.image_file = _stitch_fileextension(self.script_file, imagefile_extension)
 			self.attrib['src'] = './' + self.image_file.basename
 
+		if self.script_file.exists() and (
+				not self.image_file.exists() or
+				self.script_file.mtime() > self.image_file.mtime()):
+			logger.debug('Image did not exist or source file was modified externally, re-generating')
+			try:
+				text = self.get_text()
+				image_file, log_file = self.generator.generate_image(text)
+			except Error:
+				pass
+			else:
+				self.set_from_generator(text, image_file)
+
 	def get_text(self):
 		if self.image_file is not None and self.script_file.exists():
 			text = self.script_file.read()
@@ -192,8 +204,10 @@ class BackwardImageGeneratorModel(ImageGeneratorModelBase):
 		return self.generator.filter_source(text)
 
 	def set_from_generator(self, text, image_file):
-		image_file = adapt_from_oldfs(image_file)
 		self.script_file.write(text)
+		image_file = adapt_from_oldfs(image_file)
+		image_file._set_mtime(self.script_file.mtime())  # avoid needless regen
+
 		if image_file == self.image_file:
 			pass
 		else:
@@ -386,7 +400,7 @@ class ImageGeneratorDialog(Dialog):
 			model.set_from_generator(text, image_file)
 		model.generator.cleanup()
 
-	def __init__(self, widget, label, generator, image_file=None, text='', syntax=None):
+	def __init__(self, widget, label, generator, image_file:'LocalFile'=None, text='', syntax=None):
 		title = _('Edit %s') % label # T: dialog title, %s is the object name like "Equation"
 		Dialog.__init__(self, widget, title, defaultwindowsize=(450, 300))
 		self.generator = generator
@@ -428,8 +442,16 @@ class ImageGeneratorDialog(Dialog):
 		hbox.pack_start(self.logbutton, False, True, 0)
 
 		self.set_text(text)
-		self.imageview.set_file(self.image_file) # if None sets broken image
+		if self._file_deleted:
+			self.update_image()  # includes setting imageview file
+		else:
+			self.imageview.set_file(self.image_file) # if None sets broken image
 		self.textview.grab_focus()
+
+	@property
+	def _file_deleted(self):  # see #2102
+		" Returns True when image_file is set but doesn't exist, ie. when the file has been deleted. "
+		return self.image_file is not None and not self.image_file.exists()
 
 	def set_text(self, text):
 		buffer = self.textview.get_buffer()
@@ -461,7 +483,8 @@ class ImageGeneratorDialog(Dialog):
 
 	def do_response_ok(self):
 		buffer = self.textview.get_buffer()
-		if buffer.get_modified():
+		if buffer.get_modified() or self._file_deleted:
+			logger.debug('Image modified or did not exist, re-generating')
 			self.update_image()
 
 		if not (self.image_file and self.image_file.exists()):
--- a/zim/plugins/base/imagegenerator.py
+++ b/zim/plugins/base/imagegenerator.py
@@ -183,9 +183,10 @@ class BackwardImageGeneratorModel(ImageGeneratorModelBase):
 			self.image_file = _stitch_fileextension(self.script_file, imagefile_extension)
 			self.attrib['src'] = './' + self.image_file.basename
 
-		if self.script_file.exists() and (
+		# Regen missing/outdated image if the notebook/page isn't readonly
+		if not (notebook.readonly or (page and page.readonly)) and self.script_file.exists() and (
 				not self.image_file.exists() or
-				self.script_file.mtime() > self.image_file.mtime()):
+				self.script_file.mtime() + 1 > self.image_file.mtime()):
 			logger.debug('Image did not exist or source file was modified externally, re-generating')
 			try:
 				text = self.get_text()
@@ -204,6 +205,7 @@ class BackwardImageGeneratorModel(ImageGeneratorModelBase):
 		return self.generator.filter_source(text)
 
 	def set_from_generator(self, text, image_file):
+		# FIXME: refactor the file saving sequence (save script first, generate image second); see #2112
 		self.script_file.write(text)
 		image_file = adapt_from_oldfs(image_file)
 		image_file._set_mtime(self.script_file.mtime())  # avoid needless regen
@@ -483,7 +485,7 @@ class ImageGeneratorDialog(Dialog):
 
 	def do_response_ok(self):
 		buffer = self.textview.get_buffer()
-		if buffer.get_modified() or self._file_deleted:
+		if buffer.get_modified() or self._file_deleted:  # XXX: doesn't check notebook editable
 			logger.debug('Image modified or did not exist, re-generating')
 			self.update_image()
 
--- a/zim/gui/insertedobjects.py
+++ b/zim/gui/insertedobjects.py
@@ -220,7 +220,7 @@ class ImageFileWidget(InsertedObjectWidget):
 		if file.exists():
 			self.image = Gtk.Image.new_from_file(file.path)
 		else:
-			self.image = Gtk.Image()
+			self.image = Gtk.Image.new_from_stock(Gtk.STOCK_MISSING_IMAGE, Gtk.IconSize.DIALOG)
 		self.image.set_property('margin', 1) # seperate line and content
 		self.add(self.image)
 
@@ -231,10 +231,10 @@ class ImageFileWidget(InsertedObjectWidget):
 
 	def set_file(self, file):
 		self.file = file
-		if self.file.exists():
+		if self.file is not None and self.file.exists():
 			self.image.set_from_file(file.path)
 		else:
-			self.image.clear()
+			self.image.set_from_stock(Gtk.STOCK_MISSING_IMAGE, Gtk.IconSize.DIALOG)
 
 
 def _find_plugin(name):


introt:
sidenote: ability to edit inserted objects' source files outside zim & having the images regenerate on page reload could really elevate some workflows

introt:
@jaap-karssenberg I'd like to have some feedback from you on the following (no hurry!):

* `mtime`
  * Can it be depended on? Availability, modifiability etc.
  * Would refactoring to write the source file before generating the image be possible or would it break API too much? 
  * Should checksums be used instead, or as a fallback?
  * Should the filesystem watcher be used instead (would require keeping the page open)?
* Any tips on regenerating images on export? I'd guess getting `BackwardImageGeneratorObjectType.format()` to initialize the `BackwardImageGeneratorModel` would now suffice, but seems to require a bunch of API changes to get everything exposed - and I'm not sure whether that's the best way to handle this anyway.

zim/plugins/base/imagegenerator.py
```diff
@@ -183,6 +183,18 @@ def check_image_file(new_script_file):
 			self.image_file = _stitch_fileextension(self.script_file, imagefile_extension)
 			self.attrib['src'] = './' + self.image_file.basename
 
+		if self.script_file.exists() and (
+				not self.image_file.exists() or
+				self.script_file.mtime() > self.image_file.mtime()):
```
introt, line 188: Might want to consider `+ 1` in case of racey behavior on some filesystems: https://stackoverflow.com/questions/8246933/detect-mtime-resolution-of-filesystem

vossenjp:
If you can tell me when this hits the PPA (`deb http://ppa.launchpad.net/jaap.karssenberg/zim/ubuntu focal main`) or let me know how to check, I'll update and test my use case from #2102.  I'm running `zim-0.74.3-t202207071804~ubuntu20.04.1` on Mint-20.3 right now.

introt:
@vossenjp If you want to test right away, you can have a go using these command lines:

```
git clone git@github.com:introt/zim-desktop-wiki.git --branch=missing --depth=1
cd zim-desktop-wiki
./zim.py --index --flush $NOTEBOOK_NAME
./zim.py -D $NOTEBOOK_NAME
```

No need to compile or install anything - if you've got Zim installed via apt, you'll have the dependencies as well.

vossenjp:
Got to love Python!  I did that, and it works.  I do not see any changed icon for the missing diagram, but that might be expected due to icon locations and such.

Summary: as I see it, the things that now fix the broken image are:
1. Re-launching the test-Zim
2. Switching to the affected page and hitting CTRL-R

In other words, once Zim is running, if the diagram is deleted then you switch to the affected page, it is not fixed until you hit CTRL-R.  That still does "affect" me, because my Zim instance is long-running and I do that cleanup while it's running.  But butter logging and handling on launch is great, and CTRL-R is a trivial fix for something that I am doing that is arguably a Bad Idea anyway.  :-)

First steps I tried:
1. Be on on another page in old/PPA `zim-0.74.3-t202207071804~ubuntu20.04.1`
2. From the CLI, delete `/path/to/diagram001.png`
3. Switch to affected page = broken as per #2102
4. Close Zim
5. `git clone git@github.com:introt/zim-desktop-wiki.git --branch=missing --depth=1`
6. `cd zim-desktop-wiki`
7. `./zim.py --index --flush $NOTEBOOK_NAME` (Lots of output)
8. `./zim.py -D $NOTEBOOK_NAME`
	* Got: `DEBUG: Image did not exist or source file was modified externally, re-generating`
9. It opened to the affected page and the diagram was back!
10. Switch to another page
11. From the CLI, delete `/path/to/diagram001.png`
12. Switch to affected page = diagram is missing; I see a small hollow square containing a prohibited sign
13. CTRL-R to refresh page = diagram was back!

I was assuming the CTRL-R was needed due to having just been on that page previously, caching, etc.  But a little more switches pages around didn't change that, when I went to the affected page after deleting the file, only CTRL-F fixed it.

Then I tried:
1. Be on on other page
2. From the CLI, delete `/path/to/diagram001.png`
3. Switch to affected page = broken
4. Switch to another page
5. From the CLI `./zim.py --index --flush $NOTEBOOK_NAME`
6. Switch to affected page = diagram is missing (the usual previous debug errors about "Negative content height" and such)
7. CTRL-R to refresh page = diagram was back (and got `re-generating)!

That I didn't expect.  I'd expected the `--index --flush` to fix it, but it doesn't.  Re-tested that using:
1. From the CLI, `ls -l /path/to/diagram001.png`
2. `rm -v /path/to/diagram001.png`
3. `./zim.py --index --flush $NOTEBOOK_NAME`
4. `ls -l /path/to/diagram001.png` = not found

That's great progress, and it is probably the correct balance between page-load-time and fix-stupid-stuff-people-MIGHT-do. ;-)

introt:
Sounds like everything's working as it should (so far) - lemme clear a few things up.

> Also, once that happens the included image in the Zim page gets very small (since it doesn't exist), so it's hard to click on it to hit "Edit" to get back into the Graphviz/DOT code editor.

> I do not see any changed icon for the missing diagram

It's the "small hollow square containing a prohibited sign" - used to be just the square.

> once Zim is running, if the diagram is deleted then you switch to the affected page, it is not fixed until you hit CTRL-R

image

Yeah, that's a caching thing; you can compare the behavior to the parsing of manually written wiki formatting - just doesn't seem to want to happen without a relaunch or a reload. Though since the image object is initialized when a page is opened for the first time _after_ launch[^ie] and not at index time, you can work around it by restarting Zim after the mass deletion of all pngs[^sidenote] and never see a missing or outdated diagram (as long as their source file exists).

> I'd expected the --index --flush to fix it, but it doesn't.

Sorry for misleading you there - the flushing was mean to prevent indexing errors unrelated to this issue (see #2089)

However,

```
INFO: Open page: Foo (Foo)
DEBUG: BackgroundCheck started
DEBUG: BackgroundCheck found out-of-date
DEBUG: Index folder: /home/introt/Notebooks/Demo/Foo
DEBUG: Drop file: Foo/diagram.png
```

since information about the existence of the diagram png is saved to the index, we could force an automatic reload of reopened pages if any of their attachments have been deleted since indexing.

[^ie]: ie. unopened pages are _indexed_, but not _parsed_ - image objects aren't initialized before a page is opened
[^sidenote]: Sidenote: If you're only targeting accidental pastes, `-name pasted_image*.png` would save you the hassle

vossenjp:
Sigh.  `-name pasted_image*.png`...  NOW you tell me!?!  :-) That is so blinding obvious that I totally failed to see it, and, yes, of course that would fix my self-inflicted "problem."  I've made a note for the future.

So it sounds like so far, so good.  And personally, as above, the better logging and CTRL-R seem to be a good balance.

> since information about the existence of the diagram png is saved to the index, we could force an automatic reload of reopened pages if any of their attachments have been deleted since indexing.

I do not have a strong opinion there.  That seems to be "more automatic fixing" at a small perceived page load performance cost.  I tend to run older, constrained hardware, so the cost is more noticeable for me, but not a big deal either way.  Since this is an edge case anyway...I'd vote for KISS (keep it simple, stupid).  Or at least as simple as possible.

jaap-karssenberg:
@introt 

- Yes, I believe it is relatively safe to compare mtime between the script source file (not the page source) and the generated image. The base assumption is anyway that we are always allowed to overwrite the image file. If that needs refactoring to write the script file first, that is perfectly fine.
- For the "backward" version mtime should suffice. ( I've also implemented an "non-backward" version which does a checksum based store, like thumbnails, but this isn't used yet by any plugin. The goal there is to store the images as a cache and allow re-use in multiple pages. )
- Re-generating on export should not be special. I expect you can just trigger re-generating all images just before exporting the page based on the same mtime check and all should be fine. Or is there a more subtle issue here ?

introt:
Thanks for the reply @jaap-karssenberg 

> The base assumption is anyway that we are always allowed to overwrite the image file

To confirm: no need to check `notebook.editable`? Failing silently on readonly filesystems is probably fine, as the generation is tried only once par manual reload.

> Re-generating on export should not be special. I expect you can just trigger re-generating all images just before exporting the page based on the same mtime check and all should be fine. Or is there a more subtle issue here ?

The issue here is that the current check mirrors the one in the "non-backward" version and thus lives inside the `__init__` function of `BackwardImageGeneratorModel`, which is not initialized on export (regardless of whether the export is triggered from the GUI or using the CLI). Is the solution moving it somewhere else, and where might that be?

The bigger picture comprises the user expectations around exports: if stuff like diagram image files may be updated upon opening a page, should the export match the source files as they appear on the file system, or as the page appears when opened in Zim?

---

Steps to reproduce this issue using the code in this PR:

1. Create a page "Foo" in the notebook "Demo"
2. Insert a diagram into the page "Foo"
3. Close Zim
4. Delete the `diagram.png` file from the attachment folder
5. Export the page without opening it in the GUI: `./zim.py --export --overwrite -o /tmp/export Demo Foo`
6. List the contents of `/tmp/export_files` and the attachment directory

Expected behavior:

`diagram.png` is regenerated, and can be found from both directories

Actual behavior:

`diagram.png` is not regenerated without opening the page in Zim

jaap-karssenberg:
Hmm, you are right - I hadn't given thought to readonly notebooks. Indeed
should respect that. I think it is fair to just not generate new images
when a notebook is read-only. After all, if it is broken and then made
read-only it is fair to let it stay broken.

Same consideration on export. If it is broken, maybe we are overdoing it by
trying to fix on the fly when exporting. Let's limit scope to re-gerenating
images on open-page. And/or add a tool to scan the whole notebook for
broken images and re-generate them.

The idea with the "non-backward" version is that we can improve these
aspects and make it more seamless. Here we allow the script source to be
inline in the page (so no separate file needed) and always generate it ad
hoc when needed. By using cache folder also no concern for read-only.

In terms of API, the real problem on export is that the inline object needs
to inform the exporter that there are additional assets - e.g. a cache file
that needs to be copied - this can be resolved by adding a "list_assets()"
method. In the ImageGeneratorObjectType base class we could add a
"get_image_cache_file()" method that uses the generator and produces a file
name that can be used on both "list_assets()" and "format_...()" methods.

Would that work?  Otherwise I just need to start coding to fully explore
the territory myself. I will be offline for most of the next month, so may
take a while.

--- Jaap

introt:
Thanks for the quick response @jaap-karssenberg,

I agree with limiting the scope of this PR. I'll add the check for whether the notebook is readonly - didn't remember to consider it myself before your first reply - and request a code review.

`list_assets()` sounds like a good addition to the API, and something that the aforementioned scan tool would require. Such a tool probably ties heavily into #1584, and is definitely outside the scope of this PR.

In my opinion both the tool and the proposed API addition belong to the `next` milestone at earliest, not `0.75`. I will add a `# FIXME` as a reminder of the need to refactor the file saving sequence, as implementing the new API will probably lead to changes in the process anyway.

Issue #1820, opened 2021-12-16
Roll/Unroll parent page bug in the treeview 


tehkonst:
**Describe the bug**
When I click a small triangle near a parent page I can't unroll it back

![unroll_bug](https://user-images.githubusercontent.com/2618162/146309166-035250db-62f9-40f8-b399-49cb910a945b.gif)

1. Choose any note on the second level
2. Click the triangle near the parent entry
3. Then click again

And another one:
![unroll_bug2](https://user-images.githubusercontent.com/2618162/146310961-1ffd180f-14bd-4162-8685-61fc46cb0a56.gif)
1. Unroll any non-current branch
2. The click on it and make it active
3. Then click the triangle several times

**Expected behavior**
The treeview elements have to be rolled up and unrolled as expected.

**Environment (please complete the following information):**
 - Version 0.74.3 -- ppa
 - xUbuntu 20.04

introt:
Confirmed - after opening a page and closing its branch, it can no longer be opened. Nice catch!

introt:
Keybindings don't stop working, and after expanding with one, the branch becomes collapsible once more- yet can't be expanded again using the button.

nelas:
I also ran into this issue when `Edit > Preferences > Plugins > Page Index > Configure` is like that:

```
[x] Automatically expand sections on open page
[ ] Automatically collapse sections on close page
```

If the automatic collapse option is also activated, clicking on the triangle of the current page neither expands nor collapses the tree (it stays expanded).

rscmbbng:
Can confirm issue as well!

`$ zim --version
zim 0.74.2
`

mke21:
Zim version 0.74.3 on windows has this as well. Arrow keys still work though.

prahal:
0.74.3 on Debian amd64. Reproducible with both gtk3 3.24.24-4+deb11u2 and 3.24.34-1.
Note that here this is only reproducible on a treeview item that is selected (current page in the index).
Switching to a subpage or a sibling let me fold/unfold it anew.

sojusnik:
[...]

PR #2113, opened 2022-07-22
Fix roll/unroll issue on currently selected page in treeview.

mrffr:
Fixes a bug where some users could no longer roll/unroll a currently selected page.
This was due to autoexpand/autocollapse being called on the currently selected page whenever the roll/unroll toggle was clicked.

#1820 - The "And another one" example.

### DIFF
--- a/zim/plugins/pageindex/__init__.py
+++ b/zim/plugins/pageindex/__init__.py
@@ -400,6 +400,7 @@ class PageTreeView(BrowserTreeView):
 		self._autoexpanded = None
 		self._autoexpand = True
 		self._autocollapse = True
+		self.last_selected_path = None
 
 		column = Gtk.TreeViewColumn('_pages_')
 		column.set_expand(True)
@@ -473,6 +474,10 @@ class PageTreeView(BrowserTreeView):
 		model = self.get_model()
 		treeiter = model.get_iter(treepath)
 		mytreeiter = model.get_user_data(treeiter)
+
+		if self.last_selected_path == model.get_indexpath(treeiter):
+			return
+		self.last_selected_path = model.get_indexpath(treeiter)
 
 		if self._autocollapse:
 			self.restore_autoexpanded_path()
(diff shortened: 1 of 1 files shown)


zim/plugins/pageindex/__init__.py
```diff
@@ -400,6 +400,7 @@ def __init__(self, notebook, navigation, model=None):
 		self._autoexpanded = None
 		self._autoexpand = True
 		self._autocollapse = True
+		self.last_selected_path = None
```
introt, line 403: Prepend the variable name with an underscore (to denote private API)

zim/plugins/pageindex/__init__.py
```diff
@@ -474,6 +475,10 @@ def do_row_activated(self, treepath, column):
 		treeiter = model.get_iter(treepath)
 		mytreeiter = model.get_user_data(treeiter)
 
+		if self.last_selected_path == model.get_indexpath(treeiter):
+			return
+		self.last_selected_path = model.get_indexpath(treeiter)
```
introt, line 480: Use a local variable instead of calling `model.get_indexpath` twice

introt (changes requested):
Hi, and thanks for the PR!

The fix works great, and I couldn't spot any unintended side effects. However, I'd like to see some minor things cleaned up before it's merged; please see the attached comments.

Hope to see more PRs from you soon!

zim/plugins/pageindex/__init__.py
mrffr: I've replaced get_indexpath with get_selected_path which looks clearer to me.

zim/plugins/pageindex/__init__.py
```diff
@@ -473,6 +474,11 @@ def do_row_activated(self, treepath, column):
 		model = self.get_model()
 		treeiter = model.get_iter(treepath)
 		mytreeiter = model.get_user_data(treeiter)
+		selected_path = self.get_selected_path()
```
introt, line 477: `get_selected_path()` gives some overhead vs. a plain  `get_indexpath(mytreeiter)`, but I'd guesstimate its impact to be negligible.

introt (approved):
Thanks for cleaning up!

prahal:
Why isn't this PR merged?

jaap-karssenberg:
My bad. Somehow I thought it was merged already.

PR #2127, opened 2022-08-14
Update test job for macOS

dehesselle:
This PR improves the test job for macOS to take advantage of recent enhancements of the GitLab CI we use to build the macOS app.


dehesselle:
@jaap-karssenberg Please merge this so you don't get errors from the macOS test job when you push to the `develop` branch.

With these changes I've also taken measures to avoid such situations in the future.

PR #2134, opened 2022-08-21
Catch plugin not having command class

### LINKED ISSUE #2132: `object is not callable` when plugin has no Command
**Describe the bug**

When trying to call a command line command for a plugin, that does not provide a subclass of `Command`,
the following somewhat cryptic error is presented:

```
$ zim --plugin tasklist
ERROR: Exception in main()
  File "/usr/lib/python3.10/site-packages/zim/main/__init__.py", line 938, in main
    ZIM_APPLICATION.run(*argv[1:])
  File "/usr/lib/python3.10/site-packages/zim/main/__init__.py", line 708, in run
    cmd = build_command(args, **kwargs)
  File "/usr/lib/python3.10/site-packages/zim/main/__init__.py", line 623, in build_command
    obj = klass(cmd, pwd=pwd)
TypeError: 'NoneType' object is not callable
```

This comes down to `lookup_subclass` returning `None` if there is no such class with the `None` condition not being handled in `NotebookCommand`.

**To Reproduce**
1. Call `zim --plugin tasklist` or any other plugin that does not have a `Command` class
2. Observe error similar to the one shown above

**Expected behavior**
While this is an error case, a more precise error message would go a long way of explaining why the error is ocurring.
Looking at the code, I can already see, that there is a `UsageError('Could not load commandline command for plugin "%s"' % cmd)` prepared for this case that is just not triggered for missing `Command` classes.

**Screenshots**
n/a

* 0.74.3 - but I can also reproduce this on the current `master` branch (a08d7a10)
* OS: Arch Linux

**Additional context**
I already have a fix for this, I'll open a PR shortly

j6s:
Resolves #2132

### DIFF
--- a/zim/main/__init__.py
+++ b/zim/main/__init__.py
@@ -619,6 +619,8 @@ def build_command(args, pwd=None):
 				# Can't use following because log level not yet set:
 				# logger.debug('Error while loading: zim.plugins.%s.Command', cmd, exc_info=sys.exc_info())
 			raise UsageError('Could not load commandline command for plugin "%s"' % cmd)
+		if klass is None:
+			raise UsageError('Module %s has no commandline command' % cmd)
 	else:
 		if args and args[0].startswith('--') and args[0][2:] in commands:
 			cmd = args.pop(0)[2:]


dehesselle:
Thank you for your contribution. Please target your PR towards the develop branch. Sorry for the inconvenience, it's not documented clearly enough.

j6s:
Ah, I see - I could have looked at the branches more thourogly - unfortunately, there are differences between the branches and I'll have to cherrypick / rebase my commits onto `develop` in order to have a clean PR.

I'll request a review from you when I'm done

dehesselle:
Thank your for your understanding.
Unless the changes are (kind of) trivial - I haven't looked, I'm writing from my phone - better request a review from @introt , he's more qualified in regards to Zim's codebase than me.

j6s:
I can't actually request a review for this PR - but I updated it to point to `develop`

introt (changes requested):
Hi! Thought I felt my ears burning; delighted to make your acquaintance.

While the code does fix #2132, it seems to do so through raising either `TypeError: exceptions must derive from BaseException` for existing plugins or `UnboundLocalError: local variable 'klass' referenced before assignment` for nonexistent ones - either gets caught by the `except` on line 618; this is revealed by adding a `--debug` to the command line.

Without `--debug`, the only output is that of the `UsageError`: `ERROR: Could not load commandline command for plugin "$PLUGIN_NAME"`. I'd like the text to reflect eg. whether the plugin exists, has the command class, etc., while keeping the API (ie. not raising errors of other types inside `build_command`)

j6s:
Ah, I was wondering if I could raise a string in Python similar to throwing a string JS.
I figured that it worked, so it was valid. Thank you for pointing this out.

I moved the check outside of the try/catch statement and raised a new `UsageError` for this case.

Issue #2133, opened 2022-08-21
Tasklist: Allow opening from CLI


j6s:
**Is your feature request related to a problem? Please describe.**
I find it a bit cumbersome to open the tasklist using the mouse in the main zim window.
The ability to open the list easier and quicker would improve my workflow a lot.

**Describe the solution you'd like**
I stumbled over the following old discussion that outlined, that this could be solved by using a
`zim --plugin tasklist show_task_list` command, but it doesn't seem like it has been implemented
in the meantime as the E-Mail thread just trailed off.

PR #2135, opened 2022-08-21
Add command to show TaskList

j6s:
My humble attempt of implementing #2133.

I solved it a bit differently than outlined in the issue by using a `--show` option instead of an argument due to colissions with the `NOTEBOOK` argument.

This is my first bigger contribution to both a python codebase in general as fell as zim specifically. Don't hesitate to point stuff out that I could improve - I probably don't know about a bunch of interesting features in both.

### DIFF
--- a/zim/plugins/tasklist/__init__.py
+++ b/zim/plugins/tasklist/__init__.py
@@ -12,6 +12,7 @@
 # - force update, intialization
 
 
+from zim.main import GuiCommand
 from zim.plugins import PluginClass, find_extension
 from zim.actions import action
 from zim.config import StringAllowEmpty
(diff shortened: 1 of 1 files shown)


zim/plugins/tasklist/__init__.py
```diff
@@ -230,3 +231,62 @@ def _connect_tasklist_changed(self, widget):
...
+			self.show_task_list(window)
+
+		return window
+
+	def help(self):
+		print(usagehelp)
+
+	def show_task_list(self, window):
+		notebook = window.notebook
+		properties = TaskListPlugin().notebook_properties(notebook)
```
j6s, line 282: With my limited knowledge, this struck me as a bit strange: I get that I need to get the properties but I suspect that an instance of `TaskListPlugin` that I can use already exists somewhere else

jaap-karssenberg:
Yes, you should not create a new plugin object here. Instead import the
PluginManager, which is a singleton, and use that to get / load the plugin.
Since for this command it makes sense to load the plugin if not previously
loaded you can access it as follows:

   from zim.plugins import PluginManager
   plugin = PluginManager.load_plugin('tasklist')

Btw when opening the window, for the argument "show_inbox_next" please use
the plugin preference "plugin.preferences['show_inbox_next']".

Regards,

Jaap

jaap-karssenberg:
(PS. don't worry about the test fail on MacOS)

dehesselle:
Apologies for the breakage in macOS tests. Fixing requires https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/2177 to be merged and then you need to rebase.

PR #2143, opened 2022-08-25
Update GitHub actions

dehesselle:
Update to use latest versions.

Also remove superfluous whitespace characters.


dehesselle:
@jaap-karssenberg You can squeeze in one more little thing tonight, if you want.

PR #2177, opened 2022-10-06
Fix macOS tests in CI

dehesselle:
Due to updates in the underlying build system for macOS, the name of the variable to ignore errors related to files in /usr/local has changed. This commit updates the workflow accordingly.


dehesselle:
@jaap-karssenberg I broke this with the macOS release for 0.75.0. It's a trivial change, please merge once you find the time.

PR #2194, opened 2022-10-14
Use flatpak locale directory in flatpak

swsnr:
Check whether Zim is running inside flatpak and use the flatpak locale directory if so.

Inside flatpak locale is at `/app/share/locale`, but the standard locale directory of Python's `gettext` module is `/usr/share/locale`, hence zim fails to load translations in flatpak.

See https://github.com/flathub/org.zim_wiki.Zim/issues/19


swsnr:
Thank you for merging 🙏

PR #2202, opened 2022-10-22
TableEditor: hotfix for link visibility, documentation improvements

### LINKED ISSUE #1413: Update help entry on hyperlinks in Table cells
This is hopefully the sort of issue that any development team should like. Some feature actually works rather better than the Help documentation says.

In the "Superior Table Operations" page in the Help it says that 

`You can not use a special link text. The link text is the url, the mail-address or the Title of the linked page.`

In fact this can be done.

>`Links in cells can be entered like this:-`
>`[[https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues\|Zim wiki issues on Github]]`

The only tricky bit is the inclusion of a backslash before the vertical line character to "escape" it.

### LINKED ISSUE #2146: Tables (Table Editor plugin) render cells with links incorrectly
**Describe the bug**
Cells with links repeat content from row above instead of actual content.

**To Reproduce**
Enable "Table editor" plugin and insert following snippet:
```
| Foo         | Bar                   |
|:------------|:----------------------|
| 1           | 2345                  |
| abc def hij | klm **nop**           |
| foobar      | http://www.python.org |
```
Refresh the note. You will see "klm **nop**" instead of the link to python.org

**Expected behavior**
I should see the link.

**Screenshots**
image

 - Version 0.74.3, from repositories
 - OS: Kubuntu 22.04
 - Xorg
 - Language: Should not matter

introt:
* Fixes #2146 and improves documentation.
  * Spans now require a non-zero font size; one "[1024ths of a point](https://docs.gtk.org/Pango/pango_markup.html)" is small enough not to be visible even with longer links.
* Fixes #1413: updates manual to reflect how links inside tables actually work

### DIFF
--- a/data/manual/Plugins/Table_Editor/Superior_Table_Operations.txt
+++ b/data/manual/Plugins/Table_Editor/Superior_Table_Operations.txt
@@ -23,12 +23,12 @@ After formatting: 	{{./cell_after_text_formatting.png}}
 ==== Links in a  cell ====
 '''
 [[http://www.python.org]]
+[[http://www.python.org\|Python homepage]]
 [[foo@bar.org]]
 [[foo]]		Internal Link
 '''
 
-Note:	Only **one link per cell** can be opened. The first one is taken.
-		You can not use a special link text. The link text must be a URL, a mail-address or the title of a linked page.
+Note: Only **one link per cell** can be opened. The first one is taken.
 A link can be accessed with:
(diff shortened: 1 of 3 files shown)


introt:
Apologies for skipping local tests, though this does give me a chance to document the reason for the change in the commit message. I'll make sure to keep these two things in mind moving forward.

PR #2237, opened 2022-11-25
fix(pageview): respect orientation from exif metadata

### LINKED ISSUE #2228: Ignoring Image Exif rotation metadata
**Describe the bug**
Exif rotation metadata in images is ignored.

**To Reproduce**
Insert a jpg with rotation metadata

**Expected behavior**
Image should be displayed orientated per the metadata

**Screenshots**
image
this image displays upright in other image viewers

 - Version 0.75.1
 - OS:Linux

I believe this regression started with version 0.75

Toggoren:
Closes #2228

### DIFF
--- a/zim/gui/pageview/__init__.py
+++ b/zim/gui/pageview/__init__.py
@@ -53,7 +53,7 @@ from zim.gui.widgets import \
 	Dialog, FileDialog, QuestionDialog, ErrorDialog, \
 	IconButton, MenuButton, BrowserTreeView, InputEntry, \
 	ScrolledWindow, \
-	rotate_pixbuf, populate_popup_add_separator, strip_boolean_result, \
+	populate_popup_add_separator, strip_boolean_result, \
 	widget_set_css, \
 	LEFT_PANE, RIGHT_PANE, BOTTOM_PANE, PANE_POSITIONS
 from zim.gui.actionextension import ActionExtensionBase, os_default_headerbar
@@ -484,6 +484,33 @@ def image_file_get_dimensions(file_path):
 		raise AssertionError('Could not get size for: %s' % file_path)
 
 
+def _convert_pillow_image_to_pixbuf(image) -> GdkPixbuf.Pixbuf:
+	# check if there is an alpha channel
+	if image.mode == 'RGB':
+		has_alpha = False
+	elif image.mode == 'RGBA':
+		has_alpha = True
+	else:
+		raise ValueError('Pixel format {fmt} can not be converted to Pixbuf for image {image}'.format(
+			fmt=image.mode, image=image,
+		))
+
+	# convert to GTK pixbuf
+	data_gtk = GLib.Bytes.new_take(image.tobytes())
+
+	return GdkPixbuf.Pixbuf.new_from_bytes(
+		data=data_gtk,
+		colorspace=GdkPixbuf.Colorspace.RGB,
+		has_alpha=has_alpha,
+		# GTK docs: "Currently only RGB images with 8 bits per sample are supported"
+		# https://docs.gtk.org/gdk-pixbuf/ctor.Pixbuf.new_from_bytes.html#description
+		bits_per_sample=8,
+		width=image.width,
+		height=image.height,
+		rowstride=image.width * (4 if has_alpha else 3),
+	)
+
+
 def image_file_load_pixels(file, width_override=-1, height_override=-1):
 	"""
 	Replacement for GdkPixbuf.Pixbuf.new_from_file_at_size(file.path, w, h)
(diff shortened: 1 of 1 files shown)


zim/gui/pageview/__init__.py
```diff
@@ -515,10 +515,44 @@ def image_file_load_pixels(file, width_override=-1, height_override=-1):
 	except:
 		logger.debug('GTK failed to read image, using Pillow fallback: %s', file.path)
 
+		import PIL
```
introt: Nitpick: one import is enough: `from PIL import Image, __version__ as pillow_version_string`

zim/gui/pageview/__init__.py
```diff
@@ -515,10 +515,44 @@ def image_file_load_pixels(file, width_override=-1, height_override=-1):
...
+				# https://pillow.readthedocs.io/en/stable/releasenotes/6.0.0.html#added-exif-class
+				pillow_version = tuple(map(int, PIL.__version__.split('.')))
+				if pillow_version >= (6, 0, 0):
+					exif = image.getexif()
+				else:
+					# noinspection PyUnresolvedReferences,PyProtectedMember
+					exif = image._getexif()  # noqa: WPS437
+					if not exif:
+						exif = {}
+				orientation_value = exif[tag_id_orientation] if tag_id_orientation in exif else 0
```
introt: `0` gets handled by the default case, which is the same as `1`. Use `else 1` here to save some cycles, and document the decision with a comment.

zim/gui/widgets.py
```diff
@@ -350,14 +350,24 @@ def rotate_pixbuf(pixbuf):
 	o = pixbuf.get_option('orientation')
```
introt: tbh I didn't know we had this one already - could you add a comment to the PIL rotater describing when the fallback is needed? Afaik PIL images end up as pixbufs anyway, so I fail to see why this function can't be used to rotate those.

If the use case is `get_option` failing, PIL should only be used to parse the EXIF tag, and this function should receive it as an optional kwarg.

"Older gtk version" should no longer be a possible issue; we require Python 3.6 [released in December 2016](https://www.python.org/downloads/release/python-360/), while 2.12 required for `get_option` dates back to the early 2010s (`flip` and `rotate_simple` require 2.6).

Zim being a 10+ year old project originally built on Python 2 introduces this kind of weird version kinks, but don't worry - we don't expect this kind of digging from contributors, it's just me being a perfectionist ;)

zim/gui/pageview/__init__.py
```diff
@@ -515,10 +515,44 @@ def image_file_load_pixels(file, width_override=-1, height_override=-1):
...
+				elif orientation_value == 7:
+					image = image.transpose(Image.FLIP_TOP_BOTTOM).transpose(Image.ROTATE_90)
+				elif orientation_value == 8:
+					image = image.transpose(Image.ROTATE_90)
+				else:
+					# Wrong exif data ?
+					pass  # noqa: WPS420
+				return image
 
+			img_pil = rotate_pillow_image_by_exif_orientation(img_pil)
```
introt: Rotation should be done after the possible resize - less pixels to flip.

zim/gui/pageview/__init__.py
```diff
@@ -515,10 +515,44 @@ def image_file_load_pixels(file, width_override=-1, height_override=-1):
...
 
+		import PIL
 		from PIL import Image # load Pillow only if necessary
 
 		with Image.open(file.path) as img_pil:
+			def rotate_pillow_image_by_exif_orientation(image: Image.Image) -> Image.Image:
+				tag_id_orientation = 0x0112  # noqa: WPS339
+				# https://pillow.readthedocs.io/en/stable/releasenotes/6.0.0.html#added-exif-class
+				pillow_version = tuple(map(int, PIL.__version__.split('.')))
+				if pillow_version >= (6, 0, 0):
```
introt: Good reference linking! Debian [old-stable still has 5.4.1](https://tracker.debian.org/pkg/pillow) (last updated on 2022-02-05) on [Python 3.7.3](https://tracker.debian.org/pkg/python3-defaults), so support is appreciated.

introt (changes requested):
Thanks for the PR, and welcome to Zim! Nice work with this patch; I've added a few comments/change requests related to (Zim's) long-term maintenance, but besides those looking forward to getting this merged!

zim/gui/widgets.py
introt: (for reference, old-old-stable Debian has [2.36.5](https://tracker.debian.org/pkg/gdk-pixbuf) (ie. new enough) and [Python 3.5.3](https://tracker.debian.org/pkg/python3-defaults) (which we no longer support))

introt:
See new comments at the original issue - resized images are not being rotated, likely due to falling back to Pillow.

Toggoren:
> See new comments at the original issue - resized images are not being rotated, likely due to falling back to Pillow.

The problem is in this line. [When we resize an image, the metadata is lost.](https://github.com/GNOME/gdk-pixbuf/blob/19512e34675b658e7509957b4f99bff2c3d56979/gdk-pixbuf/gdk-pixbuf-scaled-anim.c#L130-L131)
https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/ef9181c0d7084cd09c5e7349279a06e94a39e45a/zim/gui/pageview/__init__.py#L510

Therefore, the image does not change here.
https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/ef9181c0d7084cd09c5e7349279a06e94a39e45a/zim/gui/pageview/__init__.py#L513

Initially, I was going to make a solution for unscaled images only.
Because I thought [it was an unscaled image](https://user-images.githubusercontent.com/28350144/202887547-1eebd1c8-43a7-420c-87a6-13dda074ee13.png). 🤣

zim/gui/pageview/__init__.py
Toggoren: Fixed it.

zim/gui/pageview/__init__.py
Toggoren: Fixed it.

zim/gui/pageview/__init__.py
Toggoren: 😔 Oops. Fixed it.

zim/gui/pageview/__init__.py
```diff
@@ -53,7 +53,7 @@
 	Dialog, FileDialog, QuestionDialog, ErrorDialog, \
 	IconButton, MenuButton, BrowserTreeView, InputEntry, \
 	ScrolledWindow, \
-	rotate_pixbuf, populate_popup_add_separator, strip_boolean_result, \
```
introt, line 56: If we're intending to keep rotate_pixbuf around for some other use case, I'd like to see your earlier changes (support for flips etc) incorporated into it

zim/gui/pageview/__init__.py
```diff
@@ -484,6 +484,33 @@ def image_file_get_dimensions(file_path):
 		raise AssertionError('Could not get size for: %s' % file_path)
 
 
+def _convert_pillow_image_to_pixbuf(image) -> GdkPixbuf.Pixbuf:
```
introt, line 487: Given this function can only be called from image_file_load_pixels when Pillow is loaded, why was it moved here?

zim/gui/pageview/__init__.py
```diff
@@ -494,63 +521,93 @@ def image_file_load_pixels(file, width_override=-1, height_override=-1):
...
+		error_message = f'No available fallback for load this image: {file.path}'
+		logger.debug(error_message)
+		raise TypeError(error_message)
+
+	# Let's try to find and remember the orientation before scaling,
+	# 	because we lose metadata when changing images.
+	orientation = None
+	mimetype = file.mimetype()
+	if mimetype in {'image/jpeg', 'image/tiff'}:
+		# Gtk can detect orientation in jpeg|tiff images only
```
introt, line 558: PNG images don't have an orientation tag: https://stackoverflow.com/questions/13022946/how-to-rotate-a-png-image-using-exiftool

zim/gui/pageview/__init__.py
```diff
@@ -494,63 +521,93 @@ def image_file_load_pixels(file, width_override=-1, height_override=-1):
...
-			height_override = int(h * width_override / w)
-		else:
-			width_override = int(w * height_override / h)
-
-	# Let GTK try reading the file
+	need_switch_to_fallback = True
 	try:
 		pixbuf = GdkPixbuf.Pixbuf.new_from_file(file.path)
+	except GLib.GError:
+		logger.debug(f'GTK failed to read image, using Pillow fallback: {file.path}')
```
introt, line 528: Nitpick: logging should use `%s` - https://docs.python.org/3/howto/logging.html#optimization

zim/gui/pageview/__init__.py
```diff
@@ -494,63 +521,93 @@ def image_file_load_pixels(file, width_override=-1, height_override=-1):
...
+		# 	if it is not available, we will display image it as is.
+		try:
+			# noinspection PyUnresolvedReferences
+			from PIL import Image, UnidentifiedImageError, __version__ as pillow_version_string
+		except ImportError:
+			pass
+		else:
+			pillow_version = tuple(map(int, pillow_version_string.split('.')))
+			try:
+				with Image.open(file.path) as image:
```
introt, line 572: Loading the image twice seems excessive, could the reading of exif data be merged to the prior read? Also, I'm not sure whether supporting EXIF orientation tags in PNGs is worth the cost of loading Pillow, given their rarity compared to `pasted_image.png`s https://stackoverflow.com/questions/9542359/does-png-contain-exif-data-like-jpg/9576717 - but I'll leave that to Jaap to decide.

zim/gui/pageview/__init__.py
```diff
@@ -494,63 +521,93 @@ def image_file_load_pixels(file, width_override=-1, height_override=-1):
...
 
-			# check if there is an alpha channel
-			if img_pil.mode == 'RGB':
-				has_alpha = False
-			elif img_pil.mode == 'RGBA':
-				has_alpha = True
+	w, h = pixbuf.get_width(), pixbuf.get_height()
+	b_size_override = width_override > 0 or height_override > 0
+	if b_size_override and (width_override <= 0 or height_override <= 0):
+		if orientation in {5, 6, 7, 8}:
```
introt, line 591: Before, we resized prior to rotation. I do agree that taking the orientation into account makes sense, but does this break the dimensions of all existing exif-rotated images?

zim/gui/pageview/__init__.py
```diff
@@ -494,63 +521,93 @@ def image_file_load_pixels(file, width_override=-1, height_override=-1):
...
+				width_override = int(w * height_override / h)
 
-	return pixbuf
+	if b_size_override:
+		# do not use new_from_file_at_size() here due to bug in Gtk for GIF images, see issue #1563
+		pixbuf = pixbuf.scale_simple(width_override, height_override, GdkPixbuf.InterpType.BILINEAR)
 
+	pixbuf.set_option('orientation', f'{orientation}')
+	#      ^^^ be sure to attach the tag for magic from stdlib in next line.
+	return GdkPixbuf.Pixbuf.apply_embedded_orientation(pixbuf)
```
introt, line 610: Nice use of the existing facilities!

introt (commented):
Sorry for the late review! Not sure if you closed this by accident; here's a few comments. I approve your aim for common resizing code && usage of apply_embedded_orientation, but imo we should optimize for regular ol' PNGs at the cost of spottier exif support

zim/gui/pageview/__init__.py
Toggoren: When I started working, I didn't read this line. 🤦‍♂️

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/1ccaffc181671a71b6d66cbe99ae14688db1fc6c/README.md?plain=1#L41

``GdkPixbuf.Pixbuf.apply_embedded_orientation`` available since **2.12** 🤦‍♂️

therefore, this code:

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/1ccaffc181671a71b6d66cbe99ae14688db1fc6c/zim/gui/widgets.py#L341-L361

is completely equivalent to this сode
```python3
def rotate_pixbuf(pixbuf):
	'''Rotate the pixbuf to match orientation from EXIF info.
	This is intended for e.g. photos that have EXIF information that
	shows how the camera was held.
	@returns: a new version of the pixbuf or the pixbuf itself.
	'''
	return GdkPixbuf.Pixbuf.apply_embedded_orientation(pixbuf)
```

Currently, ``rotate_pixbuf`` is used in two places.

1. https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/e4b04df4737cde00d0c159b1007361204fa7856c/zim/gui/pageview/__init__.py#L514
2. https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/1ccaffc181671a71b6d66cbe99ae14688db1fc6c/zim/plugins/attachmentbrowser/thumbnailer.py#L102

We are currently working on the first one.

And if we replacing the second place with  
```python3
pixbuf = GdkPixbuf.Pixbuf.apply_embedded_orientation(pixbuf)
```
we can completely remove ``rotate_pixbuf``.

Can we do that?

zim/gui/pageview/__init__.py
Toggoren: to reduce cognitive complexity.
If you think it's gotten worse, tell me. I'll hide it back.

zim/gui/pageview/__init__.py
Toggoren: Formatted string literals faster than printf-style  
Based this SO answer https://stackoverflow.com/a/48465349

zim/gui/pageview/__init__.py
Toggoren: Pillow uses lazy loading  
I specifically checked 🧐  
And everything is bad 😯  
PNGs without orientation are apparently loaded completely 😭  
Unlike other options 

https://gist.github.com/Toggoren/0adfb87dab4e1c2f5b4e31e78853752c#file-gistfile1-py-L159-L167

```
f0 | mean: 0.027517 0.029340 0.937864 | median: 0.024205 0.025645 0.943845
 ^                ^        ^        ^
 |                |        |        |
 |                |        |        Just X/Y
 |                |        (Y): time required to load image into pixbuf
 |                (X): time required to load image exif data
 Orientation value. 0 means no data
```

I don't know what to do 😕

zim/gui/pageview/__init__.py
Toggoren: Nope.
We need to swap the height with the width to get the correct image proportions for orientation in {5, 6, 7, 8}
I wrote a small test case to check if it works https://github.com/Toggoren/PainNotes/tree/master/PainNotes

zim/gui/pageview/__init__.py
introt: I believe you're correct, nice catch! We can indeed retire our `rotate_pixbuf` implementation. `return GdkPixbuf.Pixbuf.apply_embedded_orientation(pixbuf)` plus a `logger.warning('Using deprecated function "zim.XXX.rotate_pixbuf" - please update your code to use "GdkPixbuf.Pixbuf.apply_embedded_orientation" instead')`

zim/gui/pageview/__init__.py
introt: I get your point. It could be useful elsewhere (eg. for plugin authors) as well, maybe worth promoting to be a part of our API? You can leave it here and we can ask Jaap.

zim/gui/pageview/__init__.py
introt: I'll take that back then :)

zim/gui/pageview/__init__.py
introt: I appreciate your thoroughness! PNG exif data just is kind of a hacky addition to the spec and there ain't two ways about that. Well - we could offer to rewrite PNGs with their embedded orientation as an extra tool :P

I'd just let PNGs retain their orientation, but if you want more [pain](https://github.com/Toggoren/PainNotes) another way would be a hidden configuration preference to enable PNG exif support - up to you.

zim/gui/pageview/__init__.py
introt: Sorry, I meant whether resized images appear the same size before and after applying this patch - though now I realize that might be impossible to test given that rotation was broken before :D

jaap-karssenberg:
I have this pull request flagged for merging in 0.75.2 point release. Can you confirm it is ready to merge; or are there open items remaining? If open items are remaining, or if you see testing risks, I will push it forward to the 0.76 release.

introt:
Open items that require code changes:

- [ ] retire rotate_pixbuf with a deprecation warning

These require input from @jaap-karssenberg :
- [ ] should _convert_pillow_image_to_pixbuf be part of our API?
- [ ] should we support PNG rotation metadata at the cost of loading each PNG twice?

Imo PNG rotation metadata support should be optional if included at all; it's quite rare to encounter in the wild

jaap-karssenberg:
Sorry it took me some time to go over the patch

How I would like to proceed:
1. Put all these image manipulation functions in a separate module `zim.gui.widgets.images`
2. We can remove `rotate_pixbuf()` if we also replace its usage in the attachment browser plugin
3. The image manipulation functions need some good test cases with & without PIL to ensure it works as intended

jaap-karssenberg:
Merged while refactoring pageview module. Also made thumbnailing use the same functions, so we also allow thumbnailing of images not supported by Pixbuf now :)

PR #2117, opened 2022-08-04
Fix ControlledDict.pop on Python 3.11

oturpe:
The fix for Python Issue BPO-27275 changes `OrderedDict.pop()` behavior
so that it does not call `__delitem__` any more.
On the other hand, `ControlledDict` expects such call to happen
to emit the `changed`' event.
Fixed by overriding `pop()` in `ControlledDict`
so that it explictly emits `changed`.

Fixes #2087

PR #2245, opened 2022-12-01
Fix workflows: bump and lower versions on Linux

introt:
Python 3.6 is EOL and not supported on ubuntu-latest: https://github.com/actions/setup-python/issues/544#issuecomment-1332535877

Bumps upper Python version to 3.11 as we are compatible thanks to #1821 and #2117.


introt:
Broken; see commits at linked issue for reference

PR #2308, opened 2023-02-01
Replace usage of ``os.devnull`` with ``subprocess.DEVNULL``

Toggoren:


jaap-karssenberg:
Can you provide a rationale for this change ?

Toggoren:
> Can you provide a rationale for this change ?

Just a more fashionable  and modern way to do the same.

Similar merged pull requests:
- https://github.com/apache/airflow/pull/19354
- https://github.com/ipython/ipython/pull/13932
- https://github.com/conda/conda-build/pull/4565

introt (approved):
[Introduced](https://docs.python.org/3/library/subprocess.html) in Python 3.3, seems to do some file descriptor handling: https://github.com/python/cpython/issues/50120#issuecomment-1093466657 - though with `close_fds=True` it's not too relevant here.

zim/applications.py
```diff
@@ -311,7 +311,7 @@ def run(self, args=None, cwd=None):
 			try:
 				p = subprocess.Popen(argv,
 					cwd=cwd,
-					stdout=open(os.devnull, 'w'),
+					stdout=subprocess.DEVNULL,
 					stderr=subprocess.PIPE,
 					bufsize=4096,
 					close_fds=True
```
introt, line 317: `close_fds` defaults to `True` [from 3.2 onwards](https://docs.python.org/3/library/subprocess.html) while `subprocess.DEVNULL` was introduced in 3.3. If we're changing the code for modernization's sake, it should be done consistently on a given level (for example, a whole function/call at a time) or else we'll end up with a patchwork of modern and legacy code.

introt (changes requested):

zim/applications.py
introt: @jaap-karssenberg sorry for the hasty initial review; do you think we should have some guidelines on refactoring? Given the age of the code base, we do have some technical debt on tab, some of which can be paid now that we've bumped to 3.6+.

randomguy:
AFAIS this only adds a layer of indirection. A quick glance at os.py and subprocess.py from a Python 3.10 Linux install:
- os modules devnull is equivalent to os.path.devnull
- os.path is ntpath or posixpath determined dynamically on load and thus sets /dev/null depending on the platform 
- subprocess.DEVNULL is an int constant determining various branching in subrpocess module (they should have underscored it by convention since it is a private member?)
- all the branching in subprocess module depending on subprocess.DEVNULL constants calls back to os.devnull AFAIS

Issue #2305, opened 2023-01-30
Unhandled exception in script


dhana25252:
Failed to execute script zim_launchh due to unhandled exception: username

**Screenshots**

![image](https://user-images.githubusercontent.com/13852137/215369553-6a4a8595-a1ba-498a-9f97-448bfa14760b.png)

**Environment (please complete the following information):**
 - Version zim-desktop-wiki-0.75.1-setup-w64_x86 -- specifically mention it if you use the development branch or snapshot PPA
 - OS: windows 10
 - Language: N/A

dhana25252:
Hi Team, I'm using zim-wiki for years, it is very and an excellent program, I never saw an error like this.
Kindly reach me for any other further information if needed from my end., thanks, team.
"dhana.varadaraju@gmail.com"

dhana25252:
FYI: I tried uninstalling and reinstalling, the zimwiki run one time and again its reports the same error as I uploaded the screenshot, please take a look at that error.
Version: zim-desktop-wiki-0.75.1-setup-w64_x86
OS: windows 10

dhana25252:
Unknowingly closed, please ignore

PR #2313, opened 2023-02-04
#2305 updated to handle exception

lazearrr:
In regard to open issue #2305 

changed
os.environ['USER'] = os.environ['USERNAME']
to
os.environ['USER'] = os.environ.get['USERNAME', 'Unknown']

in order to handle KeyError: 'USERNAME'

P.S. first ever open source contribution so please let me know if I'm mistaken

### DIFF
--- a/zim/__init__.py
+++ b/zim/__init__.py
@@ -191,7 +191,7 @@ if os.name == 'nt':
 	# Windows specific environment variables
 	# os.environ does not support setdefault() ...
 	if not 'USER' in os.environ or not os.environ['USER']:
-		os.environ['USER'] = os.environ['USERNAME']
+		os.environ['USER'] = os.environ.get['USERNAME', 'Unknown']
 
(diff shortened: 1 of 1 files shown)


zim/__init__.py
```diff
@@ -191,7 +191,7 @@
 	# Windows specific environment variables
 	# os.environ does not support setdefault() ...
 	if not 'USER' in os.environ or not os.environ['USER']:
-		os.environ['USER'] = os.environ['USERNAME']
```
jaap-karssenberg, line 194: Good catch to use `get()` here to avoid exceptions. However, would prefer to keep it defaulting to `None`. Having `None` value if also `USERNAME` is not available allows the fallback in line 209 to handle it.

jaap-karssenberg:
@Lazearrr sure welcome to contribute - good catch to prevent exception here. I've added a small code comment before merging. Please have a look.

PR #2318, opened 2023-02-07
Fix flashing console window on Windows

### LINKED ISSUE #2317: Plugin(Version Control) flashing console window on Windows
**Describe the bug**
When running under Windows this plugin (and some others) causes a console window to quickly flash on the screen.

**To Reproduce**
1. Activate this plugin.
2. Check ``Autosave version when the notebook is closed``
3. Close ``Zim``

**Expected behavior**
The console window should not be shown.

**Screenshots**

https://user-images.githubusercontent.com/94177064/217226914-09359117-dab0-418c-ba3e-b16880846ef4.mp4

 - Version: at least version 0.75.x
 - OS: Windows
 - Language: Any

Toggoren:
Closes #2317

### DIFF
--- a/zim/applications.py
+++ b/zim/applications.py
@@ -353,8 +353,13 @@ class Application(object):
 		if TEST_MODE:
 			return TEST_MODE_RUN_CB(argv)
 
+		startupinfo = None
+		if os.name == 'nt':
+			startupinfo = subprocess.STARTUPINFO()
+			startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
+			startupinfo.wShowWindow = subprocess.SW_HIDE
 		try:
-			p = subprocess.Popen(argv, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+			p = subprocess.Popen(argv, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, startupinfo=startupinfo)
 		except OSError:
 			if _CAN_CALL_FLATPAK_HOST_COMMAND:
 				p = subprocess.Popen(_FLATPAK_HOSTCOMMAND_PREFIX + argv, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)


jaap-karssenberg:
Thanks for the patch - not at all familiar with this part of the windows API myself. 

As a bonus, could we connect this to the `Terminal` option in `zim/gui/applications.py` ?  For linux it is handled by selecting a terminal emulator from a hardcoded list. Seems that on windows we could just set a flag on the application object and put a conditional check in this patch - is that a correct assumption on what this does ?

Toggoren:
> Thanks for the patch - not at all familiar with this part of the windows API myself.
> 
> As a bonus, could we connect this to the `Terminal` option in `zim/gui/applications.py` ? For linux it is handled by selecting a terminal emulator from a hardcoded list. Seems that on windows we could just set a flag on the application object and put a conditional check in this patch - is that a correct assumption on what this does ?

Sorry, my knowledge of the English language is not enough to sustain a dialogue.

I will write some information in the hope that it will help.

1. This patch is a logical continuation of this patch https://bugs.launchpad.net/zim/+bug/1075113
2. The situation with ``Windows``
    1. ``subprocess.Popen`` is a wrapper over ``CreateProcess``
    2. If the [Subsystem](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format) field in the [executable](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format) has the value [IMAGE_SUBSYSTEM_WINDOWS_GUI](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_optional_header32), calling ``CreateProcess`` causes a window to appear.
    3. To prevent a window from popping up, you need to correctly form [STARTUPINFO](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-startupinfow) structure
    4. In all the examples I found, the formation of the structure looked like this https://github.com/python/cpython/blob/20cf32e761fb9eaccc142415b389998896869263/Lib/subprocess.py#L1481-L1482

PR #2337, opened 2023-03-10
Updated diagrameditor and sequencediagrameditor plugins to generate SVG

radupotop:
SVG images look significantly better in print and on the web. Have the sequence diagram editor generate SVG images instead of PNG.   

Tested on local and working fine.

Sample output files attached:

Before 
seq

After
seq

### DIFF
--- a/zim/plugins/__init__.py
+++ b/zim/plugins/__init__.py
@@ -30,7 +30,7 @@ to let plugins extend specific application objects.
 '''
 
 
-from gi.repository import GdkPixbuf, GObject
+from gi.repository import GObject
 import types
 import os
 import sys
--- a/zim/plugins/__init__.py
+++ b/zim/plugins/__init__.py
@@ -736,12 +736,6 @@ class PluginClass(ConnectorMixin):
 
 		return properties
 
-	def supports_image_format(self, fmt):
-		'''
-		Check if an image format is supported by GDK.
-		'''
-		return fmt in (f.name for f in GdkPixbuf.Pixbuf.get_formats())
-
 	@classmethod
 	def lookup_subclass(pluginklass, klass):
 		'''Returns first subclass of C{klass} found in the module of
--- a/zim/plugins/__init__.py
+++ b/zim/plugins/__init__.py
@@ -30,7 +30,7 @@ to let plugins extend specific application objects.
 '''
 
 
-from gi.repository import GObject
+from gi.repository import GdkPixbuf, GObject
 import types
 import os
 import sys
@@ -735,6 +735,12 @@ class PluginClass(ConnectorMixin):
 					pass
 
 		return properties
+
+	def supports_image_format(self, fmt):
+		'''
+		Check if an image format is supported by GDK.
+		'''
+		return fmt in (f.name for f in GdkPixbuf.Pixbuf.get_formats())
 
 	@classmethod
 	def lookup_subclass(pluginklass, klass):
--- a/zim/plugins/sequencediagrameditor.py
+++ b/zim/plugins/sequencediagrameditor.py
@@ -2,6 +2,7 @@
 # Copyright 2011 Greg Warner <gdwarner@gmail.com>
 # (Pretty much copied from diagrameditor.py)
 
+import logging
 
 from zim.plugins import PluginClass
 from zim.plugins.base.imagegenerator import ImageGeneratorClass, BackwardImageGeneratorObjectType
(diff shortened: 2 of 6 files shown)


jaap-karssenberg:
Unfortunately not all installs support SVG image format out of the box. So
should be at least an user option - or automatic check on the Gtk.Pixbuf
API whether SVG is supported.

-- Jaap

zim/plugins/sequencediagrameditor.py
```diff
@@ -41,28 +49,41 @@ class BackwardSequenceDiagramImageObjectType(BackwardImageGeneratorObjectType):
...
 	def __init__(self, plugin, notebook, page):
 		ImageGeneratorClass.__init__(self, plugin, notebook, page)
 		self.diagfile = TmpFile('seqdiagram.diag')
 		self.diagfile.touch()
-		self.pngfile = LocalFile(self.diagfile.path[:-5] + '.png') # len('.diag') == 5
 
 	def generate_image(self, text):
 		# Write to tmp file
 		self.diagfile.write(text)
+		self.imgfile = LocalFile(self.diagfile.path[:-5] + self.imagefile_extension) # len('.diag') == 5
```
radupotop, line 78: Moved imgfile generation outside of init, otherwise it's buggy when toggling formats in the user prefs.

radupotop:
@jaap-karssenberg Added user preference now

radupotop:
Unrelated to this PR, just discovered an issue in the thumbnailer plugin:

```
ERROR: Exception in thumbnail queue
  File "/home/tail/Projects/zim-desktop-wiki-1/zim/plugins/attachmentbrowser/thumbnailer.py", line 169, in _thread_main
    thumbfile, pixbuf = self._thumbmanager.get_thumbnail(file, size)
  File "/home/tail/Projects/zim-desktop-wiki-1/zim/plugins/attachmentbrowser/thumbnailer.py", line 261, in get_thumbnail
    if mtime and int(mtime) == int(file.mtime()):
ValueError: invalid literal for int() with base 10: '1678465374.410669'
DEBUG: Add window: MainWindow
```

radupotop:
@jaap-karssenberg Any thoughts on the current proposal?

jaap-karssenberg:
Sorry, haven't had time to look into it this week - will come back to you

zim/plugins/__init__.py
```diff
@@ -736,6 +736,12 @@ def notebook_properties(self, notebook):
 
 		return properties
 
+	def supports_image_format(self, fmt):
+		'''
+		Check if an image format is supported by GDK.
+		'''
+		return fmt in (f.name for f in GdkPixbuf.Pixbuf.get_formats())
+
```
jaap-karssenberg: Seems better to me to put this helper method in the more specialized `zim/plugins/base/imagegenerator.py` base class. Or otherwise as a function in `zim/gui/widgets.py` (like the  `rotate_pixbuf()`  funtion)

jaap-karssenberg:
@radupotop just added one comment on where to put the helper method. Also noticed test suite fails due to `plugin_preferences` being defined as a property instead of a list or tuple. Other than these two items patch is OK with me and willing to merge.

Kind regards,

Jaap

radupotop:
@jaap-karssenberg I've made the changes if you can take another look, thanks!

jaap-karssenberg:
I see two test failure pop up
- On python 3.6 the pixbuf.name attribute doesn't exist -- probably there is a "get_name()" in the API ?
- The preference that was introduced is not documented in the manual page

THese are things I can fix myself, but afraid won't get to them this week.

--- Jaap

radupotop:
@jaap-karssenberg Should be sorted now, thanks

jaap-karssenberg:
@radupotop sorry, but I'm afraid the test on the documentation still fails. 

You can see the results on GitHub test runs here: https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/2337/checks
(You can ignore the segmentation fault for the windows run, that one also fails on the main branch since recent commit, still investigating why.)

Locally you can run the test suite by running the `./test.py` script in the source folder. For this one in particular, the fast version should be sufficient, which runs with `./test.py --ff`

radupotop:
@jaap-karssenberg Should be sorted now.

Probably unrelated, I get the following when running the tests locally with `test --ff`

```
Zim unitest Test Report

772 tests run
72 skipped
0 errors
1 failures
FAIL: runTest (tests.preferencesdialog.TestManual)

  File "/home/radu/repos/zim-desktop-wiki-1/tests/preferencesdialog.py", line 36, in runTest
    self.assertTrue(label in manual, 'Preference "%s" not documented in manual page' % label)
AssertionError: False is not true : Preference "Show the cursor even for pages that cannot be edited" not documented in manual page
```

jaap-karssenberg:
Commit merged - thanks for contributing !

PR #2375, opened 2023-06-06
Fix typo in CONTRIBUTING.md

m-holger:


jaap-karssenberg:
Agree it is a typo, but since it is a verbatim quote of an email, I prefer note to touch it

PR #2379, opened 2023-06-08
Use Homebrew to run test suite on macOS

### LINKED ISSUE #2369: macOS CI: 10.15 image no longer working
**Describe the bug**
It was announced [a long time ago](https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/) and although the original timeline was delayed, macOS 10.15 is no longer available in GitHub CI.

**To Reproduce**
Push to development branch. A job gets created, remains in pending state for a while and finally fails as it is not able to acquire a runner with the requested image.

**Background information**
I have actually looked into upgrading macOS CI a long time ago, but was never able to figure out why the (otherwise unmodified) job fails on newer macOS versions.

**Next steps**
Look into updating macOS CI once again and if it still fails, remove it from the workflow.

dehesselle:
Fixes https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/2369

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,13 +52,8 @@ jobs:
 
 
   test_macos:
-    runs-on: macos-10.15
-    env:
-      # a fixed path is required to use precompiled dependencies
-      WRK_DIR: /Users/Shared/work
-      # this contains a link to the job artifact we need
-      GITLAB_CI_YAML: https://gitlab.com/dehesselle/zim_macos/-/raw/master/.gitlab-ci.yml
-    name: Python 3.x on macos-10.15
+    runs-on: macos-12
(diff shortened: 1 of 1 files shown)


dehesselle:
Noteworthy: the only way to get the test suite to pass is to disable the `TestPlugins()` test case for now (in `tests/plugins.py`), otherwise a segfault will render the whole test job useless. Reason is unknown at this point. Requires someone to investigate.

dehesselle:
@jaap-karssenberg Please merge so that macOS doesn't ruin the tests any longer.

jaap-karssenberg:
Thanks!

PR #2383, opened 2023-06-11
Allow URLs to end with underscore

### LINKED ISSUE #2380: Pasting URL ending with underscore will generate a hyperlink ignoring the final underscore
**Describe the bug**
When pasting a URL ending with underscore, the final underscore is not made part of the auto-generated hyperlink, causing the link to be incorrect.

**To Reproduce**
1. Copy a URL ending with underscore, such as https://example.com/demo_
2. Paste it in Zim
3. `https://example.com/demo` (without underscore) is hyperlinked (in blue), while the trailing underscore `_` has no link associated (in black).

**Expected behavior**
The whole `https://example.com/demo_` should generate a hyperlink.

**Screenshots**
image
It's hard to see, but the underscore is not blue. To demonstrate more clearly:

image
Pointer before underscore: pointing hand, hyperlink works

image
Pointer after underscore: no pointing hand, hyperlink doesn't work

 - Version: 0.75.1 (installed via Ubuntu PPA)
 - OS: Linux Ubuntu 22.04
 - Language: English, but URL is international

**Additional context**
This is critical when pasting Twitter links: often, name parkers take all the common names, so often, new legit studios will suffix their Twitter with `_`. Failing to add it will fall on unrelated accounts.

introt:
Fixes #2380

What is the `url_trailing_punctuation` list based on?

### DIFF
--- a/zim/formats/wiki.py
+++ b/zim/formats/wiki.py
@@ -101,7 +101,7 @@ url_re = re.compile(
 	re.VERBOSE
 )
 
-url_trailing_punctuation = ('?', '!', '.', ',', ':', '*', '_', '~', "'", '"')
+url_trailing_punctuation = ('?', '!', '.', ',', ':', '*', '~', "'", '"')
 	# do not add ";" here, it is handled separatedly in the function
 
 


jaap-karssenberg:
It is based on the Github Flavored Markdown spec: https://github.github.com/gfm/#autolinks-extension-

The rationale is that the "_" is part of wiki markup and possible ambiguity in the parser is resolved by this rule. 

In the Dumper this rule is also checked, and URL links that do have trailing characters get decorated link style. In the related bug report it is the decoration that goes wrong in my mind.

Issue #2012, opened 2022-04-17
Link Map plugin does not work in Windows 10


frogmouse:
The Link Map plugin requires the external modules _graphviz_ and _xdot_ to be installed. Once this is done, Zim continues to report that _xdot_ doesn't exist.  This is true in the sense that _xdot_ is not functional -- it cannot find the module '_gi'_ from _graphviz_ -- though it reports no errors.

The [upstream developer is aware of this](https://github.com/jrfonseca/xdot.py/issues/70), has suggested a work-around that does not fix the problem for me, and has marked this problem "wontfix".

The plugin works as it should on linux, and is a valuable addition to Zim.   I'd like to be able to use it when running windows.

fidthorn2vc:
If run the command "import xdot" in python , it says "ModuleNotFoundError: No module named 'numpy'".

Plugin worked for me after installation mingw-w64-x86_64-python-numpy package in msys2.

introt:
@frogmouse does the plugin work after installing `mingw-w64-x86_64-python-numpy` as suggested by @fidthorn2vc ?

frogmouse:
I'm not certain what this would do, but I haven't tried it -- I have no desire to install msys2, mingw, or cygwin on this box.  I did check out numpy and xdot in my Win10 python3 and got this result:

C:\Users\frogmouse>python
Python 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> import xdot
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\frogmouse\AppData\Roaming\Python\Python38\site-packages\xdot\__init__.py", line 26, in <module>
    from . import ui
  File "C:\Users\frogmouse\AppData\Roaming\Python\Python38\site-packages\xdot\ui\__init__.py", line 3, in <module>
    from .window import DotWidget, DotWindow
  File "C:\Users\frogmouseimport \AppData\Roaming\Python\Python38\site-packages\xdot\ui\window.py", line 25, in <module>
    import gi
ModuleNotFoundError: No module named 'gi'
>>>

My workaround is to have a directory shared between Windows and a VBox VM running the linux of my choice.  I can use do most of my editing in the Win Zim and get the link map to work in the linux instantiation.

[...]

PR #2386, opened 2023-06-15
#2012 fix ImportError for xdot

k4nzdroid:
During the building stage, we need to install Python modules for plugins. If not, There is no chance to add python modules anymore.


dehesselle:
Future advice: do not close pull requests and open new ones because something doesn't work. Just fix your code locally and push again, this will update (and fix) the existing pull request.

Issue #2399, opened 2023-06-28
Fuzzy search in the "Jump to" dialog.


k4nzdroid:
see #2398

PR #2400, opened 2023-06-28
#2399 Fuzzy search in the "Jump to" dialog.

k4nzdroid:


jaap-karssenberg:
Thanks for coming up with a patch - quick code review this looks like a good improvement :)

At least needs a few test cases for the newly added methods.

jaap-karssenberg:
Busy making the update myself - will merge today or tuesday

Issue #2405, opened 2023-07-05
[Feature Request] Allow swtiching between HPaned and VPaned in ImageGeneratorDialog.


k4nzdroid:
see #2404

PR #2406, opened 2023-07-05
#2405 Support swtiching layout for ImageGeneratorDialog

k4nzdroid:
Add a [Rotate Layout] button to allow users to change the layout for the Source Editor and the Image Viewer in ImageGeneratorDialog.


k4nzdroid:
Writing test cases is not something I am good at, but I will try it afterwards.

Issue #2424, opened 2023-07-20
Screenshot plugin fails on Wayland on Fedora 38 


oturpe:
**Describe the bug**
Screenshot plugin fails on Wayland on Fedora Workstation 38.
This is GNOME 44.2 desktop environment.
When using an X session, the screenshot is captured as expected.

The following output appears in the debug log:

```
DEBUG: Action: insert_screenshot
DEBUG: Opening dialog "Insert Screenshot"
DEBUG: Dialog response OK
INFO: Spawning: ('import', '-silent', '-window', 'root', '/tmp/zim-ex3ub765/insert-screenshot.png') (cwd: /home/otto)
DEBUG: Process started with PID: 5239
DEBUG: Closed dialog "Insert Screenshot"
import: unable to read X window image 'root': Resource temporarily unavailable @ error/xwindow.c/XImportImage/5065.
import:  `/tmp/zim-ex3ub765/insert-screenshot.png' @ error/import.c/ImportImageCommand/1289.
DEBUG: Running ErrorDialog
DEBUG: Some error occurred while running "import"
NoneType: None
ERROR: Some error occurred while running "import"
```

**To Reproduce**
1. Start Zim.
2. **Edit** → **Preferences** → **Plugins**, enable _Insert Screenshot_
3. **Insert** → **Screenshot...**
4. **OK**
5. Observe dialog: _Some error occurred when running "import"_
6. **Close**
7. **Tools** → **View Debug Log**
8. Observe the log rows given in description. 

**Expected behavior**
Screenshot is taken and added to current cursor position.

**Screenshots**
n/a

**Environment (please complete the following information):**
 - Version 0.75.2
 - OS: Fedora Workstation 38, GNOME 44.2, Wayland.
 - Language: Initially observed with fi-FI, reproduces also with C.UTF-8. I do not think this is language related.

**Additional context**
Usually, when I have encountered screenshot problems with Fedora,
they have been related to the fact that capturing a screenshot on Wayland is only possible through the screenshot portal.
Perhaps the screenshot plugin is not aware of this and is just attempting to do it as it is done on X?

introt:
> Usually, when I have encountered screenshot problems with Fedora,
they have been related to the fact that capturing a screenshot on Wayland is only possible through the screenshot portal.

The plugin source code references `scrot`, `import`, `boxcutter` and `gnome-screenshot`. According to your logs, its trying to use `import` which in turn tries to use X.

[...]

PR #2429, opened 2023-07-22
Add support to take screenshots on macOS

dehesselle:
This is a follow-up to https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/2424

- Introduce `screencapture` to support screen captures on macOS.
  - Using `-T0` for `full` is actually a workaround (it says "delay 0 seconds") as this parameter cannot be set to `None`.
- Use `platform.system()` instead of `os.name` as we need a way to detect macOS ("Darwin").

### DIFF
--- a/data/manual/Plugins/Insert_Screenshot.txt
+++ b/data/manual/Plugins/Insert_Screenshot.txt
@@ -9,7 +9,7 @@ This plugin adds a dialog that allows you to take a screenshot of the desktop or
 
 Supported tools:
 
-Linux / unix:
+Linux / unix (X11 only):
 * import (ImageMagick)
 * scrot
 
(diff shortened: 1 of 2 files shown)


introt (changes requested):
Looks good, thanks!

Please update Plugins:Insert Screenshot to include "macOS: screencapture". If you want to go the extra mile, you may also add `(X11)` to the Linux options as per #2424 they don't work on Wayland

dehesselle:
As long as it stays a separate commit I'm not opposed to letting it piggyback on this PR.

introt (approved):
Thanks, LGTM!

introt:
...after  #2431?

PR #2439, opened 2023-07-30
Port from GtkSpell to Gspell

a-andre:


jaap-karssenberg:
Thanks for the patch. I'm a bit puzzled why you choose to remove the gtkspell support completely. The way the module is written with multiple "adaptors" would easily allow keeping gtkspell support as a fallback when loading GSpell fails. Have you considered that option?

a-andre:
I haven't thought about keeping gtkspell and adding Gspell as an additional option.

A lot of tools have switched from gtkspell to Gspell, so it's more likely that the latter is already available on the system. And as fallback there is still gtkspellcheck.

A third option would also mean more maintenance work.

jaap-karssenberg:
Thanks for the effort of figuring out how to make Gspell work. I have reworked the patch to use a separate adapter object and put you as co-author on the patch.

PR #2440, opened 2023-07-30
Simplify Gio, GLib and Gtk imports

a-andre:
All three are required dependencies.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -22,12 +22,9 @@ logger = logging.getLogger('tests')
 from functools import partial
 
 
-try:
-	import gi
-	gi.require_version('Gtk', '3.0')
-	from gi.repository import Gtk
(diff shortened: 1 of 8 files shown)


introt (commented):
Looks good; I'm for merging as long as it doesn't - for whatever reason - break automatic testing. Apparently we need a maintainer with write permission to accept workflow runs now so guess it's up to @jaap-karssenberg to press some button - likely at #2455 before this one

jaap-karssenberg:
Afraid this one is "won't fix" for me. Intention is/was to have at the non-UI classes run also in an environment without GLib / Gtk. E.g. to run a notebook server. Probably this is broken today, as we lack a direct test for it, but won't remove the ``try .. except ...`` blocks to make it easier to revive that function.

PR #2441, opened 2023-07-30
Use reverse DNS name scheme

a-andre:


dehesselle:
I propose that we take the opportunity here to also fix the naming according to spec. Because it has to be `zim-wiki` (hyphen), not `zim_wiki` (underscore).

a-andre:
> I propose that we take the opportunity here to also fix the naming according to spec. Because it has to be `zim-wiki` (hyphen), not `zim_wiki` (underscore).

I'm afraid hyphen won't work. For proper flatpak support underscore has to be used.
> `org.example-site.Foo`
> This ID is not valid according to the DBus specification. You can use org.example_site.Foo instead.

https://docs.flatpak.org/en/latest/conventions.html#application-ids

See also https://github.com/flatpak/flatpak/issues/4974#issuecomment-1169862657.

introt (approved):
> Perhaps D-Bus would ideally have used dashes and not allowed underscores, but it's at least 15 years too late to make that compatibility-breaking change, so underscores are what we have and that will not change.

This seems to indeed be the way if we're going to use the reverse DNS name scheme.

PR #2465, opened 2023-09-03
fix: Typo in zim/notebook/layout.py

Toggoren:


introt (approved):
Nice catch! Seems like this function isn't used anywhere, and with this many eyes I don't think it needs tests right now.

PR #2473, opened 2023-09-08
fix: enable keybindings if find bar is in focus

e3rd:
`self.find_bar` is the third parent of the search button, not the first one

image


introt:
Closes #2491

introt (approved):
Closes #2491

rzwisler:
Yep, this fixes it, thank you!

PR #2481, opened 2023-09-20
Automatically set data dir for a virtualenv install

jhermann:
This also fixes the silent swallowing of the AssertionError that is supposed to report any data dir problem.


jaap-karssenberg:
Only after merging noticed that you generated the PR against the "release" branch and not the "develop" branch - please don't do that again ! - my bad for not checking :(

Issue #2382, opened 2023-06-11
Fwd: [abrt] Zim: _move_file_and_folder(): notebook.py:599:_move_file_and_folder:zim.notebook.notebook.PageExistsError: <exception str() failed>


cheese:
https://bugzilla.redhat.com/show_bug.cgi?id=2213835

Version-Release number of selected component:
Zim-0.75.1-2.fc38

Additional info:
reporter:       libreport-2.17.10
kernel:         6.3.5-200.fc38.x86_64
cmdline:        /usr/bin/python3 /usr/bin/zim
cgroup:         0::/user.slice/user-1000.slice/user/app.slice/app-gnome-zim-269451.scope
uid:            1000
reason:         notebook.py:599:_move_file_and_folder:zim.notebook.notebook.PageExistsError: <exception str() failed>
executable:     /usr/bin/zim
type:           Python3
package:        Zim-0.75.1-2.fc38
runlevel:       N 5
exception_type: zim.notebook.notebook.PageExistsError
crash_function: _move_file_and_folder
interpreter:    python3-3.11.3-2.fc38.x86_64

Truncated backtrace:
notebook.py:599:_move_file_and_folder:zim.notebook.notebook.PageExistsError: <exception str() failed>

Traceback (most recent call last):
```
  File "/usr/lib/python3.11/site-packages/zim/notebook/operations.py", line 243, in <lambda>
    GObject.idle_add(lambda: next(my_iter, False), priority=GObject.PRIORITY_LOW)
                             ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/zim/notebook/operations.py", line 270, in __iter__
    progress = next(self._do_work)
               ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/zim/notebook/notebook.py", line 559, in move_page_iter
    self._move_file_and_folder(path, newpath)
  File "/usr/lib/python3.11/site-packages/zim/notebook/notebook.py", line 599, in _move_file_and_folder
    raise PageExistsError(newpath)
zim.notebook.notebook.PageExistsError: <exception str() failed>
```

Local variables in innermost frame:
self: <Notebook: Notes>
path: <PageIndexRecord: kubernetes>
newpath: <Path: FOSS:kubernetes>
file: <LocalFile: /home/renich/Documents/Notebooks/Notes/kubernetes.txt>
[...]

PR #2482, opened 2023-09-21
Improve error handling for UI actions using `NotebookOperation`

oturpe:
Fixes issue #2382.

## Fix PageExistsError and PageReadOnlyError inheritance

Classes `PageExistsError` and `PageReadOnlyError` were inherited directly from `zim.Error`,
yet their implementations only make sense as subclasses of `PageError`. Make it so.

Previously, the error message for these two errors was just the page name.
After this fix, the page name is prefixed with _Page already exists_ and Can not modify page respectively.
This makes the error messages much more understandable.

## Propagate exceptions from progress dialogs

`NotebookOperation` performs its work inside the GTK main loop.
Any unhandled exceptions occuring within the main loop are processed within the main loop
and do not propagate to code that started operation.
To allow informing the calling side about exceptions, `NotebookOperation` has attribute exception,
which together with the cancelled is filled by `NotebookOperation`'s top level exception handler.

However, for many UI actions that use NotebookOperation wrapped in a `ProgressDialog`,
the exception attribute is not considered at all.
For these dialogs, the user does not see the expected error dialog should an exception occur.

Fix by re-raising any exception reported by NotebookOperation in the following actions:

* MovePageDialog
* ExportDialog
* check_and_update_index, which has its own menu item, and is also  started in multiple other UI actions

## Avoid unhandled exceptions in `NotebookOperation`

`NotebookOperation` is intended to be executed in the GTK main loop via `run_on_idle()`.
Even if such execution terminates in an unhandled exception, execution of Zim still continues.
Thus, normally the effect of an unhandled exception is to print an error message and a traceback in Zim's debug log.

However, if `sys.excepthook` is overridden, different action may result.
In particular, in Fedora Linux, Automatic Bug Reporting Tool (ABRT)'s Python plugin overrides `sys.excepthook`
to notify the user about unhandled exception in Zim,
and also to upload a crash report to Fedora Analytics Framework.
Since multiple `NotebookOperation` subclasses use exceptions to communicate failures that are preventable by the user,
such as trying to create a page that already exists,
this is not desirable.

To avoid triggering `sys.excepthook`,
stop re-raising exceptions in the already existing `NotebookOperation` catch-all exception handler.
Just print them to the debug log instead.

Note that since the catch-all handler already communicates the exceptions to `NotebookOperation` caller via the `cancelled` and `exception` properties,
it could be argued that all handling of exceptions, even logging them, should be left to the caller.
However, there are places where `NotebookOperation` subclasses are executed,
but those properties are not checked afterwards at all,
and thus the *only* error handling they already have is the debug log.
Since the author of this change does not feel comfortable adding exception handling to all those places,
the solution is to keep logging the exception inside `NotebookOperation`.

Since the error logging is now done explicitly via Python's logging
mechanism, tests also need to be modified to expect the new log rows.

## Also some typo fixes in comments

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -376,7 +376,7 @@ class LoggingFilter(logging.Filter):
 
 	# Due to how the "logging" module works, logging channels do inherit
 	# handlers of parents but not filters. Therefore setting a filter
-	# on the "zim" channel will not surpress messages from sub-channels.
+	# on the "zim" channel will not supress messages from sub-channels.
 	# Instead we need to set the filter both on the channel and on
 	# top level handlers to get the desired effect.
 
--- a/tests/uiactions.py
+++ b/tests/uiactions.py
@@ -253,8 +253,9 @@ class TestUIActions(tests.TestCase):
 			dialog.set_input(name='ExistingPage')
 			self.assertRaises(PageExistsError, dialog.do_response_ok)
 
-		with tests.DialogContext(renamepage):
-			self.uiactions.move_page()
+		with tests.LoggingFilter('zim', 'Page already exists'):
+			with tests.DialogContext(renamepage):
+				self.uiactions.move_page()
 
 	def testRenamePageNonExistingPageFails(self):
 		from zim.notebook import PageNotFoundError
@@ -264,8 +265,9 @@ class TestUIActions(tests.TestCase):
 			dialog.set_input(name='NewName')
 			self.assertRaises(PageNotFoundError, dialog.do_response_ok)
 
-		with tests.DialogContext(renamepage):
-			self.uiactions.move_page(page)
+		with tests.LoggingFilter('zim', 'No such page'):
+			with tests.DialogContext(renamepage):
+				self.uiactions.move_page(page)
 
 	def testRenamePageWithPageUpdateHeading(self):
 		page = self.notebook.get_page(Path('MyPage'))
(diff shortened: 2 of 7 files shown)


oturpe:
I have updated content and description of this pull request.

introt (approved):
Looks good to me, thanks!

oturpe:
Since I realized that for tests, there is a serious conflict with this merge request and #2490, I decided to close the other one and merge its content here. The merged content is in a separate commit. Sorry for the noise.

jaap-karssenberg:
I agree that the error handling in the operations makes sense this way.

Want to merge this path, but feel the refactoring of the location of the PageNotFoundError(s) is a bit of noise - refactor of PageError itself makes sense, but the NotFound errors only make sense in the notebook context and therefore should belong in the notebook module.

oturpe:
Thank you for the review. I'll revert the move of  `PageError` subclasses from `notebook.py` to `page.py`.

oturpe:
@jaap-karssenberg I have applied the simplification you proposed. Please review again. CI error messages are about GSpell, so I do not think they are about this pull request. When I ran the test suite locally, it passed (with 21 skipped tests).

jaap-karssenberg:
Hi @oturpe ,

Thanks for the update!  I've merged your commits in the develop branch.

Regards,

Jaap

PR #2489, opened 2023-10-11
Add full-text indexing and search functionality to zim search

maclomhair:
This is an attempt at providing fast full-text search by leveraging the sqlite module FTS5, as mentioned on #64 

While the indexing and search itself is implemented in a plugin, modifications to the core zim code are made to include an extra column and to make use of the additional search functionality if enabled.

Tests and a help page for the plugin are included as well.

Also, currently another minor change not directly related to the indexed FTS functionality is included: the command line search option now also show the ranking.

My initial tests have shown over 95% speed increase in searching. At the same time, the index size of a typical 1500-page, 500 words each notebook increased from 1.7 MiB to around 5.5 MiB, which I think is still acceptable.

### DIFF
--- a/data/manual/Plugins/Indexed_Full_Text_Search.txt
+++ b/data/manual/Plugins/Indexed_Full_Text_Search.txt
@@ -1,0 +1,9 @@
+Content-Type: text/x-zim-wiki
+Wiki-Format: zim 0.6
+Creation-Date: 2023-10-11T16:52:26+02:00
+
+====== Indexed Full Text Search ======
+
+This plugin allows to massively speed up the search in page contents, by up to 95%. Simply enable the plugin. The index will be automatically recreated, after which the faster full-text search will be available.
+
+This is achieved by caching a reverse index of tokens (i.e. words) in the index using ''sqlite''. For that to work, it requires a version of ''sqlite'' with the FTS5 extension. This new index will be around 5x bigger than before (for a 1500-page notebook with around 500 words per page, the index grows from 1.7 MiB to about 5.5 MiB).
--- a/tests/indexed_fts.py
+++ b/tests/indexed_fts.py
@@ -1,0 +1,23 @@
+# coding=utf-8
+
+import tests
+
+from zim.plugins import PluginManager
+
+class TestIndexedFTS(tests.TestCase):
+
+	def testIndexing(self):
+		'''Check indexing of Indexed Full Text Search plugin'''
+		plugin = PluginManager.load_plugin('indexed_fts')
+		notebook = self.setUpNotebook(content=tests.FULL_NOTEBOOK)
+
+		notebook.index.check_and_update()
+
+		# One page will never be indexed: the root page
+		self.assertNotEqual(
+			notebook.index._db.execute(
+				"SELECT count(*) FROM pages WHERE fts_id IS NOT NULL;"
+			).fetchone()[0], 0
+		)
+
+
--- a/tests/indexed_fts.py
+++ b/tests/indexed_fts.py
@@ -3,7 +3,12 @@
 import tests
 
 from zim.plugins import PluginManager
+from zim.plugins import indexed_fts
 
+@tests.skipIf(
+	indexed_fts.IndexedFTSPlugin.check_dependencies()[0] == False,
+	"Indexed FTS plugin not available"
+)
 class TestIndexedFTS(tests.TestCase):
 
 	def testIndexing(self):
@@ -13,7 +18,8 @@ class TestIndexedFTS(tests.TestCase):
 
 		notebook.index.check_and_update()
 
-		# One page will never be indexed: the root page
+		# Only pages with content will actually be indexed, so
+		# after indexing some should be FTS-indexed.
 		self.assertNotEqual(
 			notebook.index._db.execute(
 				"SELECT count(*) FROM pages WHERE fts_id IS NOT NULL;"
--- a/zim/main/__init__.py
+++ b/zim/main/__init__.py
@@ -593,9 +593,17 @@ class SearchCommand(NotebookCommand):
 
 		selection = SearchSelection(notebook)
 		selection.search(query)
-		for path in sorted(selection, key=lambda p: p.name):
-			print(path.name)
 
+		if self.opts.get("with-scores", False):
+			sorted_sel = sorted(selection.scores.items(),
+				key=lambda i:i[0].name, reverse=False)
+			sorted_sel.sort(key=lambda i:i[1], reverse=True)
+
+			for result in sorted_sel:
+				print(str(result[1]) + "\t" + result[0].name)
+		else:
+			for path in sorted(selection, key=lambda p: p.name):
+				print(path.name)
 
 class IndexCommand(NotebookCommand):
 	'''Class implementing the C{--index} command'''
--- a/data/manual/Plugins.txt
+++ b/data/manual/Plugins.txt
@@ -1,5 +1,5 @@
 Content-Type: text/x-zim-wiki
-Wiki-Format: zim 0.4
+Wiki-Format: zim 0.6
 
 ====== Plugins ======
 
@@ -16,6 +16,7 @@ The following plugins are standard included with zim:
 * [[+Equation Editor|Equation Editor]]
 * [[+GNU R Plot Editor|GNU R Plot Editor]]
 * [[+Gnuplot Editor|Gnuplot Editor]]
+* [[+Indexed Full Text Search|Indexed Full Text Search]]
 * [[+Insert Screenshot|Insert Screenshot]]
 * [[+Insert Symbol|Insert Symbol]]
 * [[+Inline Calculator|Inline Calculator]]
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -64,7 +64,8 @@ __all__ = [
 	'tasklist', 'tags', 'imagegenerators', 'tableofcontents',
 	'quicknote', 'attachmentbrowser', 'insertsymbol',
 	'sourceview', 'tableeditor', 'bookmarksbar', 'spell',
-	'arithmetic', 'linesorter', 'commandpalette', 'windowtitleeditor'
+	'arithmetic', 'linesorter', 'commandpalette', 'windowtitleeditor',
+	'indexed_fts'
 ]
 
 
--- a/tests/search.py
+++ b/tests/search.py
@@ -5,6 +5,7 @@ import tests
 
 from zim.search import *
 from zim.notebook import Path
+from zim.plugins import indexed_fts
 
 class TestSearchRegex(tests.TestCase):
 
@@ -243,6 +244,20 @@ class TestSearchFiles(TestSearch):
 
 	def runTest(self):
 		'''Test search API with file based notebook'''
+		TestSearch.runTest(self)
+
+@tests.skipIf(
+	indexed_fts.IndexedFTSPlugin.check_dependencies()[0] == False,
+	"Indexed FTS plugin not available"
+)
+class TestSearchIndexed(TestSearch):
+	'''Test case for integration with the indexed_fts plugin'''
+
+	def setUp(self):
+		plugin = PluginManager.load_plugin('indexed_fts')
+		TestSearch.setUp(self)
+
+	def runTest(self):
 		TestSearch.runTest(self)
 
 
--- a/zim/gui/searchdialog.py
+++ b/zim/gui/searchdialog.py
@@ -176,8 +176,8 @@ class SearchResultsTreeView(BrowserTreeView):
 				column.set_expand(True)
 			self.append_column(column)
 
-		model.set_sort_column_id(1, Gtk.SortType.DESCENDING)
-			# By default sort by score
+		# Don't sort here because we'll do more elaborate sorting later manually#
+		#model.set_sort_column_id(1, Gtk.SortType.DESCENDING)
 
 		self.connect('row-activated', self._do_open_page)
 		self.connect('destroy', self.__class__._cancel)
@@ -225,7 +225,7 @@ class SearchResultsTreeView(BrowserTreeView):
 			else:
 				score = -1 # went missing !??? - technically a bug
 			row[self.SCORE_COL] = score
-			order.append((score, i))
+			order.append((path, i, score))
 			seen.add(path)
 
 		# Add new paths
@@ -234,11 +234,12 @@ class SearchResultsTreeView(BrowserTreeView):
 			score = results.scores.get(path, 0)
 			model.append((path.name, score, path))
 			i += 1
-			order.append((score, i))
+			order.append((path, i, score))
 
-		# re-order
-		#order.sort() # sort on first item, which is score
-		#model.reorder([x[1] for x in order]) # use second item
+		# sort by score, then by name. This doesn't seem to work by setting a sort column.
+		order.sort(key=lambda i: i[0].name)
+		order.sort(key=lambda i: i[2], reverse=True)
+		model.reorder([x[1] for x in order])
 
 		self.hasresults = len(model) > 0
 
--- a/zim/notebook/index/pages.py
+++ b/zim/notebook/index/pages.py
@@ -138,7 +138,7 @@ class PagesIndexer(IndexerBase):
 			file = self.layout.root.file(filerow['path'])
 			format = self.layout.get_format(file)
 			mtime = file.mtime()
-			tree = format.Parser().parse(file.read())
+			tree = format.Parser().parse(file.read(), file_input=True)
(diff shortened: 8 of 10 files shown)


maclomhair:
I'd be grateful for any comments on this, in particular on how to do things "properly" so they fit nicely into zim's existing codebase! :)

maclomhair:
It seems that some functionality of sqlite3 I used here is only available in their almost newest (3.43) release and that's why the checks are failing...

data/manual/Help/Config_Files.txt
```diff
@@ -235,3 +235,5 @@ One hidden property is "''end_of_line''" which determines the end-of-line conven
 A second hidden option is "''disable_trash''" which defaults to ''False''. If enabled this will cause zim to avoid using the system trash for this notebook, see the section about deleting in [[Pages]].
 
 A third hidden option is "''shared" which defaults to True'' (however will be ''False'' for older notebooks). If ''True'', zim stores the notebook cache in the home folder (''$XDG_CACHE_HOME''). Usually, this is what you want, however in particular cases, e.g. when the notebook resides in an encrypted folder, you may want to set this to ''False'' in order to make sure the cache is actually stored in the notebook folder (using a ".zim" hidden folder).
+
+A fourth hidden option is "''fts_index''", defaulting to False. If True, zim will use the index to store a "reverse index" for fast full-text search of the page content. Therefore, the index will increase in size but full-text search will be significantly faster. If this option is changed, the index needs to be rebuilt from scratch.
```
introt: "the index **will be** rebuilt" suffices, as that is handled automatically

data/manual/Plugins/Indexed_Full_Text_Search.txt
```diff
@@ -0,0 +1,9 @@
+Content-Type: text/x-zim-wiki
+Wiki-Format: zim 0.6
+Creation-Date: 2023-10-11T16:52:26+02:00
+
+====== Indexed Full Text Search ======
+
+This plugin allows to massively speed up the search in page contents, by up to 95%. Simply enable the plugin. The index should be automatically recreated, after which the faster full-text search should be available.
```
introt: `s/should/will/g` :)

tests/indexed_fts.py
```diff
@@ -0,0 +1,23 @@
...
+
+	def testIndexing(self):
+		'''Check indexing of Indexed Full Text Search plugin'''
+		plugin = PluginManager.load_plugin('indexed_fts')
+		notebook = self.setUpNotebook(content=tests.FULL_NOTEBOOK)
+
+		notebook.index.check_and_update()
+
+		# One page will never be indexed: the root page
+		self.assertNotEqual(
```
introt, line 23: Is it always one page? If so, assertNotEqual(..., 0) -> assertEqual(..., 1)?

data/manual/Help/Config_Files.txt
introt: Is this hidden option needed, if the user can enable the plugin via either the GUI or modifying the config file listing enabled plugins?

data/manual/Plugins/Indexed_Full_Text_Search.txt
```diff
@@ -0,0 +1,9 @@
...
+
+This is achieved by caching a reverse index of tokens (i.e. words) in the index using ''sqlite''. For that to work, it requires a version of ''sqlite'' with the FTS5 extension. This new index will be around 5x bigger than before (for a 1500-page notebook with around 500 words per page, the index grows from 1.7 MiB to about 5.5 MiB).
```
introt, line 9: If the hidden option is kept, it should be documented here as well.

zim/main/__init__.py
```diff
@@ -552,8 +552,11 @@ def run(self):
 
 		selection = SearchSelection(notebook)
 		selection.search(query)
-		for path in sorted(selection, key=lambda p: p.name):
-			print(path.name)
+
+		for result in sorted(selection.scores.items(),
+			key=lambda i:i[1], reverse=True):
+
+			print(str(result[1]) + "\t" + result[0].name)
```
introt: > Also, currently another minor change not directly related to the indexed FTS functionality is included: the command line search option now also show the ranking.

Blocker: This may break scripts and shouldn't be enabled by default

introt (changes requested):
Nice work! The building of the index doesn't seem to take too much longer, and the search results show up instantly, even when searching for single letters - after dismissing the following error:

```
zim-desktop-wiki/zim/plugins/indexed_fts.py", line 68, in process_index_fts
    query_results = db.execute(
OperationalError: no such table: pages_fts
```

Besides that, the main blockers are changing default CLI behavior (I'm okay with showing the scores when given a flag, eg. `--show-scores`) and breaking the test suite.

I've commented some nitpicky details, but one more thing I'd like to see would be to have the pages with the same score sort alphabetically in the GUI.

maclomhair:
Thank you for reviewing and your helpful tips! I'm working on incorporating them in the next few days. Regarding the process of adding these changes, would you rather have me rewrite the commits and force-push to the pull-request, or should I instead append commits fixing the misbehaviour of previous commits (or something different altogether)?

> ```
> zim-desktop-wiki/zim/plugins/indexed_fts.py", line 68, in process_index_fts
>     query_results = db.execute(
> OperationalError: no such table: pages_fts
> ```
I'm still in the process of understanding how this error might have happened. The table should be created after reindexing, and reindexing should happen upon plugin load, so in my understanding, the only possible way this could have happened was if no reindexing actually happened, or if the sqlite version was too old...

Thanks again for your help!

data/manual/Help/Config_Files.txt
maclomhair: This commit actually sneaked in from my first approach at the problem where I thought a plugin wouldn't do... the current patchset doesn't even include any checks on this, so it has no effect.

TLDR: no, I'll remove this hidden option!

tests/indexed_fts.py
maclomhair: No, unfortunately it seems it's not that easy. I think the indexing happens only for pages with content, so some are indexed, but empty ones and placeholders are not. I'll clarify the comment!

zim/main/__init__.py
maclomhair: I now added the option "--with-scores" to the search and default is as before :)

maclomhair:
I think I implemented  your other suggestions but this last one looks a bit more difficult to me:

> [...] but one more thing I'd like to see would be to have the pages with the same score sort alphabetically in the GUI.

My impression was that per [zim/gui/searchdialog.py, Line 179](https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/c88cf3cb53896bf272e87704826b77e82eddb3ef/zim/gui/searchdialog.py#L179C18-L179C18) the sorting is handled by the TreeView widget. Because of this, I am unsure on how to properly have it sorted first by index then alphabetically, while considering that the user might re-sort it by clicking on the column headers.

introt:
> Regarding the process of adding these changes, would you rather have me rewrite the commits and force-push to the pull-request, or should I instead append commits fixing the misbehaviour of previous commits (or something different altogether)?

I don't think we have guidelines, I'm okay with appending new commits until everything is otherwise ready to be merged.

> I'm still in the process of understanding how this error might have happened. The table should be created after reindexing, and reindexing should happen upon plugin load, so in my understanding, the only possible way this could have happened was if no reindexing actually happened, or if the sqlite version was too old...

Yeah, it's weird - especially because it displays results instantly afterwards anyway :D 

> My impression was that per [zim/gui/searchdialog.py, Line 179](https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/c88cf3cb53896bf272e87704826b77e82eddb3ef/zim/gui/searchdialog.py#L179C18-L179C18) the sorting is handled by the TreeView widget. Because of this, I am unsure on how to properly have it sorted first by index then alphabetically, while considering that the user might re-sort it by clicking on the column headers.

I'm not very good at Gtk stuff, sorry. I guess you could add a hidden column and fill that with the custom ordering, but that feels quite hackish.

I'll try to review the new changes next weekend :+1:

introt:
Nice work! It seems my distro doesn't satisfy the new version requirement, but the code changes look good. While testing the cli flag, it seems that if the scores aren't shown the pages are sorted alphabetically, but with the scores shown the pages with the same score aren't sorted. I'm down to getting this merged as-is, the version check should take care of the previous error - though we do need to bump the version in the flatpak as well.

maclomhair:
Thank you for reviewing it once more!

Regarding the error that occured when testing the first draft, I think it had to do with sqlite's version containing the FTS5 module but being a version below 3.43 which doesn't yet contain the "contentless_delete" option that we need. That way, the table creation failed, triggering the exception in the plugin and possibly deactivating the plugin right away (although I'm not sure zim handles it that way but it would make sense). With the plugin deactivated, the search fell back to the traditional search functionality, yiedling the same results but slower (but maybe it was fast in the test because the entire notebook had been reindexed shortly before so all the files where already cached in memory and going through them was therefore fast?). So at least that's my theory of what might have happened.

Regarding the CLI flag to also print scores I thought it makes sense to order by score, but if you'd rather have it sorted alphabetically I don't mind! Do you want me to make that change?

I have no experience with flatpaks, but to me it seems that the newest runtime still doesn't ship sqlite version 3.43... The gnome runtime version 45 [is based on](https://gitlab.gnome.org/GNOME/gnome-build-meta/-/blob/bb5323f7cddf3394de9586ff1750d1ea23243dc0/elements/freedesktop-sdk.bst#L7) freedesktop-sdk version 23.08, which in turn [still uses](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/4b0d53c0da2400382d8a4764f063177db200254d/elements/components/sqlite.bst#L39) sqlite version 3.42. I don't know how to get a more recent version in there...

introt:
No problem!

> but maybe it was fast in the test because the entire notebook had been reindexed shortly before so all the files where already cached in memory and going through them was therefore fast?

The bottleneck shouldn't be the speed of my SSD, and besides if relevant caching took place, the effect would be observed without the patch during subsequent searches, so I'd be interested to have a further look into this because it could mean there's a way to speed up the "traditional" search as well.

> Regarding the CLI flag to also print scores I thought it makes sense to order by score, but if you'd rather have it sorted alphabetically I don't mind! Do you want me to make that change?

I'd love the [complex sort](https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts) if you don't mind. Regarding the tree view, what I was suggesting seems to be called the [Decorate-Sort-Undecorate](https://docs.python.org/3/howto/sorting.html#decorate-sort-undecorate) idiom.

> I have no experience with flatpaks, but to me it seems that the newest runtime still doesn't ship sqlite version 3.43... The gnome runtime version 45 [is based on](https://gitlab.gnome.org/GNOME/gnome-build-meta/-/blob/bb5323f7cddf3394de9586ff1750d1ea23243dc0/elements/freedesktop-sdk.bst#L7) freedesktop-sdk version 23.08, which in turn [still uses](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/4b0d53c0da2400382d8a4764f063177db200254d/elements/components/sqlite.bst#L39) sqlite version 3.42. I don't know how to get a more recent version in there...

We could stick a newer sqlite in there manually, but imo waiting it out is also fine.

maclomhair:
Thank you for the links to python's sort documentation. That helped a lot :) I now added the desired ordering for the CLI search output, and it turned out that the same could be applied to the GUI search results so both should now always be sorted score first, name second!

maclomhair:
I just noticed that the full-text indexing function had a minor flaw: words with markup were split at their markup. I just now pushed a one-line change to fix this so words are indexed as a whole even with markup in the middle. But it's a complete riddle to me that the very same test suite now fails...

introt:
The macos fail is unrelated, seems the test suite is broken: `ModuleNotFoundError: No module named 'gi'`.

The windows fail is also from another part of the test suite:

```
2023-11-05T07:58:09.5481949Z testIndexing (tests.indexed_fts.TestIndexedFTS.testIndexing) ... ok
2023-11-05T07:58:09.5484281Z 
2023-11-05T07:58:09.5484539Z ======================================================================
2023-11-05T07:58:09.5485231Z FAIL: runTest (tests.widgets.TestFileDialog.runTest)
2023-11-05T07:58:09.5486211Z ----------------------------------------------------------------------
2023-11-05T07:58:09.5486617Z Traceback (most recent call last):
2023-11-05T07:58:09.5487237Z   File "D:/a/zim-desktop-wiki/zim-desktop-wiki/tests/widgets.py", line 414, in runTest
2023-11-05T07:58:09.5487924Z     self.assertIsInstance(adapt_from_oldfs(myfolder), LocalFolder)
2023-11-05T07:58:09.5488651Z AssertionError: None is not an instance of <class 'zim.newfs.local.LocalFolder'>
2023-11-05T07:58:09.5489085Z 
2023-11-05T07:58:09.5489300Z ----------------------------------------------------------------------
2023-11-05T07:58:09.5489677Z Ran 997 tests in 47.121s
2023-11-05T07:58:09.5489852Z 
2023-11-05T07:58:09.5490025Z FAILED (failures=1, skipped=38, expected failures=21)
```

Let's see if these are present during other runs, and if not, retrigger the suite.

maclomhair:
Ok it seems the test suite ran again successfully yesterday in pull request #2511 but AFAIU [only people with write permissions on a repo can re-run the test suite](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs) - @introt do you mind triggering the test suite for this pull request again? Thanks!

introt:
No `+w` here, sorry! You can amend your latest commit and force push or wait for @jaap-karssenberg

maclomhair:
Ok thanks for the info! It seems like the current develop git head also doesn't successfully pass the test suite so I think I'll wait with a force push until I see the test suite passing again... (or I could somehow try debugging the macos test job but I have zero experience in this and don't even know where to start)

dehesselle:
@maclomhair I think this is/was some weird outside-influence. I've just pushed the current `develop` branch to my fork and the test suite runs fine on macOS. Windows still broken, but with an actual test failure and not a "failed to import xyz" oddity.

maclomhair:
Thanks for the hint! Huh interesting and good to know! If I understand correctly, the test suite ran on your repo on the same commit that's the tip here (db81ebde65529ae79ae4a6713f3b2e5e20bc6c58) and it ran through the macOS test suite but failed on the windows test suite. The same commit ran through (the same?) test suite in this repo and there the macOS test suite had a glitch but the windows test suite ran through successfully... oof that seems confusing to me.

dehesselle:
Yes, exactly. You can see the run here https://github.com/dehesselle/zim-desktop-wiki/actions/runs/7011010243

I got the idea to test that because there is new PR here in this repo where the test suite did not fail, so it's not something that broke in the `develop` branch.

Without doing too much digging, the runs used different image versions. We only request `macos-12` in the workflow, but the exact image version (GitHub frequently releases updates) is not specified and does change.

jaap-karssenberg:
Thanks for the PR - it looks very complete already.

A few attention points that need to be addressed before merging:
- I see a new field "fts_id" in the page index table, if possible I would like to avoid having plugin specific fields in a core table. Is it possible to do this linking in a separate table or as part of the fts table ?
- Plugin hook is now hard coded into the search processing code - let me think how to make that a proper extendable where plugins can add there handlers
- The change to add more commandline options could be a separate PR to keep things clean - but this is minor for me

maclomhair:
Thanks for taking a look!

>     * I see a new field "fts_id" in the page index table, if possible I would like to avoid having plugin specific fields in a core table. Is it possible to do this linking in a separate table or as part of the fts table ?

Yes, doing this linking in a small separate table should be no problem! The fts table cannot be used for that, though, because it is defined by the fts5 module of sqlite.
 
>     * The change to add more commandline options could be a separate PR to keep things clean - but this is minor for me

That's also no problem, will do!

maclomhair:
>     * I see a new field "fts_id" in the page index table, if possible I would like to avoid having plugin specific fields in a core table. Is it possible to do this linking in a separate table or as part of the fts table ?

I've now reversed the additional fields in the table and added a separate table to keep track of page id to full-text-search document id - no core tables should now be modified by the plugin! :)

I'll try to rebase the changes onto the newest develop branch next, and also combine changes into meaningful commits in the process.

maclomhair:
The only remaining point is now the not-so-clean implementation of the search hook directly looking for the plugin. If you have a broad idea on how that could look cleaner and where I could get started, I might be able to take a shot at that, too (in a while probably, though) :)

jaap-karssenberg:
Thanks!  For me it looks ready to merge as-is.  The search hook is isolated enough that I can refactor that myself. Need to do bit more thinking how to extend the search functionality in a flexible way.

Jaap

PR #2490, opened 2023-10-12
Avoid unhandled exceptions in `NotebookOperation`

### LINKED ISSUE #2382: Fwd: [abrt] Zim: _move_file_and_folder(): notebook.py:599:_move_file_and_folder:zim.notebook.notebook.PageExistsError: <exception str() failed>

oturpe:
`NotebookOperation` is intended to be executed in the GTK main loop via `run_on_idle()`.
Even if such execution terminates in an unhandled exception, execution of Zim still continues.
Thus, normally the effect of an unhandled exception is to print an error message and a traceback in Zim's debug log.

However, if `sys.excepthook` is overridden, different action may result.
In particular, in Fedora Linux, Automatic Bug Reporting Tool (ABRT)'s Python plugin overrides `sys.excepthook`
to notify the user about unhandled exception in Zim,
and also to upload a crash report to Fedora Analytics Framework.
Since multiple `NotebookOperation` subclasses use exceptions to communicate failures that are preventable by the user,
such as trying to create a page that already exists,
this is not desirable.

To avoid triggering `sys.excepthook`,
stop re-raising exceptions in the already existing `NotebookOperation` catch-all exception handler.
Just print them to the debug log instead.

Note that since the catch-all handler already communicates the exceptions to `NotebookOperation` caller via the `cancelled` and `exception` properties,
it could be argued that all handling of exceptions, even logging them, should be left to the caller.
However, there are places where `NotebookOperation` subclasses are executed,
but those properties are not checked afterwards at all,
and thus the *only* error handling they already have is the debug log.
Since the author of this change does not feel comfortable adding exception handling to all those places,
the solution is to keep logging the exception inside `NotebookOperation`.

Fixes #2382 
Relates to #2482

### DIFF
--- a/zim/notebook/operations.py
+++ b/zim/notebook/operations.py
@@ -144,7 +144,7 @@ except ImportError:
 
 
 from zim.signals import SignalEmitter
-from zim.errors import Error
+from zim.errors import Error, log_error
 
 NOOP = lambda: None
 
@@ -279,7 +279,7 @@ class NotebookOperation(SignalEmitter):
 		except Exception as err:
 			self.cancelled = True
 			self.exception = err
-			raise
+			log_error(err)
 		finally:
 			if self.notebook._operation_check == self:
 				self.notebook._operation_check = NOOP # stop blocking


oturpe:
It looks like `TestLoggingHandler` in the test suite failed some related tests because errors were logged to the debug log.
This is foreign territory for me, but I will take a look when I find some time.

oturpe:
So, the test suite checks that no warnings or errors are logged during tests.

I did research some "all handling of exceptions, even logging, is left to the caller" avenue. I _think_ #2482 fixes all UI action cases. But there is also at least one background case, `IndexUpdateOperation` here, which I do not know how to fix properly:
https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/c88cf3cb53896bf272e87704826b77e82eddb3ef/zim/notebook/index/__init__.py#L419
(Not that that code currently works properly: In case of certain errors, it will ceaselessly retrigger the update operation, fully saturating a CPU core and spamming the debug log — but that is a different story.)

So, I will look into making it possible to say "this test case is expected to log N log rows that are warning or worse". That should not be very complicated.

oturpe:
I turns out that there is already a suitable mechanism, `LoggingFilter`. But now that I got to modifying the tests, I realized that while otherwise separate, this pull request and #2482 are going conflict there. Instead of trying to juggle both, I will instead just close this and add avoiding of unhandled exceptions to #2482.

introt:
> (Not that that code currently works properly: In case of certain errors, it will ceaselessly retrigger the update operation, fully saturating a CPU core and spamming the debug log — but that is a different story.)

Ah, I've been meaning to look into this - thanks for tracking that down! Let's open an issue for it

PR #2495, opened 2023-10-22
Fix #2457: The ``Insert Date and Time`` dialog box does not correctly process Unicode strings from the ``dates.list`` file

### LINKED ISSUE #2457: The ``Insert Date and Time`` dialog box does not correctly process Unicode strings from the ``dates.list`` file
**Describe the bug**
The ``Insert Date and Time`` dialog box does not correctly process Unicode strings from the ``dates.list`` file

**To Reproduce**
1. Add the following string ``%Y-%m-%d ❗`` to the ``dates.list`` file
2. Open ``Insert Date and Time`` dialog
3. The date is displayed as ``INVALID: %Y-%m-%d ❗``

**Expected behavior**
Date should be displayed like ``2023-08-24 ❗``

**Screenshots**
n/a

 - Version: Any
 - OS: Windows
 - Language: Any

**Additional context**

Tech:
``strftime()`` does not always handle Unicode strings correctly. 
Also, processing differs for different locales and OS.

Social🤷:
In 🌍modern 🆕world, it seems🧐 😱impossible to exist💯 with⚪ut actively using emoji.   
Zim wiki must keep up with the times.💪🏻

Minimal reproducible example:
```python
import datetime; datetime.datetime.now().strftime('%Y ❗')
```
Gives
```console
User@DESKTOP MINGW64 ~
$ python
Python 3.11.5 (main, Aug 27 2023, 19:14:12)  [GCC 13.2.0 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime; datetime.datetime.now().strftime('%Y ❗')
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'locale' codec can't encode character '\u2757' in position 3: encoding error

```

Toggoren:
Fixes: #2457

### DIFF
--- a/zim/datetimetz.py
+++ b/zim/datetimetz.py
@@ -230,7 +230,12 @@ def strfcal(format, date):
 
 def strftime(format, date):
 	# TODO: deprecate this function
-	return date.strftime(format)
+	# see issue #2457
+	# When we pass some unicode characters (e.g. emoji)
+	# to strftime under Windows we get a UnicodeEncodeError exception.
+	# To avoid this, we convert all non-ASCII characters to their \uXXXXXX representations,
+	# then pass to the strftime function and convert back to a Unicode string.
+	return date.strftime(format.encode('unicode-escape').decode()).encode().decode('unicode-escape')
 
 
 if __name__ == '__main__': #pragma: no cover
--- a/tests/datetimetz.py
+++ b/tests/datetimetz.py
@@ -50,6 +50,18 @@ class TestDateTimeZ(tests.TestCase):
 		#s = datetime.strftime('%V', dt)
 		#self.assertTrue(isinstance(s, str) and len(s) > 0)
 
+		s = datetime.strftime('%Y 道', dt)
+		self.assertTrue(isinstance(s, str) and len(s) == 4 + 1 + 1)
+
+		s = datetime.strftime('%Y ❗', dt)
+		self.assertTrue(isinstance(s, str) and len(s) == 4 + 1 + 1)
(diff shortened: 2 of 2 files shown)


zim/datetimetz.py
```diff
@@ -230,7 +230,7 @@ def replacefunc(matchobj):
 
 def strftime(format, date):
 	# TODO: deprecate this function
-	return date.strftime(format)
+	return date.strftime(format.encode('unicode-escape').decode()).encode().decode('unicode-escape')
```
introt, line 238: Could you add an inline comment for this? The tests pass on my Linux system without this change, so given the above TODO about deprecation someone further down the line might re-break this if the purpose of this encode-decode chain isn't explained. The current commit message, "avoid strftime failed when accepted unicode string under Windows", would imo suffice, with some more context (source of solution: the internet vs. manual testing?) in the commit message.

introt (commented):
The comment is good, thanks. Those tests should trigger the UnicodeEncodeError, but is there a reason the last two might have a greater-than number of characters?

Toggoren:
> The comment is good, thanks. Those tests should trigger the UnicodeEncodeError, but is there a reason the last two might have a greater-than number of characters?

Previously, there were some problems with string length calculation.  
https://stackoverflow.com/questions/29109944  
https://stackoverflow.com/questions/30775689  

I didn't bother to find out the relevance of this problem and just counted graphemes, their number is bound to be lesser or equal than to the number of code points.

PR #2497, opened 2023-10-23
fix: Make test suite pass under non "en_US" locales

### LINKED ISSUE #2492: Test ``TestMenuDocs.testAllMenuItemsAreDocumented`` is falling where it shouldn't.
**Describe the bug**
Test ``TestMenuDocs.testAllMenuItemsAreDocumented`` is falling where it shouldn't. 🫠

**To Reproduce**
1. Run ``python test.py tests.mainwindow.TestMenuDocs.testAllMenuItemsAreDocumented`` under Windows 10 with russian locale
2. See error  
    ```bash
    WARNING: foo
    testAllMenuItemsAreDocumented (tests.mainwindow.TestMenuDocs.testAllMenuItemsAreDocumented) ... FAIL
    
    FAIL: testAllMenuItemsAreDocumented (tests.mainwindow.TestMenuDocs.testAllMenuItemsAreDocumented)
      File "blabla/zim-desktop-wiki/tests/mainwindow.py", line 308, in testAllMenuItemsAreDocumented
        self.assertTrue(label in self.manual, 'Menu item "{}" not documented'.format(label))
    AssertionError: False is not true : Menu item "**Back <Alt><Влево>**" not documented
    
    Ran 1 test in 1.137s
    
    FAILED (failures=1)
    
    Wrote test report to test_report.html
    Wrote test_times.csv
    ```

**Expected behavior**
This test should have passed.

**Screenshots**
N/A

 - Version: Any
 - OS: Windows, Linux
 - Language: Any except ``en_US``(?)

**Additional context**
```powershell
PS C:\Users\User> systeminfo /fo csv | ConvertFrom-Csv | select OS*, System* | Format-List


OS Name             : Microsoft Windows 10 Enterprise LTSC
OS Version          : 10.0.19044 N/A Build 19044
OS Manufacturer     : Microsoft Corporation
OS Configuration    : Standalone Workstation
OS Build Type       : Multiprocessor Free
System Manufacturer : Micro-Star International Co., Ltd.
System Model        : MS-7D95
System Type         : x64-based PC
System Directory    : C:\Windows\system32
System Locale       : ru;Russian
```

I looked at the source code for ``Gtk.accelerator_get_label``. 
https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkaccelgroup.c#L1704-1718
https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkaccellabel.c#L773-935
https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/po/ru.po#L170-173
Apparently it needs a hack to turn back localized string into non-localized one.
Or look for matches not in ``manual/Help/Menu_Items.txt``.

Аfter some more research, I found that this behavior apparently affects Linux too.
```sh
user@user-pc:~$ locale
LANG=de_DE.UTF-8
LANGUAGE=de_DE:ru:en_US:en
LC_CTYPE="de_DE.UTF-8"
LC_NUMERIC=de_DE.UTF-8
LC_TIME=de_DE.UTF-8
LC_COLLATE="de_DE.UTF-8"
LC_MONETARY=de_DE.UTF-8
LC_MESSAGES="de_DE.UTF-8"
LC_PAPER=de_DE.UTF-8
LC_NAME=de_DE.UTF-8
LC_ADDRESS=de_DE.UTF-8
LC_TELEPHONE=de_DE.UTF-8
LC_MEASUREMENT=de_DE.UTF-8
LC_IDENTIFICATION=de_DE.UTF-8
LC_ALL=
user@user-pc:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.6 LTS
Release:	18.04
Codename:	bionic
```
```sh
user@user-pc:blabla/zim-desktop-wiki$ python3 test.py tests.mainwindow.TestMenuDocs.testAllMenuItemsAreDocumented
WARNING: foo
testAllMenuItemsAreDocumented (tests.mainwindow.TestMenuDocs) ... FAIL

FAIL: testAllMenuItemsAreDocumented (tests.mainwindow.TestMenuDocs)
  File "blabla/zim-desktop-wiki/tests/mainwindow.py", line 304, in testAllMenuItemsAreDocumented
    self.assertTrue(label in self.manual, 'Menu item "{}" not documented'.format(label))
AssertionError: False is not true : Menu item "===== Datei =====" not documented

Ran 1 test in 0.135s

FAILED (failures=1)

Wrote test report to test_report.html
Wrote test_times.csv
```

Toggoren:
Fixes #2492

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -34,6 +34,7 @@ import unittest
 from unittest import skip, skipIf, skipUnless, expectedFailure
 
 
+os.environ['LANGUAGE'] = 'C.UTF-8'
 gettext.install('zim', names=('_', 'gettext', 'ngettext'))
 
 FAST_TEST = False #: determines whether we skip slow tests or not


introt (changes requested):
The suite should identically on all locales, not skip tests on non-en_US locales - please use `locale.setlocale` to set the locale for the whole test suite to en_US

Toggoren:
> The suite should identically on all locales, not skip tests on non-en_US locales - please use `locale.setlocale` to set the locale for the whole test suite to en_US

The user may not have the ``en_US`` locale installed. What should we do in this case?

dehesselle:
Allow me to chime in here. The common practice is to set locale to `C` if you don't actually test locale-specific things.

Toggoren:
> Allow me to chime in here. The common practice is to set locale to `C` if you don't actually test locale-specific things.

oh I didn't know that. Thank you for explaining

introt (commented):
Thanks for chiming in, @dehesselle!

I get not wanting to import an extra module, but I'd prefer setting the whole locale instead of the language in case eg. the decimal point affects some future arithmetic plugin test. What do you think?

jaap-karssenberg:
@introt @dehesselle my initial reaction was the opposite - would rather "decorate" individual tests with locale settings, so you ensure everything else is robust under locale of the system running the test. Not sure what is the standard practice here. Do you have experience in other projects on this?

introt:
Seems to depend on the project. Imo separating locale-specific tests (do we have any besides rtl stuff?) and running the rest under the same sanitized environment makes most sense - removing moving parts is usually helpful for replicating failures. Perhaps a `--sanitize-environment` flag or similar could be used here?

Issue #310, opened 2018-02-06
feature request: show broken file links in red instead of blue


nbeaver:
Currently, the only way to tell when a file link is broken is to click on it
and receive a "No such file or folder" error message.

It would be useful if a visual indicator, preferably the color of the link,
showed whether or not the file or folder existed.

This should be refreshed after loading a page and after editing or adding a
link.

This cannot be accomplished with a plugin, as per this mailing list reference:

> You will need to patch the core code of the pageview to accomplish this.

> The reason is that the gtk.TextTag is private within the editor widget. The
> parsetree only passes on the content of the page, but indeed does not contain
> the widget objects.

> So to implement this feature the parsetree needs to be updated to contain a
> link attribute whether or not the target exists, plus the widget should be
> updated to understand this attribute and render the link differently.

> — Jaap Karssenberg

https://lists.launchpad.net/zim-wiki/msg03879.html

Original request on February 8, 2015, archived here:

https://web.archive.org/web/20171220085551/https://bugs.launchpad.net/zim/+bug/1419531

Self-Perfection:
I like the idea. Links to missing notes also should be showed red.

mikem23:
Might be nice to unify this with #197 and let users configure style for link:broken or link:external.

Issue #197, opened 2017-08-28
Mark external links


ruedin:
It would be nice if Zim could identify external links with an icon, like [https://commons.wikimedia.org/wiki/File:Icon_External_Link.png](https://commons.wikimedia.org/wiki/File:Icon_External_Link.png) . Currently internal and external links look the same.

robertvazan:
Or more easily, create a new style, say `external-link`, and give it paler blue color by default.

mikem23:
possibly related: #310

Issue #513, opened 2018-08-18
Add tel: skype: and maybe others as external links


morciej:
When trying to add tel:+18005551212 as a link, I would expect to have it open my phone application; instead a new page is created. This would be extremely useful in task lists.

Self-Perfection:
That was already briefly discussed https://bugs.launchpad.net/zim/+bug/1572542

I still think support URI scheme would be nice.

jaap-karssenberg:
Problem is that anything that starts with a word followed by a ":" could also be a zim page. So need bit more specific matching of the URI to prevent conflicts. This could be implemented in a plugin.

introt:
As a workaround, since Zim falls back to using the web browser for unsupported `protocol://`-links, Linux users can set a custom url protocol handler as their browser. You'll need the `//` to prevent Zim from parsing the URI as a page link, but that's easy to strip (eg. with `"${1#//}"` in Bash), and can be hidden using the Interwiki links in `urls.list`. See eg. https://unix.stackexchange.com/questions/497146/create-a-custom-url-protocol-handler

PR #2555, opened 2024-01-21
add "internal-link" style to show internal links differently

X-Wei:
This is to add the ability to make internal links look different than external links.

cf. https://github.com/zim-desktop-wiki/zim-desktop-wiki/discussions/2553#discussioncomment-8198243

Tested:
With this change, the internal links are rendered differently: 
image

And it's possible to add an entry in `styles.conf` to customize: 

```
[Tag internal-link]
foreground=#8e7cc3
underline=PANGO_UNDERLINE_SINGLE
```

with the modified styles, the internal links render as specified: 
image

### DIFF
--- a/zim/gui/pageview/textbuffer.py
+++ b/zim/gui/pageview/textbuffer.py
@@ -285,6 +285,7 @@ class TextBuffer(Gtk.TextBuffer):
 		'sub': {'rise': -3500, 'scale': 0.7},
 		'sup': {'rise': 7500, 'scale': 0.7},
 		'link': {'foreground': 'blue'},
+		'internal-link': {'foreground': 'green'},
 		'tag': {'foreground': '#ce5c00'},
 		'indent': {},
 		'bullet-list': {},
@@ -796,8 +797,12 @@ class TextBuffer(Gtk.TextBuffer):
 		if hasattr(href, 'uri'):
 			href = href.uri
 		assert isinstance(href, str) or href is None
-
-		tag = self.create_tag(None, **self.tag_styles['link'])
+		url: str = href or ''
+		is_external_link : bool = url.startswith('http://') or url.startswith('https://')
+		if is_external_link:
+			tag = self.create_tag(None, **self.tag_styles['link'])
+		else: 
+			tag = self.create_tag(None, **self.tag_styles['internal-link'])
 		tag.zim_tag = 'link'
 		tag.zim_attrib = attrib
 		if href == text or not href or href.isspace():
--- a/zim/gui/pageview/textbuffer.py
+++ b/zim/gui/pageview/textbuffer.py
@@ -797,12 +797,9 @@ class TextBuffer(Gtk.TextBuffer):
 		if hasattr(href, 'uri'):
 			href = href.uri
 		assert isinstance(href, str) or href is None
-		url: str = href or ''
-		is_external_link : bool = url.startswith('http://') or url.startswith('https://')
-		if is_external_link:
-			tag = self.create_tag(None, **self.tag_styles['link'])
-		else: 
-			tag = self.create_tag(None, **self.tag_styles['internal-link'])
+		is_internal_link = (link_type(href or '') == 'page')
+		styles = self.tag_styles['internal-link' if is_internal_link else 'link']
+		tag = self.create_tag(None, **styles)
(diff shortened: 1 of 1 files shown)


X-Wei:
This commit shall fix https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/197 
And one can mimic this PR and make a similar commit to fix https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/310

zim/gui/pageview/textbuffer.py
```diff
@@ -796,8 +797,12 @@ def _create_link_tag(self, text, href, **attrib):
 		if hasattr(href, 'uri'):
 			href = href.uri
 		assert isinstance(href, str) or href is None
-
-		tag = self.create_tag(None, **self.tag_styles['link'])
+		url: str = href or ''
+		is_external_link : bool = url.startswith('http://') or url.startswith('https://')
```
introt: This doesn't detect all external links, eg. `mailto:` links (see also: #513). Link parsing can get pretty complicated quite quickly so it'd be best to use our existing facilities: the `link_type` in `zim.parsing`. You can see how it is used here to display a different tool tip based on whether the link is internal or external:

https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/70355fa66c9b6abfbb15aec2675cbae55a6153fe/zim/gui/pageview/__init__.py#L879-L888

(You can also see another instance of anchors being anchors (see also: #1942))

introt (changes requested):
That was quick - nice work! Please see the attached comment for suggested changes re: link parsing.

zim/gui/pageview/textbuffer.py
X-Wei: Thanks!
I've updated the PR, PTAL.

zim/gui/pageview/textbuffer.py
```diff
@@ -796,8 +797,9 @@ def _create_link_tag(self, text, href, **attrib):
 		if hasattr(href, 'uri'):
 			href = href.uri
 		assert isinstance(href, str) or href is None
-
-		tag = self.create_tag(None, **self.tag_styles['link'])
+		is_internal_link = (link_type(href or '') == 'page')
+		styles = self.tag_styles['internal-link' if is_internal_link else 'link']
```
introt, line 802: Nit: `is_internal_link` is good for readability, but `styles` only exists to get unpacked and doesn't communicate extra info compared to `self.tag_styles`

introt (approved):
Looks good to me, thanks!

jaap-karssenberg:
This is indeed an easy improvement and I will merge it.

However a limitation is that this change works on link creation, but does not take into account editing links. Not sure if that is really an issue. Would need to think through the ways you can edit links.

X-Wei:
> This is indeed an easy improvement and I will merge it.
> 
> However a limitation is that this change works on link creation, but does not take into account editing links. Not sure if that is really an issue. Would need to think through the ways you can edit links.

Thanks Jaap! 
What do you mean by "editing links"? I've been using this change for a few days, I can edit a link by right-click or pressing ctrl-K, and if I change an internal link to external, the appearance would change immediately.

btw I've fixed the import and by default added a double underline for internal links (I find this to be easier to distinguish), @introt please take another look, thanks!

introt:
> What do you mean by "editing links"? I've been using this change for a few days, I can edit a link by right-click or pressing ctrl-K, and if I change an internal link to external, the appearance would change immediately.

To me this sounds like what Jaap intended; other possible edge cases include copying and pasting links - if done between notebooks, should interwiki links look internal or external?

> btw I've fixed the import and by default added a double underline for internal links (I find this to be easier to distinguish), @introt please take another look, thanks!

Can't comment on styling, that's up to our BDFL :)

X-Wei:
@introt gentle ping on the review

introt (approved):
> Can't comment on styling, that's up to our BDFL :)

jaap-karssenberg:
Merged with some updates. Changed it to "page-link" instead of "internal-link" to allow for other specific link types later and updated config file and manual file as well.

Kept both link types blue by default and put underline in the default config file for external links. But you can configure to your taste now :)

Thanks for demonstrating how easy this feature was to patch in - I was thinking a bit to difficult about it.

PR #2569, opened 2024-02-06
Plugin for detecting and merging changes on disk [updated]

joukhaha:
Added simple test case (+bugfixes) as discussed in previous request (#1986). The plugins has been seemingly working on an arch installation, but has not been tested extensively in other distros or OS, so no guarantees...

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -111,7 +111,8 @@ class MergerPageViewExtension(PageViewExtension):
            logger.debug('Reset poll timer')
            self._autoload_timer = GObject.timeout_add(
                    timeout*1000, # s -> ms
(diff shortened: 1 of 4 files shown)


introt (commented):
Could you do some cleanup? There seems to be quite a lot of commented out bits without explanation

introt:
As for the rest, I'm assuming Jaap requires more comprehensive tests than simple bringup - cases involving merge conflicts and the expected output of your plugin. I suggest refactoring such that what you're currently verifying using "if Debug" blocks and your eyes can be verified by the test suite. Please let me know if you need more concrete examples and I'll try to get back to you in a timely manner (less than a month, I hope).

PR #2570, opened 2024-02-09
Draft: Table editor improvements

tridian-tn:
PR for WIP on various small improvements to the tableeditor plugin.

| Issue| Description| Commit(s)| Notes |
|--------|--------|--------|-------|
| Discussion #2564 | Tab key navigation through cells |  | WIP: cell navigation cosmetic issue |

### DIFF
--- a/zim/plugins/tableeditor.py
+++ b/zim/plugins/tableeditor.py
@@ -763,68 +763,97 @@ class TableViewWidget(InsertedObjectWidget):
 	def on_treeview_key_press_event(self, treeview, event):
 		''' Event handler for key-press-events in treeview '''
 		keyname = Gdk.keyval_name(event.keyval)
-		shift_pressed = event.state & Gdk.ModifierType.SHIFT_MASK
-		path, col = treeview.get_cursor()
-		columns = [c for c in treeview.get_columns()]
-		colindex = columns.index(col)
-		rowmove = 0
-		colmove = 0
 
+		row_direction = 0
+		col_direction = 0
+
+		# Detect cursor movements
 		# ISO_Left_Tab should be enough to signify shift-tab, but there's some doubt in sources on the Web, so also
 		# cope with 'Tab' where shift is pressed.
-
-		# Move cursor to the next/previous column/row
 		if keyname == 'ISO_Left_Tab':
-			colmove = -1
+			col_direction = -1
 		elif keyname == 'Tab':
-			colmove = -1 if shift_pressed else 1
+			shift_pressed = event.state & Gdk.ModifierType.SHIFT_MASK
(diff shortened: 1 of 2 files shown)


zim/plugins/tableeditor.py
tridian-tn, line 1: @jaap-karssenberg @introt sorry to bother but I've hit a problem with work I'm doing here and I'm stumped. If either of you can help then that would be brilliant but I also appreciate you may not have the time so if you know of someone else I can ask, I'd be very grateful for a pointer.

In summary, these four commits in this PR (which will be squashed into one when I've finished, with temporary logging removed) add tab/shift-tab navigation between cells in the table editor. In general, it works well but there's an occasional issue if a table cell's contents is clicked on *after* gaining focus, when the text is deselected and the cursor is ready for editing. If you then press tab a few times (and quickly), the focus moves between cells as expected...and then out of the blue, the editable cell appears in the top left of the table. It works as you'd expect, but it's in the wrong place. Press tab again and the editable cell goes back to where it should be.

I've read enough to know that this is the `editable` of the Gtk.CellRendererText, and is (I believe) a `GtkEntry` widget. I've also checked through logging that the various cell editing events appear to be firing in the right order.

Beyond that, and a failed attempt to find something helpful in the GTK docs, I'm not sure what to check next. I've attached a recording of the behaviour. A few things of note:

* When I didn't click on a cell's contents, the tabbing was fine
* Then when I did, you can see (briefly!) the focused edited cell popping into the top left
* ....but it's frustratingly intermittent

ZimTableTabbing

I'm not a Python or GTK programmer by trade, so this is all quite new to me and I'm probably missing something obvious. If anything strikes you as something to try, or look into, I'd be really grateful for a hint.

Many thanks,
Trystan

introt: Hi Trystan,

Given the popping into the top left, I get the feeling something is getting reset. It might have to do with `on_cell_changed` or `autosave_title_cell` as they set `self.currently_edited` to `None`. To see if either could be relevant, you could add a counter that increments on every tab and runs `self.currently_edited = None` on every third tab. If that ain't it, it might be time to break out a real debugger to record what variables change when that happens - simple logging might not cut it here.

Happy Hacking,
introt

tridian-tn: Thanks for taking a look @introt . I'll see what I can find

tridian-tn:
2x completed fixes moved to https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/2588 PR

tridian-tn:
Closing pull request - I currently don't have time to work on it. Will open a new clean request if I get a chance to return to it.

PR #2587, opened 2024-03-17
Fix: remove pageindex last selected filter

PixelOfDeath:
The "last selected" filter is unneccessare because there is no event triggered if the page is already selected.

Also this filter causes a bug where selecting the previous selected page does not work, if the page switching happened with anything else then selecting it in the pageindex. E.g. clicking a link in a page to a page, or using the back-button.

### DIFF
--- a/zim/plugins/pageindex/__init__.py
+++ b/zim/plugins/pageindex/__init__.py
@@ -400,7 +400,6 @@ class PageTreeView(BrowserTreeView):
 		self._autoexpanded = None
 		self._autoexpand = True
 		self._autocollapse = True
-		self._last_selected_path = None
 
 		column = Gtk.TreeViewColumn('_pages_')
 		column.set_expand(True)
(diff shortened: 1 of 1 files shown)


jaap-karssenberg:
Can you describe a specific scenario that shows the validity of this change - e.g. behavior before / after - so I can verify?

Thanks,

Jaap

PixelOfDeath:
Very simple example. Two pages, one has a link to the other page like the screenshot.

If this link is used instead of selecting Page2 in the page index, then the _last_selected_path value is not updated. And after that clicking Page1 in the page index has no effect. (bug)

If this event filter actual has a reason to exist that I do not understand (maybe on Mac?), then the _last_selected_path must be updated on other page changing events, too.

example

jaap-karssenberg:
Thanks for looking into this - looks indeed this is left over logic that no longer serves a purpose. Will merge and keep on eye during usage for any edge case I forgot about.

PR #2593, opened 2024-03-31
Text Expander

mbluett88:
Version 0.1

Author: Mike Bluett
Description: A simple foldable (collapsible/expandable) text area plugin for Zim.


mbluett88:
Hopefully, this is the correct way to incorporate this simple version of my Text Expander as a plugin for Zim.

jaap-karssenberg:
Hi Mike, thanks for submitting this as a plugin, this makes it easier to merge in the future. For now I'll consider it a "draft" proposal.

The next step would be to add documentation - the test suite now fails because there is no manual page for this plugin - and test cases specific for this plugin to verify it works and prevent accidental breaks once it is in the main code.

Also, as discussed previously, I highly encourage looking for a name / description that more reflects the use case of the plugin object rather than the name of the widget used. My tip would be to think about what would be the dead-tree paper equivalent these foldable sections. E.g. annotations ?

Hope this helps to find your way forward.

Regards,

Jaap

mbluett88:
"Also, as discussed previously, I highly encourage looking for a name / description that more reflects the use case of the plugin object rather than the name of the widget used. My tip would be to think about what would be the dead-tree paper equivalent these foldable sections. E.g. annotations ?"

These sections are really not annotations. To me an annotation is what you would add as a side note in the column next to some text.

A word that more closely describes this functionality is "commentary". But, the problem with that is that it will not be obvious to people what it's purpose is. Maybe people who are formalists when it comes to languages might understand it's purpose.  That is why "foldable text" or "text collapser" or "text expander" are more intuitive. It is really not necessary to inspect documentation to know what it does. Whereas, "commentary" definitely would require perusing the documentation. Annotations, likewise. It is not intuitive how those would actually be used. My two cents. But, you are the one who maintains Zim so it is probably best that the choice be yours.

pushkarajthorat:
I have tried this plugin, it looks promising. I am new to the overall ecosystem, can we somehow integrate it with the bullets to make them foldable?

introt:
> My tip would be to think about what would be the dead-tree paper equivalent these foldable sections. E.g. annotations ?

I'd skip the equivalent and name it "Foldable text sections" or similar, as folding is the verb used in other software - text expander sounds like text snippet shortcuts, eg. https://github.com/introt/zim-plugin-text-replacer - though instead of separate blocks, those tend to fold existing lists or heading levels as alluded to by @pushkarajthorat (you might want to see/follow #1667)

jaap-karssenberg:
Thanks @Introt, that suggestion makes sense

--- Jaap

mbluett88:
Since Jaap seems adamant in having this Text Expander (or possibly better name "Folding Text" as proposed by Introt) implemented as a plugin, I have been working towards that and have succeeded with a lot of it (i.e, formatting, bullets, links, etc.). However, part of the Links effort requires having the "Remove Link" menu item and the "Edit Link or Object" menu item enabled for use within the Expander TextView. This enabling occurs in PageView.do_mark_set().

All of my efforts to enable these menu items have failed, unless of course, I just force them to be permanently enabled, which is not what was originally intended.

The only solution I can see is to have a reference to the pageview instance available within the Expander plugin.

I have another implementation where these menu items work properly but it is not based on a plugin and it, of course, makes use of the pageview instance.

If someone has a suggestion of how to make these menuitems function the way they were intended to function, without making use of the pageview instance, I am all ears.

If not, I will continue development with the version I have that is not based on a plugin.

PR #2596, opened 2024-04-04
Fix macOS CI

dehesselle:
This fixes the test job on macOS. Most prominent issue was the lack of not having specified the minor version of the interpreter. Also we're using now Homebrew's Python and not the other installation that's present on the runner.

Sadly I've had to disable one more test to get to the green checkmark. Someone who actually uses Homebrew needs to take a look at this: how to tell Python to use the correct sqlite3 library with FTS support? I've tried changing `PATH` , `DYLD_LIBRARY_PATH`, `LD_LIBRARY_PATH` but couldn't get it to work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,16 +65,20 @@ jobs:
       - name: Install dependencies
         run: |
           brew install pygobject3 gtk+3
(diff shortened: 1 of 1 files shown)


dehesselle:
Hi @jaap-karssenberg , this is an easy merge as it only touches the workflow's macOS job.

PR #2610, opened 2024-05-05
distractionfree - set caret-color to text color

### LINKED ISSUE #2574: Blinking cursor  is not visible when using dark backgrounds for Windows Binary of Zim 0.75.2 “Distraction Free Editing” plugin.
**Description**

The following issue applies to the Windows binary. When using _Zim 0.75.2 “Distraction Free Editing”_ plugin, the blinking cursor  is not visible when using dark backgrounds. It appears that when using this mode, the cursor color is black (irrespective of what the text color is set to). So if you configure the plugin to a black background, you won’t be able to see where the cursor is located on the screen. In addition, when you have the plugin configured for a dark background, you won’t be able to see some of the menu bar buttons as they are also black in color.

**To Reproduce This Behavior on Windows**

1. Go to Edit /  Preferences / Plugins / Distraction Free Editor / Configure
2. Set text background color to black.
3. Set text foreground color to white.
4. Click 'ok' and exit the preferences window.
5. Hit *F11* to enter the distraction free editor.

**Expected behavior**

I expected that the blinking cursor would be the same color as the text foreground color in the distraction free mode, similar to the behavior of the Zim Windows binary in versions 0.6x.

**Environment**

-  Using WIndows 10 (22H2)
 - Zim Version 0.75.2

introt:
fixes #2574

problem: is the default the exact text color? if not, this'll degrade the appearance

alt fix: separate caret-color pref in plugin - perhaps some would enjoy a more colorful cursor anyway?

### DIFF
--- a/zim/plugins/distractionfree.py
+++ b/zim/plugins/distractionfree.py
@@ -62,6 +62,7 @@ class DistractionFreeMainWindowExtension(MainWindowExtension):
 		css = '''
 		#zim-pageview text {
 			color: %(textcolor)s;
+			caret-color : %(textcolor)s;
 			background-color: %(basecolor)s;
 		}
 		#zim-window-main-box scrolledwindow{


jaap-karssenberg:
For discussion for further fixing: maybe distraction free should have the option to not change the colors and just use the dark theme mode?  (If dark mode is the default, just change screen layout, if it is not the default switch the application preference on full screen.)

PR #2620, opened 2024-06-02
Add Foldable Text section within a Zim page.

mbluett88:
This is not based on a plugin. Instead it uses InsertedObjectAnchor and is inserted into a Zim page from the "Insert" menu option.

Limitations:
  * Text is not searchable
  * Text cannot be selected with the mouse
  * SourceView cannot be added to a Foldable Text section
  * Right-click context menu does not include all options within a Foldable Text section

### DIFF
--- a/data/manual/Help/Menu_Items.txt
+++ b/data/manual/Help/Menu_Items.txt
@@ -197,6 +197,9 @@ Insert an image.
 **Attachment**
 Attach a file to the current page. Will prompt for a file, the file will then be copied to the notebook folder and a link inserted in the current page.
 
+**Foldable Text**
+Insert an expandable/collapsible section of text within a page.
+
 **Screenshot**
 Prompt with options to take a screenshot and insert it as an image.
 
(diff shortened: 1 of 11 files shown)


mbluett88:
I am looking for input on how to implement the right-click context menu items (such as Edit Link, Edit Properties) within the Foldable Text (Expander) textview.

So, if you add a link or an image I would expect to see the related menu items when right-clicking on that object.

I have spent a good deal of time trying to find a solution but so far no luck.

I am thinking it might be necessary to create a Menu class which both PageView and ExpanderWidget can use to accomplish this goal.

gacarrillor:
I'd like to say that this would be a HUGE improvement for Zim.
I have notes with a lot of content that I would like to collapse and only visit sporadically, but that belong to the same note. 

Thanks for raising this and for working on this. I hope users can see this landing in Zim in the (near) future!

shanness:
This would be so awesome.  The "Text is not searchable" is a show stopper for me though.  I'm currently a bit torn between migrating to obsidian or digging in to commits like this (or your unfinished plugin) to try to get folding sections.
Zim is awesome, I've got 6600  pages going back 20 years.

PR #2625, opened 2024-06-14
Skip favicon.ico in tests/attachmentbrowser

oturpe:
gdk-pixbuf 2.42.11 [removes support][1] for multiple less common image formats, including ICO. Skip that format in thumbnail generation tests to make the test pass.

Other than the test failure, Zim remains functional with gdk-pixbuf 2.42.11. Attachment Browser still works even for ICO files. The thumbnails naturally fail to show up, but otherwise the browser remains functional. ICO files cannot be attached to pages anymore, instead error dialog with text _File type not supported: image/x-icon_ is displayed.

[1]: https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/commit/e052a112075a19fb75f1f2ff3de4c82923de13f2

### DIFF
--- a/tests/attachmentbrowser.py
+++ b/tests/attachmentbrowser.py
@@ -29,6 +29,8 @@ class TestThumbnailCreators(tests.TestCase):
 			for i, basename in enumerate(dir.list_names()):
 				if basename.endswith('.svg'):
 					continue # fails on windows in some cases
+				if basename.endswith('.ico'):
+					continue # Support removed in gdk-pixbuf 2.42.11
 				file = dir.file(basename)
 				thumbfile = thumbdir.file('thumb--' + basename)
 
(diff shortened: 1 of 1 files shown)


oturpe:
The test failure at _MSYS2 Python 3.x on windows-latest_ is:

```
FAIL: runTest (tests.widgets.TestFileDialog.runTest)
  File "D:/a/zim-desktop-wiki/zim-desktop-wiki/tests/widgets.py", line 414, in runTest
    self.assertIsInstance(adapt_from_oldfs(myfolder), LocalFolder)
AssertionError: None is not an instance of <class 'zim.newfs.local.LocalFolder'>
```

That is not related to this change at all, and I see the same in other recent pull requests, too. So I don't think there is anything to fix in this pull request because of that.

Issue #2604, opened 2024-04-27
Assign each page a hidden and unused UUID stored in the page headers


Osndok:
**Is your feature request related to a problem? Please describe.**
I have several ideas for zim plugins and compelling workflows, but most of them require identifying a particular page even if it has been moved or renamed (which is why using a page's path does not work).

**Describe the solution you'd like**
I would like zim to opportunistically store an unseen UUID in the headers of each zim page as it creates or saves them.

Ideally, there would also be a way to ask zim via a custom URL protocol handler to navigate to a page having a particular UUID, but just having the identifier in place is technically enough for my purposes (as I can scan all the pages and reverse-lookup the page's current path).

**Describe alternatives you've considered**
* Tracking low-level file inode numbers external to zim.
* Tracking movement via git history and its implicit file-renamed detection logic.

**Additional context**

Example of how a stored page might change:
```diff
diff --git a/Home.txt b/Home.txt
index 62257fa..996a81d 100644
@@ -1,6 +1,7 @@
 Content-Type: text/x-zim-wiki
 Wiki-Format: zim 0.6
 Creation-Date: 2024-04-27T11:36:04-05:00
+UUID: aefc3ab9-9cd4-458f-a3bb-d41fef8b61cc
 
 ====== Home ======
 Created Saturday 27 April 2024
 
 Hello, World!
```

The UUID should not change when for ordinary UI operations (load/save cycles, rename/move page), but might change if the user effects or removes the UUID line during an "edit source" operation.

I searched open issues, but only found this discussed implicitly (as a pre-req) on #54

jaap-karssenberg:
I'm open to a patch that sets such a header on page creation. Would need
some testing, but I believe the headers are not changed in any operation
after page creation.

tzimisces:
I was looking for that functionality also.  I've made a PR for it.

PR #2630, opened 2024-07-02
Addresses issue #2604: adds an identifier to pages.

tzimisces:
ADD: Extends the header of a Zim-page with an extra field named
'Page-Identifier'. The field gets added to a page on page
creation, and also - if the field is not present yet - when
it is saved.


jaap-karssenberg:
Initial patch looks good to me; thanks!

Next step would be to add some API around it for plugins to work with:
- Add the page ID to the pages index table & add a method to the index view to retrieve pages by ID
- Add a "getter" or property to the Page object to get the page ID and trigger initialization if the page did not have one yet (for existing notebooks)

Kind regards,

Jaap

tzimisces:
Hello Jaap,

Sorry for the late reply, it took a while for me to make some time to look into this further.

I've fabricated something which works for new and existing pages. New pages and pages without unique ID's get one assigned when a page is saved.
I'm assuming that this always happens via the Notebook so the triggering of signals will cause the ID being picked up and stored in the database.

Editing with an external editor does not work yet (hmm, maybe my assumption made above isn't quite right), and I haven't looked at importing existing files for now - I'll do that later on.

In the mean time, I'm open to suggestions :)

Regards,

  Bob

Osndok:
It appear that the number in the title did not link the issue, so adding that here: #2604

jaap-karssenberg:
Thanks for the patch so far - looks good and I'm tempted to go ahead and merge it. Two items I would like to change/see changed:

1. For the new methods added in the index view and in the page object, I miss test cases
2. The argument for the "page-changed" signal changed from "content" to "tree" - I would prefer to provide the content and the meta data as separate field - current the tree object is mutable so I worry about side effects of using the tree here -- OTOH I do want to make the tree object immutable in the future, so that would make this a lesser issue -- have you changed all places where this signal is used?

tzimisces:
So a finally made some time to get to grips with this - Python isn't a daily routine for me :)

Ad 1) I added a number of testcases for the added / changed functionality.
Ad 2) I merely changed the parameter name to reflect the naming used in the calling method - the object type wasn't changed :) Unless I'm missing something in the semantics, I suppose this is no problem?

While adding the testcases I found some minor misbehaviour - that has been fixed.

The behaviour that when importing a file containing a page identifier, an already registed page with the same page identifier will have its identifier removed, may be not the nicest way of doing things, but then this can be considered a corner case anyway.

zim/notebook/index/pages.py
```diff
@@ -76,7 +77,8 @@ def __init__(self, db, layout, filesindexer):
 				mtime TIMESTAMP,
 
 				source_file INTEGER REFERENCES files(id),
-				is_link_placeholder BOOLEAN DEFAULT 0
+				is_link_placeholder BOOLEAN DEFAULT 0,
+				page_identifier TEXT UNIQUE,
```
Osndok, line 81: Do we need a database migration (or version number increment) somewhere to properly support this, or is this change enough for that?

zim/notebook/index/pages.py
```diff
@@ -200,9 +202,9 @@ def _insert_page(self, pagename, is_link_placeholder, file_id=None):
 		sortkey = natural_sort_key(pagename.basename)
 		try:
 			self.db.execute(
-				'INSERT INTO pages(name, lowerbasename, sortkey, parent, is_link_placeholder, source_file)'
-				'VALUES (?, ?, ?, ?, ?, ?)',
-				(pagename.name, lowerbasename, sortkey, parent_row['id'], is_link_placeholder, file_id)
+				'INSERT INTO pages(name, lowerbasename, sortkey, parent, is_link_placeholder, source_file, page_identifier)'
+				'VALUES (?, ?, ?, ?, ?, ?, ?)',
+				(pagename.name, lowerbasename, sortkey, parent_row['id'], is_link_placeholder, file_id, None)
```
Osndok, line 207: With just a cursory reading, I'm not sure why we skip the `page_identifier` here.

zim/notebook/index/pages.py
```diff
@@ -260,14 +262,41 @@ def update_parent(self, parentname, allow_cleanup=lambda r: True, oldrow=None):
...
+			# Another situation is the existence of "ephimeral" pages. That is,
+			# pages which are present in the database, but not on the
+			# filesystem. This happens when a page is moved, and some other page
+			# refers to it, but its link is not updated.
+			#
+			# TODO Handle duplicate page identifiers on another level.
+			self.db.execute(
+				'UPDATE pages SET mtime=?, page_identifier=NULL WHERE page_identifier=?',
+				(mtime, page_identifier),
+			)
```
Osndok, line 285: I might be reading this wrong, but this smells a bit paranoid, or maybe that we may be trying to solve a problem we don't have in the wrong place.

i.e. I didn't see a migration/schema change, so I don't even know if the db even requires `page_identifier` to be unique, but much like your comment here... it should be handled elsewhere. Having dupes in the database or throwing here is not as bad as silently changing (thus dereferencing) another page when we are unsure which one is 'correct', and it also makes every `update_page` need two db operations when the expected case is that this one will only find "ourself".

I do understand that page_identifier is conceptually unique, but let me see if understand a plausible cause and effect:
1. Given that `PageA` exists with id `1`
2. Some other process duplicates it (thus duplicating it's identifier). e.g. the user does `cp PageA.txt PageB.txt`
3. Zim detects the new `PageB` and imports it
4. When `PageB` is updated, this code wants to remove `PageA`'s identifier, and declare `PageB` as true holder of id `1`
5. When `PageA` is reloaded from disk later, and updated, this code will remove `PageB`'s identifier in the db, and declare that `PageA` is the true holder of id `1`

...do I have that about right? It seems a bit thrashy, but maybe you had a `mv` in mind for step-2.

Assuming we do want to enforce uniqueness (which I could see the value of), then it seems to me that the existing db entry would win (assuming the page is still present at that path), and the newbie that we are updating needs to be assigned a new id, not the other way around.

Osndok (commented):
Thank you for the work, I'm really looking forward to getting something like this merged/released.

zim/notebook/index/pages.py
tzimisces: Ah, yes, we do indeed need a database migration. I'll increase the database version to 0.9 - if I'm correct, that triggers a re-initialisation and is the way to go?

PR #2655, opened 2024-08-19
Use markdown blockquote to export indented paragraphs

### LINKED ISSUE #41: Support blockquotes
Support blockquotes to mark sections of text that are a quotation

- source: prefix line with one or more ">" (email convention)
- rendering: show one or more lines in the left margin
- implementation: embedded pageview

Important to test is that the quote can contain any markup, including headings, lists, images. By using an embedded pageview object, all features are supported, but nested.

Depends on making embedded pageviews work in the first place. This means menubar actions are directed to the view that has focus and signals are propagated. Probably needs some kind of facade design where a facade manages multiple nested buffers.

raigoinabox:
Fixes #41

Hello!

I implemented a simple solution to support Markdown blockquotes. I used the thought in the last comment: to map an indented paragraph to a blockquote. I even first thought of it myself, started to implement it, saw a TODO about an open issue and after searching, saw the same thought in there. It was really easy to implement it this way.

- [x] I updated the test export.markdown file using the test update.sh script.
- [x] Verified that my change is in unit tests (the export.markdown file changed)
- [x] I ran all tests, saw no errors.
- [x] I exported a file with indented and not indented paragraphs, saw the correct result there.

For a little background, I have a fun and ambitious project of enabling zim to use markdown format by default (reading and writing). When I started using the testing system to start showing me parser errors, this missing functionality immediately jumped out and is in some form required to make all the tests pass after a Markdown parser is added.

### DIFF
--- a/tests/data/formats/export.markdown
+++ b/tests/data/formats/export.markdown
@@ -8,13 +8,13 @@ reprehenderit in voluptate velit esse cillum dolore eu fugiat
 nulla pariatur.  Excepteur sint occaecat cupidatat non proident,
 sunt in culpa qui officia deserunt mollit anim id est laborum.
 
-Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do
-eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim
-ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
Cool! I'll have a look at merging it this weekend

Regards,

Jaap

Issue #1727, opened 2021-11-06
Settings Option to Generally Accommodate Dark Themes


wattahay:
(Alternatively, include a second style file that can replace the default)

### Problem

Users become rapped up in tweaking colors to **get the out-of-the-box features**.

There are various issues with dark theming outside of adopting the system theme. While these can be changed manually, **adoption of dark themes has become very widespread**, and a simple option would provide a great level of convenience cumulatively.

- Link text is too dark, as to blend into the background
- Verbatim text is too dark, as to blend into the background
- Highlighted text is too light, as compared to the system theme's light text
- [Maybe] Horizontal lines and table lines? (not problems for me personally)
- [Maybe] Icons as well? (not problems for me personally

### Related Open issues:

- https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/1404
- https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/1178
- https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/1003
- https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/681
- https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/132

### Solution 1
GUI Dark Theme option to swap style file

### Solution 2
GUI Dark Theme option to swap style file _and_ icons

### Solution 3
Provide alternate style file with dark system themes in mind

jaap-karssenberg:
There is a generic Gtk option to switch to a dark version of the default
theme. We could expose an option in the preferences UI that uses that.

I also support the idea of shipping multiple style configs. To implement
this, the file would need to contain a key with it's name (translatable,
like names in .desktop files) and the preferences could show a dropdown of
the files found.

To address the icons, we could also ship multiple icon sets. In this case I
would prefer the style config file to contain the name / path of the icons
to be used, so you have a single style selection for both font colors and
icons.

-- Jaap

manuel-91:
I would be happy to see a dark mode for zim. However not too many people seem to need it...

The Link text is barely visible, especially on older/darker/cheaper monitors or in nightmode. Can we just set the Link color to this light blue? That would be a quickfix (with much less effort) for those who use a dark system theme... here the comparison of the Help->About menue (light blue) and the dark blue (normal note link):

![image](https://user-images.githubusercontent.com/45081089/159159749-7cfb6aae-d251-4186-ae96-c27b940fe20b.png)

In light mode it look like this:

[...]

PR #2688, opened 2024-10-19
Update default colors to match the Tomorrow Theme

bruvellu:
Dear Jaap,

As discussed in #1727, this PR proposes to update the default Zim color scheme to match the [Tomorrow Theme](https://github.com/chriskempson/tomorrow-theme). This palette provides a more balanced set of colors that adapt equally well for light and dark themes. It is also similar to the current theme, so it won't be a drastic change of style. I put some before/after screenshots below.

I would love to hear what you and the community think about these changes. There are some tweaks and ideas that I would like to discuss if they make sense to be included as defaults. But I'll add these updates later on. For now, any feedback on this initial version is welcome!

## Headings
tomorrow-zim-headings

## Formatting
tomorrow-zim-formatting

## Task priority
tomorrow-zim-priority

## Dark theme (formatting)
tomorrow-zim-formatting-dark

Cheers,
Buno

### DIFF
--- a/data/style.conf
+++ b/data/style.conf
@@ -38,18 +38,18 @@ strikethrough = true
 
 [Tag code]
 # For in-line verbatim text (gray background)
-foreground=#4d4d4c
+foreground=#4d4d4d
 background=#efefef
-foreground[darktheme]=#c5c8c6
+foreground[darktheme]=#c7c7c7
 background[darktheme]=#3c3c3c
 family = monospace
 #wrap-mode = none
 
 [Tag pre]
 # For multi-line verbatim blocks (gray background)
-foreground = #4d4d4c
+foreground = #4d4d4d
 paragraph-background = #efefef
-foreground[darktheme] = #c5c8c6
+foreground[darktheme] = #c7c7c7
 paragraph-background[darktheme] = #3c3c3c
 family    = monospace
 wrap-mode = none
--- a/data/style.conf
+++ b/data/style.conf
@@ -27,14 +27,12 @@ style = italic
 [Tag mark]
 # For highlighted text (yellow)
 background=#eab700
-#FIXME: Dark background is not being applied
-background[darktheme]=#f0c674
+background[darktheme]=#b58c1d
 #underline = single
 
 [Tag strike]
 # For struck-out text (gray)
 foreground=#8e908c
-#FIXME: Dark foreground is not being applied
 foreground[darktheme]=#969896
 strikethrough = true
 
@@ -43,7 +41,6 @@ strikethrough = true
 foreground=#4d4d4c
 background=#efefef
 foreground[darktheme]=#c5c8c6
-#FIXME: Dark background is not being applied
 background[darktheme]=#282a2e
 family = monospace
 #wrap-mode = none
@@ -52,7 +49,6 @@ family = monospace
 # For multi-line verbatim blocks (gray background)
 foreground = #4d4d4c
 paragraph-background = #efefef
-#FIXME: Dark theme is not being applied
 foreground[darktheme] = #c5c8c6
 paragraph-background[darktheme] = #282a2e
 family    = monospace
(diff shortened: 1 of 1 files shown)


UnisTorvalds:
I find that on light theme the contrast (therefore readability) is reduced, esp. with headlines, highlights, links, and task priorities.  I would see this as a regression.
However it's true that on dark theme readability is drastically improved.
I don't use dark theme so personally I am ambivalent to this change.  Would there be an easy (in app) way to change themes without swapping css files?

bruvellu:
Hi @UnisTorvalds, thanks for the input. Tomorrow is indeed a medium-contrast theme, so the colors are not so bright. However, I still find it comfortably readable and more balanced for the eyes compared to the default colors.

But I agree with you that some colors are a bit muted. Here's Tomorrow's blue `#4271ae` versus GitHub's links `#0969da`. We can tweak these specific colors to improve their contrast without altering the color scheme too much to make sure that Zim's default theme has good overall readability.

Alternatively, we could use another base color scheme. There are other options that might fit in case Tomorrow is not appealing for most (but I haven't tested them).

In any case, for now, the only way to change these colors in Zim is by editing the `style.conf` or CSS files (I believe), so not _in app_ options.

bruvellu:
I've made a few changes to the light theme to improve the contrast for highlighted and verbatim text. Colors for headings, tags, and links remain the same for now. But I did an experiment with color-coding the checked tasks. Please let me know what you think.

@jaap-karssenberg ~~I just noticed that `background[darktheme]` properties are not being applied. Could you take a look?~~ **Edit:** Sorry, they are applied when I manually set the `Prefer dark theme` option in Zim. I was testing by switching my desktop theme from light to dark (in this case, they are not applied).

tomorrow-zim-formatting-update1

bruvellu:
A few more updates. Tomorrow's colors are meant for foreground text. When I used them as background for highlighted text and priority tasks, they were either too dark (in the light theme) or too light (in the dark theme). Now, I tweaked them accordingly for better contrast, aiming for a [minimum contrast ratio](https://www.w3.org/TR/WCAG21/#contrast-minimum) around 5:1. Here's how the new version compares to Zim 0.75.2:

Task priorities light theme:
tomorrow-zim-formatting-priority1

Task priorities dark theme:
tomorrow-zim-formatting-priority-dark1

Formatting dark theme:
tomorrow-zim-formatting-dark-formatting1

bruvellu:
A few more tweaks to the light theme, mostly to improve contrast for colored text. Here's how it currently looks:

tomorrow-zim-formatting-update2

jaap-karssenberg:
I've also made a change to the default `style.conf`, it now supports different color tags for light and dark theme, removing the need for one set that works with both. I'm open to including more styles, but the readability for both light and dark themes should be addressed now.

bruvellu:
Hi @jaap-karssenberg, I don't understand what you mean. The modifications in this PR already use the new light/dark tags in `style.conf` and I've tweaked the colors of both variants to be consistent and have a good contrast for general use. This is how light/dark versions currently look:

zim-lightdark

Let know if you want anything changed, or if you prefer not to have these as default, how should I submit as additional style.

Cheers!
Bruno

jaap-karssenberg:
Hi @bruvellu ,

Apologies, I was not reading carefully and took this as a "work around" for #1727 with one color set for both light and dark themes. Reading more closely I understand you work on two variants.

Previous I took the "Tango" palette as the basis, as that was positioned from the freedesktop.org side to be a good x-platform default palette. I think indeed many of your tweaks make sense and indeed improve the aesthetics.

Looking at the screenshots my personal preference would be to keep the brighter green from the original Tango palette for the headings - but this is also me being used to that color for the past 19 years, so might not be unbiased ;)  Also I would not prefer the strike-through & text color for tasks because it makes my notes look really busy.

Thinking what would be a good way to decide on aesthetics in an "objective" way.

Regards,

Jaap

jaap-karssenberg:
Playing around with the theme, but afraid I'm not convinced to change the default. I think the next step is to make a preference for the color scheme that can be changed via a dropdown or similar.

Kind regards,

Jaap

bruvellu:
Hi @jaap-karssenberg!

Yes, it makes sense to stick with the standard Tango colors as the default. The colored strikethroughs were an experimental idea I had while tweaking this theme over the past months. I ended up liking them as visual cues, but it can indeed feel busy with extensive task lists. I'll then close this PR and maintain the Tomorrow theme `style.conf` on a [separate repository](https://github.com/bruvellu/zim-tomorrow-theme). Do you have an idea already on the best way to implement the colorscheme preference? (perhaps something similar to the source view plugin?) I'd be happy to help to set this up.

Cheers,
Bruno

jaap-karssenberg:
Created #2766 to track adding additional style config to zim

Issue #2713, opened 2024-12-31
Moving line up or down picks up checkboxes


sojusnik:
**Describe the bug**
When using the move line up or down shortcuts from the line sorter plugin and a checkbox is in the way, then the moved text will pick up the checkbox, as shown in example 1 and 2 of the following screencast. This only occurs when the moved text has formatted text (example 1) or a linked text (example 2). With pure text (example 3) this bug doesn't appear.

[zim.webm](https://github.com/user-attachments/assets/8f5a1a2c-5dcf-41b6-8243-10935d09c068)

**Environment (please complete the following information):**
 - Version: 0.76.0 (stable PPA)
 - OS: Ubuntu 24.04

PR #2720, opened 2025-01-05
Fix issue with linesorter plugin (#2713)

roland-ruedenauer:
Moving a line up/down as implemented before triggered insertion of bullet when inserting text into a list context. Using insert_parsetree_at_cursor() avoids this unwanted behaviour. In addition, this change should also fix the text editing undo functionality.

### DIFF
--- a/tests/linesorter.py
+++ b/tests/linesorter.py
@@ -87,6 +87,18 @@ class TestLineSorterWindowExtension(tests.TestCase, TextBufferTestCaseMixin):
 		self.extension.move_line_up()
 		self.assertEqual(self.get_text(), 'B line\nA line\nC line\n')
 
+	def testMoveUpStyledTextIntoCheckboxList(self):
+		self.set_text('[ ] Checkbox #1\n[ ] Checkbox #2\nLine with **bold** text\n')
+		self.place_cursor(32)
+		self.extension.move_line_up()
+		self.assertBufferEqual(self.buffer, '[ ] Checkbox #1\nLine with <strong>bold</strong> text\n[ ] Checkbox #2\n')
+
+	def testMoveUpLineWithPageLinkIntoList(self):
+		self.set_text('* List item #1\n* List item #2\nLine containing a [[page link]]\n')
+		self.place_cursor(30)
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
Afraid I'm not convinced this is the right fix. I used the internal representation to make sure we do no changes to the content of the line. Using parsetree may affect the content.

Maybe you can contribute a test case for the issue instead? This would help me to understand what to fix with the internal format insert.

--- Jaap

roland-ruedenauer:
In my comment above I tried to explain what happened: it is the code at [textbuffer.py#L2124](https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/develop/zim/gui/pageview/textbuffer.py#L2124)

Please also note that the code for sorting lines already makes use of inserting parsetrees.

In my manual tests I found no problem so far with my changes. Under what conditions do you see the moved content being affected?

roland-ruedenauer:
This might be also of interest for you:
if there is a link or formatted text contained in the moved line, then do_insert_text() is called with a text element not ending with a newline character. See the check `... and not string.endswith('\n')` in do_insert_text. 

As mentioned in #2713 moving a plain text line up or down works as expected. Meaning if linedata in the linesorter plugin contains not just a single text item, then the insertion issue is triggered.

jaap-karssenberg:
Yes it did use parse trees, however with the "raw" attribute which is now
no longer existing. So I ported it to the new interface.

Op zo 5 jan. 2025 20:42 schreef Roland Rüdenauer ***@***.***>:

jaap-karssenberg:
An automated test case would help me along quite a bit and it ensures
issues do not come back in later changes.

Op zo 5 jan. 2025 21:52 schreef Jaap Karssenberg ***@***.***

roland-ruedenauer:
Sure. I will provide a test case within the next days.

But could you also provide some details about the possible data modification using the parsetree insertion you mentioned before?

roland-ruedenauer:
@jaap-karssenberg I've added 2 test cases. Please have a look.

Issue #841, opened 2019-05-13
Inline data for "image generator" inserted objects


jaap-karssenberg:
Currently all "image generator" plugins (equation, diagram, ...) rely on a data file and an image file. This is not robust and contaminates the attachment folder.

Instead data should be inline in the wiki page like for other inserted object types - e.g. code blocks.

The image can than be stored in a cache file using a hash of the data as the file name. Similar to how thumbnails are stored.

This will also adres issue #418.

pfalcon:
The benefit of the current solution is that any object is represented as just image tag (with a `type` query param). This is very smart and scalable scheme, e.g. any 3rd-party tool don't need to know anything about a particular object, it can just show a pre-rendered image instead.

Similarly, this offers an instant method to support arbitrary plugin for Markdown-format notebooks. In https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/26#issuecomment-300757202, you write:

> Probably also requires disabling some features in zim when source is markdown (e.g. insert of rich objects).

But there's absolutely no need to do that. In Markdown, any object can be just represented by native Markdown image syntax, just as in Zim-wiki format, objects are represented by native image syntax. I.e. what's now `{{./diagram.png?type=diagram}}` in Markdown would be `![diagram.png](./diagram.png?type=diagram)`.

Summing up, I'd recommend to be cautious about this change, weigh pros and cons. I guess I personally would be -1 on it due to reasons above.

jaap-karssenberg:
I think for the native markdown format the ability to edit inline is more
important than rendering.

Probably there will still be an markdown export format flavor.

Op vr 11 sep. 2020 22:09 schreef Paul Sokolovsky <notifications@github.com>:

pfalcon:
> I think for the native markdown format the ability to edit inline is more
important than rendering.

It seems that WYSIWYG (Rich Text) editing is "selling point" of Zim: how it differs from many other similar apps, and what users like and don't want to give up (based on the comments in https://github.com/zim-desktop-wiki/zim-desktop-wiki/issues/26).

[...]

PR #2736, opened 2025-01-11
Port equations plugin to use in-wiki-objects instead of separate files

maclomhair:
As mentioned in #841 the code for storing source data of image-generating plugins within the wiki page itself is already available. This PR tries to make use of this new feature by porting the equation editor plugin.

I've opted to add a new plugin so that the old equations can still be viewed. I couldn't figure out a nice way to e.g. have the same plugin load old equations and then store them in the new format - maybe someone has ideas how to properly solve this?

Another problem with the approach I've taken here is that objects which are stored as block elements in the wiki markup are now sometimes displayed as true inline elements. The way I did this was by removing the last newline before the object upon rendering, and re-inserting it when creating the element tree from the content of the textbuffer. I'm not sure this is a clean approach, particularly when also considering exporting to HTML.

AFAIK this is the first plugin to leverage the new capabilities of the image generator plugin infrastructure, I ran into a few tiny bugs, I think...

I'd be very grateful for any and all feedback on this!


maclomhair:
I realize that the tests still need fixing and updating, and I think ideally, we could just use sha256 instead of md5 for hash generation - even though there are no security issues here, changing it should be super easy and effortless, especially since this code path was unused (AFAIK) so far!

maclomhair:
Currently, the unit tests are failing because both the old file-based and the new embedded equation plugins reference the same help page. This is also because currently the embedded equations are a whole new plugin and I don't know how to tackle that properly. 

Maybe, if you have time @jaap-karssenberg could you provide guidance on how this could be properly done in your opinion? In the [commit where you added the framework](https://github.com/zim-desktop-wiki/zim-desktop-wiki/commit/55ab6f2b69ee3e745bd267e56f11a53b7ad10aa3) you mentioned a toggle in the settings but I'm not sure how that would be technically implemented as the image generator extension classes are defined in the plugin on load time...

One way would maybe also be to just merge the little fixes to make the embedded image generator plugin infrastructure work and keep the inline equations plugin out-of-tree for now... whatever you prefer!

jaap-karssenberg:
My thinking was that it should be one plugin supporting multiple object types. This means the automatic discovery of the generator class should be overruled by explicit identification of the generator class in the objecttype class.

I think the real bottleneck then is how to do the user side. How do users interact with these options?
- Should the insert menu have entries have options for both versions or only one?
- If only one, there should be a notebook property to control to preferred type (inline or separate file)
- Should there be a way to deviate from the default - e.g. in the Insert Equation dialog?
- Should there be a way to convert an equation to the other variant?

Probably best to start simple, having 2 entries in the menu is not a big deal to start. However, also need to think forward when all image generator plugins have an inline version, then the menu will be unreadable with duplicates.

Hope this helps,

Jaap

maclomhair:
Thanks! That helps a lot and I now know how I can continue (once I find the time)!

maclomhair:
Thanks to your comment, I figured out that indeed it is very easy to have both kinds of equations supported by the same plugin (I didn't even need to overrule the automatic generator class detection, don't know why I thought that was a problem) - so thanks again!

Lacking better ideas, I now left the options to insert equatios as two separate menu entries... Probably it'd be better to have some sort of option to check in the insert dialog on which format to pick for saving the equation to disk, but maybe that's good enough for a first version?

Placing the format option in the insert dialog would have the advantage of providing the infrastructure also for the other image generator plugins.

maclomhair:
@introt I hope it's ok to tag you like this. I was wondering wether you'd be open to review/discuss possible improvements to this PR? (As to not take too much of Jaap's time before it is reasonably ready for merge...) I have the feeling it's not yet polished enough but I'm also unsure of how to move forward to get it into shape. Thanks a lot!

zim/plugins/base/imagegenerator.py
```diff
@@ -159,7 +160,11 @@ def set_from_generator(self, text, image_file):
 		# TODO: use index table to keep track and clean up when ref count is zero ?
 		self.data = text
 		self.image_file = self._new_image_file()
-		image_file.moveto(self.image_file)
+		try:
+			image_file.moveto(self.image_file)
+		except FileExistsError:
```
introt, line 165: In what situation is this error raised? Would it be more appropriate to forcefully overwrite the file to ensure it is up-to-date rather than to fail silently?

zim/plugins/base/imagegenerator.py
```diff
@@ -140,13 +141,13 @@ def __init__(self, notebook, page, generator, attrib, data):
...
-		cache_dir = self.notebook.folder.folder('_images')
+		cache_dir = self.notebook.cache_dir.folder('images')
 		if self.data:
 			content = []
 			for k, v in sorted(self.attrib.items()):
-				content.extend([k, v])
+				content.extend([k, str(v)])
 			content.append(self.data)
-			basename = hashlib.md5(''.join(content).encode()).hexdigest() + self.generator.imagefile_extension
+			basename = hashlib.sha224(''.join(content).encode()).hexdigest() + self.generator.imagefile_extension
```
introt, line 150: I'd leave md5 here; as you said it's not a security concern and the hash is only used to check if the image exists in cache -- no need to spend the watts

zim/plugins/equationeditor.py
```diff
@@ -21,6 +23,8 @@ class InsertEquationPlugin(PluginClass):
 		'name': _('Insert Equation'), # T: plugin name
 		'description': _('''\
 This plugin provides an equation editor for zim based on latex.
+Equations are stored as separate files in the attachments folder,
+or in the wiki page itself.
```
introt, line 27: nit, I'd prefer a "can" to imply user choice + why, eg. "can be stored either in-line in the page source, or as files in its attachments folder that can be modified using an external editor" -- the manual can then expand on that re: automated change detection and re-rendering

zim/plugins/equationeditor.py
```diff
@@ -62,6 +66,18 @@ def format_latex(self, dumper, attrib, data):
 		raise ValueError('missing source') # parent class will fall back to image
 
 
+class EquationImageObjectType(ImageGeneratorObjectType):
+
+	name = 'equationembedded'
```
introt, line 71: I'd follow the `image+x` pattern -- assuming this needs to be unique, `image+embeddedequation`?

introt (changes requested):
Hi! Sorry for the delay. I agree that _an_ inline toggle belongs to the edit dialog -- you might want to override it on a per-equation basis in some workflows involving external editors, even if the default would be set per-notebook in the future. Haven't had a chance to properly play with this, but am open to collaboration -- hopefully in a more timely manner moving forward.

Issue #2728, opened 2025-01-08
New Indexed Full-Text Search doesn't find words with umlauts


sojusnik:
**Describe the bug**
When the new Indexed Full-Text Search is enabled, then the global search doesn't find words containing umlauts (in my case [German umlauts](https://www.berlitz.com/blog/german-umlaut-meaning-letters)) in the content of pages, but still finds such words in page titles.

**Environment (please complete the following information):**
 - Version: 0.76.0
 - OS: Ubuntu 24.04

maclomhair:
This might be a problem with the sqlite FTS5 tokenizer used (as described in [sqlite FTS5 docs](https://www.sqlite.org/fts5.html) section 4.3.1). Maybe the tokenizer replaces umlauts with their ascii counterparts. Do you find content matches if you search for the word without umlauts, e.g. "Apfel" instead of "Äpfel"?

sojusnik:
That's it! Just tried it out. But then why "Äpfel" in the page title are still found, but not in the content, where only searching for "Apfel" does?

maclomhair:
The reason for this is that two different search techniques are used to look in page titles and in the page texts with the full text search plugin. Simplified, one looks through all the page titles directly, one by one, to find any matches, while the other one keeps an index of words and pages where they occur, and only looks up the words of your search there to get the results. Now, the second one by default first "cleans" the words before putting them in the index, which includes removing any diacritics or making an "a" out of an "ä". I hope this explanation helps...

In any case, there should be a solution, i.e. restoring the old, expected behavior: tell the full text search plugin to not clean the words before putting them in the index. The solution should be a one-line fix, if you are feeling like it and want to try it, you can try changing line 168 in the file `zim/plugins/indexed_fts.py` from `tokenize = 'unicode61 remove_diacritics 2',` to `tokenize = 'unicode61 remove_diacritics 0',` and see if that does the trick. (If you try it, you might need to disable and reenable the plugin to make it work, or regenerate the index)

Maybe I can take a look eventually, too, I just don't know when.

sojusnik:
Changed to `tokenize = 'unicode61 remove_diacritics 0'`, saved the file, disabled the plugin, restarted zim, enabled the plugin and rebuilt the index (tools > check and update index), but Äpfel still isn't found.

maclomhair:
Hmm too bad... one last thing you could try is changing the line, closing zim, removing the `index.db` file and opening zim again to see if it works then...

[...]

Issue #2733, opened 2025-01-10
Issue with indexed_fts search plugin when query contains "*"


xvusrmqj:
**Describe the bug**
Search not support * wildcard ,  I search 【agick】 keyword  ， not find the result. so i add * 【*agick】， a error modal show .

add:
* I want to search ImageMagick keyword, and if search it , there are 3 results list. but search agick return 0 result, search *agick the error occured, 
* already enabled the indexed full-text search plugin 

**To Reproduce**
1. Go to Search Modal
2. input keyword with *
3. click find button 

**Expected behavior**

**Screenshots**
![image](https://github.com/user-attachments/assets/7c535e14-b64c-4270-b144-917ca6592baa)

**Environment (please complete the following information):**
 - Version [0.76.0]
 - OS: [windows]
 - Plugins:  already enabled the indexed full-text search plugin.

**Additional context**
the show text as follows :
This is zim 0.76.0
Platform: nt
Locale: zh_CN cp936
FS encoding: utf-8
Python: (3, 10, 8, 'final', 0)
PyGObject: (3, 42, 1)

======= Traceback =======
  File "zim/gui/searchdialog.py", line 101, in _search
  File "zim/gui/searchdialog.py", line 197, in search
  File "zim/search.py", line 319, in search
  File "zim/search.py", line 428, in _process_group
  File "C:\Program Files\Zim Desktop Wiki\_internal\share\zim\plugins\indexed_fts.py", line 102, in process_index_fts
    query_results = db.execute(
OperationalError: unknown special query: agick

jaap-karssenberg:
Modified title to be more descriptive of the actual issue. Also clarify this is about "search" not "find".

Regards,

Jaap

maclomhair:
Thanks for the bug report! I misunderstood the query language of the full text search engine, but I think I have an idea how to get the intended functionality working with indexed FTS. I'll see if I can make a pull request that addresses this issue.

maclomhair:
@xvusrmqj Could you see yourself testing the referenced pull request #2740 to see if it fixes this issue? Only if it's not too much of a hassle, of course!

xvusrmqj:
> [@xvusrmqj](https://github.com/xvusrmqj) Could you see yourself testing the referenced pull request [#2740](https://github.com/zim-desktop-wiki/zim-desktop-wiki/pull/2740) to see if it fixes this issue? Only if it's not too much of a hassle, of course!

[...]

PR #2740, opened 2025-01-14
Fix issues #2728 and #2733

maclomhair:
Issue #2728 is because the tokenizer was set to remove diacritics on words, but not in the search query. To be closer to the expected, traditional behavior of the search, I disabled the removal of diacritics now and introduced a change to detect if the full-text table schema is outdated, in which case the tables are recreated. This should fix the issue for everyone without having to delete the entire index.db.

Issue #2733 stems from a misunderstanding about the sqlite full-text search query language when I implemented the plugin. The query language doesn't actually support proper globbing. Instead, I now simulate the globbing by first globbing through all terms known to the index and constructing a query from that, then searching with that newly constructed query instead. I haven't done performance measurements on this approach but I expect that it should be unnoticable fast in general. A potential speedup might otherwise be to introduce an index on all tokens known to sqlite FTS.

I'm looking forward to the review and I'm happy to incorporate the review findings!

### DIFF
--- a/data/manual/Plugins/Indexed_Full_Text_Search.txt
+++ b/data/manual/Plugins/Indexed_Full_Text_Search.txt
@@ -7,3 +7,11 @@ Creation-Date: 2023-10-11T16:52:26+02:00
 This plugin allows to massively speed up the search in page contents, by up to 95%. Simply enable the plugin. The index will be automatically recreated, after which the faster full-text search will be available.
 
 This is achieved by caching a reverse index of tokens (i.e. words) in the index using ''sqlite''. For that to work, it requires a version of ''sqlite'' with the FTS5 extension. This new index will be around 5x bigger than before (for a 1500-page notebook with around 500 words per page, the index grows from 1.7 MiB to about 5.5 MiB).
+
+===== Options =====
+The option **Remove diacricitcs before indexing** alters the indexer such that for example "Apfel" and "Äpfel" are both indexed as "Apfel", and "avião" will be indexed as "aviao". This means searching for "apfel" will then return both pages with matches for "Apfel" and for "Äpfel". While this may be desireable for some use cases, it is turned off by default to more closely resemble the search functionality without this plugin.
+Note that currently, if "Remove diacritics before indexing" is set to true, in the example above, searching for "Äpfel" will not return any results due to implementation shortcomings of the full-text search module.
+
+The option **Additional token characters** can be used to set additional characters which will be considered part of a word (a "token"). Adding "-", for example, will lead to words such as "lorem-ipsum" being added to the index entirely, so they can be searched for using "lorem-ipsum" as a search string but not "lorem ipsum".
+
+After changing these options, the index of each notebook must be re-generated.
--- a/zim/plugins/indexed_fts.py
+++ b/zim/plugins/indexed_fts.py
@@ -94,11 +94,30 @@ sqlite.
 
 		# All keywords passed to this functions are content-related so
 		# we don't need to check the term.keyword property.
-		# Beware: FTS5 supports a complex search syntax, including "*"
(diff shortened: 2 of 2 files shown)


jaap-karssenberg:
I can't really contribute review here, as I'm not familiar with this SQLite module. But happy to merge as is, when you think the patch is good to go.

--- Jaap

maclomhair:
Ok! I'll see if I can get the bug reporters to test that the changes fix the issues for them to make sure... maybe I can get some more feedback. If that doesn't work out - I've looked over the changes again now and I'm confident that they do fix the issues.

roland-ruedenauer:
Hi @maclomhair,

I did some testing on Windows using your latest changes and found global search to work really well.
Search terms containing (german) umlauts is working for me and wildcard search using "*" too.

Some minor issues (not directly related to #2728 or #2733) I found

* a search term ending in a quotation mark can trigger an error if the prefix is short (e.g. "ref?")
* searching for phrases using double quotes as mentioned in the manual doesn't seem to be working - e.g. searching for "C++20"
* accidently searching for "*" took really long and I wasn't able to cancel the search via the cancel button (I finally had to kill the zim process)
* using && or || operators as mentioned in the manual does not work (also without indexed search pluging, I guess)

sojusnik:
@maclomhair Can you please check if your updates also fix searching for words containing hyphens, like in lorem-ipsum, since they aren't found in your initial version?

Happens on Ubuntu 24.04 with Zim 0.76.0.

maclomhair:
Thanks for your testing @roland-ruedenauer and your hints @sojusnik!

> * a search term ending in a quotation mark can trigger an error if the prefix is short (e.g. "ref?")

I'll fix this in the next revision of this pull request! Thanks for pointing it out!

> * searching for phrases using double quotes as mentioned in the manual doesn't seem to be working - e.g. searching for "C++20"

My testing showed that using double quotes in search works normally but the problem here is that the full-text search sees "+" as a separator token, so it indexes only "C" and "20" instead. It is possible to specify custom mapping for which characters are tokens and which are separators, and I think making the mapping customizable is the way to go here.

> * accidently searching for "*" took really long and I wasn't able to cancel the search via the cancel button (I finally had to kill the zim process)

I'll fix this by adding a fast return if mistakenly searching for "*", so that the search immediately terminates without results instead.

> * using && or || operators as mentioned in the manual does not work (also without indexed search pluging, I guess)

This seems to be a search bug, not related to FTS. Using "AND" or "OR" instead seems to work, though!

> Can you please check if your updates also fix searching for words containing hyphens, like in lorem-ipsum, since they aren't found in your initial version?

This again is because "lorem-ipsum" is by default split into "lorem" and "ipsum"... I plan to add an option where people can add custom letters which should be considered tokens to be indexed.

jaap-karssenberg:
If nothing else comes up, I'll merge these fixes on Thursday evening and push out a new release. Next "point" release will be roughly a month later, so further improvements can land there.

maclomhair:
I have now reworked the fixes a little bit to give users more options to fine-tune their full-text indexing. 

Specificly, I have added options to enable/disable the removal of diacritics, which by default is turned off to better resemble how traditional search works.

A second option that can be globally configured is a string where additional characters which are considered tokens (i.e. part of a word) can be specified. Adding "+" there, for example, will lead to strings such as "C++20" being indexed as one string.

I have also added a help text for both options which can be found in the manual.

The plugin remembers which options were set when the index was last created, and updating the options will automatically lead to the recreation of the full-text search index should it be necessary.

I think this approach is overall better suited for the different user needs regarding full-text search. I've tested the code a couple of times and I think it's good to go.

(Ah and of course, I've also added a short return path if accidentally searching for "*", and "?" should be properly escaped now)

Issue #2613, opened 2024-05-15
LinkMap plugin does not work in Ubuntu 24.04 LTE  because of "xdot" dependencies failed


max-ant:
**Describe the bug**
Ubuntu 24.04 Desktop, Zim 0.75.2 installed from AppCenter
In log:
```python 
WARNING: /usr/lib/python3/dist-packages/zim/plugins/linkmap.py:24: UserWarning: w xdot.DotWidget is deprecated, use xdot.ui.DotWidget instead  from xdot import DotWidget
```

**To Reproduce**
``` sh 
sudo apt install graphviz python3-graphviz xdot libxdot4
```
In Zim open preferences -> Plugins

``` python  zim -D
DEBUG: Found 'xdg.Mime', but no database - falling back to 'mimetypes'
DEBUG: Loading config from: <ConfigFile: /home/max/.config/zim/preferences.conf>
DEBUG: New extendable: <zim.plugins.InsertedObjectTypeMap object at 0x759a876754c0>
DEBUG: Loading plugin: pageindex
DEBUG: Loading plugin: pathbar
DEBUG: Loading plugin: toolbar
DEBUG: Loading plugin: insertsymbol
DEBUG: Loading plugin: printtobrowser
DEBUG: Loading plugin: sourceview
DEBUG: Load extension: <class 'zim.plugins.sourceview.SourceViewObjectType'>
DEBUG: register_object: "code"
DEBUG: Loading plugin: bookmarksbar
DEBUG: Loading plugin: attachmentbrowser
DEBUG: Loading plugin: backlinkpane
DEBUG: Loading plugin: journal
DEBUG: Loading plugin: spe
DEBUG: Loading plugin: tableeditor
DEBUG: Load extension: <class 'zim.plugins.tableeditor.TableViewObjectType'>
DEBUG: register_object: "table"
DEBUG: Loading plugin: tableofcontents
DEBUG: Loading plugin: tags
DEBUG: Loading plugin: diagrameditor
DEBUG: Load extension: <class 'zim.plugins.diagrameditor.BackwardDiagramImageObjectType'>
DEBUG: register_object: "image+diagram"
INFO: This is zim 0.75.2
DEBUG: Python version is sys.version_info(major=3, minor=12, micro=3, releaselevel='final', serial=0)
DEBUG: Platform is posix
DEBUG: Not running from a source dir
DEBUG: Set XDG_DATA_HOME to /home/max/.local/share
[...]

PR #2752, opened 2025-01-23
Fix deprecated import of DotWidget from xdot module #2613

norfcran:
Attempt to fix the following warning message related to BUG report #2613:
```
WARNING: /home/jklust/zim-desktop-wiki/zim/plugins/linkmap.py:24: UserWarning: w xdot.DotWidget is deprecated, use xdot.ui.DotWidget instead
  from xdot import DotWidget
```

### DIFF
--- a/zim/plugins/linkmap.py
+++ b/zim/plugins/linkmap.py
@@ -21,7 +21,7 @@ logger = logging.getLogger('zim.plugins')
 
 try:
 	import xdot
-	from xdot import DotWidget
+	from xdot.ui import DotWidget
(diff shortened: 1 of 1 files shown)


norfcran:
The same issue prevents a start of zim on macos version 15.2.

dehesselle:
Can you clarify: how does a warning prevent the start of the app?

jaap-karssenberg:
I think you need another fallback if the new "fixed" import fails to the old behavior. No way to guarantee everybody has the up to date dependency.

PR #2898, opened 2025-11-14
Feature : edit alt image

rui-nibau:
The feature adds a text field in the image edit box to modify the alt text.

Capture d’écran du 2025-11-13 20-47-10

### DIFF
--- a/tests/pageview.py
+++ b/tests/pageview.py
@@ -3551,7 +3551,7 @@ class TestPageviewDialogs(tests.TestCase):
 			}
 		)
 		file = tests.ZIM_DATA_FOLDER.file('zim.png')
-		buffer.insert_image_at_cursor(file, '../MYPATH/./data/zim.png')
+		buffer.insert_image_at_cursor(file, '../MYPATH/./data/zim.png', alt='Alternate Text')
 		dialog = EditImageDialog(None, buffer, notebook, Path(':some_page'))
 		self.assertEqual(dialog.form['width'], 48)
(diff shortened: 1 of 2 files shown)


bruvellu:
How does the alt text get rendered in the wiki syntax?

rui-nibau:
Hello.

It seems that it is not documented in the manual.

Image with Alternate text is rendered as : 

`{{image.jpg|Alternate text}}`

It's a native feature of Zim.

Merge request just add a way to edit it graphically.

tests/pageview.py
```diff
@@ -3551,7 +3551,7 @@ def testVarious(self):
 			}
 		)
 		file = tests.ZIM_DATA_FOLDER.file('zim.png')
-		buffer.insert_image_at_cursor(file, '../MYPATH/./data/zim.png')
+		buffer.insert_image_at_cursor(file, '../MYPATH/./data/zim.png', alt='Alternate Text')
```
introt, line 3554: Changing instead of adding should be fine here since we weren't testing for empty alt text before and both variants are included in the export tests -- though now that I thought about it, those do not cover an alt text without other attributes nor equations with an alt text (do we want to also add this feature there or to hardcode "image of an equation" in the plugin?), see https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/e0f94978f8cc85c7e7da1997966e9cc223c9d519/tests/data/formats/wiki.txt#L49

PR #2899, opened 2025-11-17
fix: (ui) notebook dialog path color in dark theme

rui-nibau:
With dark theme on, The path of a notebook in the Notebook dialog is really hard to read.

This fix adjusts contrast between foreground and background.

Color used : [Gnome HIG color palette](https://developer.gnome.org/hig/reference/palette.html)

### DIFF
--- a/zim/gui/notebookdialog.py
+++ b/zim/gui/notebookdialog.py
@@ -23,6 +23,7 @@ import zim.main
 from zim.newfs import FilePath, LocalFile, LocalFolder
 from zim.notebook import get_notebook_list, get_notebook_info, init_notebook, NotebookInfo
 from zim.config import data_file
+from zim.config.manager import ConfigManager
 from zim.gui.widgets import Dialog, IconButton, encode_markup_text, ScrolledWindow, \
 	strip_boolean_result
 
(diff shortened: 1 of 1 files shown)


zim/gui/notebookdialog.py
```diff
@@ -98,6 +99,12 @@ def __init__(self, notebooklist=None):
 		else:
 			self.notebooklist = notebooklist
 
+		# Fix path foreground on dark themes
+		# Dark theme : light 5 (#9A9996)
+		# Light theme : dark 2 (#5E5C64)
+		dark_theme = ConfigManager.preferences['GtkInterface']['prefer-dark-theme']
+		self._path_color = '#9A9996' if dark_theme else '#5E5C64'
```
introt, line 106: Zim's color scheme is based on the Tango palette, but the original color `#5a5a5a`, your replacement for it (`#5E5C64`) and the suggested dark theme variant are all from outside it according to [Wikipedia](https://en.wikipedia.org/wiki/Tango_Desktop_Project). I'd suggest replacing these with the closest ones from the Aluminium group, but Jaap has the final word on design matters.

rui-nibau: I choose those colors to align with the modern Gnome HIG color palette as Zim is a GTK app and to stay close to the previous colors.

I will try the same thing with the Tango palette.

jaap-karssenberg:
I'll merge this fix for now as it mitigates the issue.  My ideal solution would be to query the theme on the "insensitive" text color or a similar theme element and use that.  Or use the "styles.conf" configuration.

rui-nibau:
Hello.

> My ideal solution would be to query the theme on the "insensitive" text color or a similar theme element and use that.

I don't know how to do that but I will investigate

> Or use the "styles.conf" configuration.

This I know how to do.

Thanks

PR #2909, opened 2025-12-11
versioncontrol plugin : notebook properties

rui-nibau:
The merge request adds notebook properties to the versioncontrol plugin : it allows to override global configuration on a notebook basis.

Use case : We enable autosave globally but we want to disable it on a specific notebook because it is part of a project already versionned and we don't want the zim commits to "interfere" with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ debian/python-module-stampdir/
 
 # Rider
 .vs
+
+.pylintrc
+.vscode/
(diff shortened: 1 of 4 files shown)


introt (changes requested):
This is a useful feature, thank you!

Please document the feature in the Manual and - since our runner seems to be broken atm - run the test suite locally to confirm that it's happy (should be since the property labels are equal to the preferences).

rui-nibau:
Hello.

will update documentation this week-end

The tests were ok when i run them locally. Will do another pass.

rui-nibau:
- [x] Documentation updated
- [x] Test suite ok locally (Fedora 43, python 3.14.2)

PR #2931, opened 2026-02-23
Accept both image/png and image/apng MIME types for a PNG image

maclomhair:
On my platform, "image.png" is identified as an APNG (an animated PNG file). Since PNG is a valid filename extension [according to Wikipedia](https://en.wikipedia.org/wiki/APNG), I presume that this is not a bug in the MIME database and not an indication that no MIME database is present either.

However, this leads to zim not accepting the `xdg.Mime` module and failing to open certain files properly (e.g. FreeCAD files are not recognized that way) on my system.

This fix allows using the xdg.Mime module in environments where the `image/apng` type is selected for PNG images.

### DIFF
--- a/zim/newfs/base.py
+++ b/zim/newfs/base.py
@@ -574,7 +574,7 @@ mimetypes = None
 try:
 	import xdg.Mime as xdgmime
 	mytype = xdgmime.get_type('image.png', name_pri=80)
-	if str(mytype) != 'image/png':
+	if str(mytype) not in ('image/png', 'image/apng'):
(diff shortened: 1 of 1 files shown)


maclomhair:
Ah and according to the [pyxdg docs](https://pyxdg.readthedocs.io/en/latest/mime.html#xdg.Mime.get_type), `get_type` is deprecated and we could switch to `get_type2`. If this is desired, I can quickly replace all occurences and add this change to the pull request. I tested it on my system and it produced the same results. On the other hand, I don't think `get_type` is going away anytime soon, so we could just stay with it and not risk any regressions.

PR #2942, opened 2026-03-27
Add basic markdown support

ThorstenHackbarth:


oxidworks (commented):
Tested locally on Python 3.12 / zim 0.76.3 — all 81 tests pass (including 25 markdown-specific). Great work on this long-awaited feature!

The architecture follows zim's existing 3-level parser pattern nicely, YAML front matter is the right choice for metadata, and the test coverage is solid. A few findings from my review:

### 1. Checkbox roundtrip data loss

`CHECKED_BOX` (`[*]`) and `XCHECKED_BOX` (`[x]`) both dump as `[x]`, losing the distinction on save:

```
Input:   - [*] Done     (CHECKED_BOX, green checkmark)
Output:  - [x] Done     (XCHECKED_BOX, red X)
```

The parser already handles `[*]` correctly (line 105), so the fix is just in the Dumper:

```python
CHECKBOX_MARKS = {
    UNCHECKED_BOX: '[ ]',
    XCHECKED_BOX: '[x]',
    CHECKED_BOX: '[*]',       # was [x] — preserves roundtrip
    MIGRATED_BOX: '[>]',
    TRANSMIGRATED_BOX: '[<]',
}
```

### 2. `dump_line` return type

`dump_line` (line 789) returns a bare string `'---\n'` while all other dump methods return lists. This works by accident since Python iterates the string as individual characters, which happen to concatenate to the same result. Suggest:

```python
def dump_line(self, tag, attrib, strings=None):
    return ['---\n']
```

### 3. Design question: `__text__` semantics

Currently `__text__` maps to `MARK` (highlight). In CommonMark/GFM, `__text__` means **bold**. Users opening existing `.md` files in zim might be surprised. This may be intentional, but worth documenting.

### 4. Minor: unrelated changes in PR

The PR includes fixes to `quicknote.py`, `searchdialog.py`, `indexed_fts.py` that seem unrelated to markdown support. Splitting them into a separate PR would make review easier for the maintainer.

Overall this is solid work and I'd love to see it merged. Happy to help with testing or additional edge case coverage.

christiang7:
It's working with a new notebook of markdown files, and it uses the Markdown markup, very nice, thanks. In my test, I encountered a problem with local links. Because I am working with a little odd naming structure for non-note files. For example a PDF file ``test.pdf``, I create this structure:

```bash
test.pdf
test.pdf.md
```

In the ``test.pdf.md`` is a local file link to ``test.pdf`` like here:

```bash
# test.pdf
[../test.pdf](../test.pdf)
```

After a refresh of ``test.pdf.md``, then the link changed to ``[../test.pdf](../test.pdf.md)``. If I use the ui for edit link, I get for the link destination ``..:test.pdf`` and it should be ``../test.pdf``. It would be nice, that the parser preserve local links and not linking to the note file. In the zim-wiki markup everything is fine and working as expected and no changed local links after refresh of page.

jaap-karssenberg:
> It's working with a new notebook of markdown files, and it uses the Markdown markup, very nice, thanks. In my test, I encountered a problem with local links. Because I am working with a little odd naming structure for non-note files. For example a PDF file `test.pdf`, I create this structure:
> 
> ```shell
> test.pdf
> test.pdf.md
> ```
> 
> In the `test.pdf.md` is a local file link to `test.pdf` like here:
> 
> ```shell
> # test.pdf
> [../test.pdf](../test.pdf)
> ```
> 
> After a refresh of `test.pdf.md`, then the link changed to `[../test.pdf](../test.pdf.md)`. If I use the ui for edit link, I get for the link destination `..:test.pdf` and it should be `../test.pdf`. It would be nice, that the parser preserve local links and not linking to the note file. In the zim-wiki markup everything is fine and working as expected and no changed local links after refresh of page.

I believe this is not due to the markdown support - it has to do with how zim resolves links. Potentially you get the same behavior with a ".txt" link. Will look into this.

oxidworks:
I traced through the code to verify. The issue is actually in the **Markdown parser/dumper**, not in the general link resolution logic.

**Before the fix (`b29b859`):**

1. `parse_link()` (markdown.py) converts `../test.pdf` → `..:test.pdf` by replacing `/` with `:`, treating it as an internal page link
2. `dump_link()` converts it back `..:test.pdf` → `../test.pdf` then unconditionally appends `.md` → `../test.pdf.md`

The general link resolution in `zim/notebook/index/pages.py` (`resolve_link`) is never involved here because `iter_href()` in `zim/formats/__init__.py:436` filters out non-page links (`link_type('../test.pdf')` returns `'file'` due to the `../` prefix matching `is_path_re`).

**The fix in `b29b859` is correct** — it checks `os.path.splitext()` before deciding whether to convert path separators or append `.md`. File references (`.pdf`, `.docx`, etc.) are now preserved as-is.

One minor edge case: if a Zim page name happens to contain a file extension (e.g., a page literally named `report.pdf`), the dumper would now treat it as a file reference and skip appending `.md`. Unlikely in practice, but worth noting.

jaap-karssenberg:
@ThorstenHackbarth  I've started integrating this branch - will push my changes to a branch in the repository over the weekend

christiang7:
The issue of local links is resolved, thank you. Unfortunately, I encountered other problems. 
First, the links to notes with brackets ``()`` in their name are not working. For an example: note with name ``note (first)`` and the link to it in another note ``Other note``:
```markdown
# Other note
[note (first)](note_(first).md)
```
The link destination is then ``note (first``, it should be ``note (first)``.
Second, if I create a link to a note with spaces in the name like the example before, then I get the HTML replacement character ``%20`` in the source of the file. Like the example before I get instead:
```markdown
# Other note
[note (first)](note%20(first).md)
```
I think, it should be with the ``_`` character. Is this intended or a bug? In zim-wiki markup its working as expected.

jaap-karssenberg:
For links I am thinking to add a "wiki link" extension to the native markdown. That also allows to keep the zim page convention. Than we only need to covert files in export markdown.

christiang7:
A solution could be to use internal links convention ``[[path/to/note]]`` used in other note-taking software like Obsidian or Foam.

oxidworks:
Reproduced both issues on `b29b859`.

**Bug 1 — brackets:** The link regex `[^)]+` at [`markdown.py:142`](https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/b29b859/zim/formats/markdown.py#L142) stops at the first `)`, so balanced parens in URLs (allowed by CommonMark) break parsing. The roundtrip is **destructive**:

```
Input:     [note (first)](note_(first).md)
Parsed:    href = 'note_(first'           (.md) leaks out as text
Re-dumped: [note (first)](note_(first.md).md)
```

Any notebook with files like `report_(v2).pdf` silently corrupts on save. The same regex is used for images on line 141. Needs either balanced-parens matching or backslash-escape support.

**Bug 2 — spaces:** [`markdown.py:731`](https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/b29b859/zim/formats/markdown.py#L731) encodes `' '` as `%20`. The *internal* roundtrip is consistent (line 461 decodes it back), but on disk Zim stores the file as `note_first.md`, not `note%20first.md` — so the link resolves to a non-existent file for anything outside Zim's own parser, and breaks Obsidian/Foam/pandoc interop.

The `[[wiki link]]` extension would solve both for internal page links, but external file references (`../attachments/report_(v2).pdf`) still go through this code path, so Bug 1 remains a data-loss hazard independently.

PR #3021, opened 2026-08-28
newfs: fix _joinuri() for UNC hosts with non-word characters (WSL mounts)

### LINKED ISSUE #2387: Unable to open zim folders on devices mounted through WSL
**Describe the bug**
I am on a dual-boot machine.  When booted up on my Windows drive, if I mount one of my Linux drives through WSL and try to open a Zim project folder, Zim will immediately crash with a TypeError.

Crash Log:
```
This is zim 0.75.1
Platform: nt
Locale: en_US cp1252
FS encoding: utf-8
Python: (3, 10, 8, 'final', 0)
PyGObject: (3, 42, 1)

======= Traceback =======
  File "zim/main/__init__.py", line 962, in main
  File "zim/main/__init__.py", line 733, in run
  File "zim/main/__init__.py", line 764, in _run_cmd
  File "zim/main/__init__.py", line 806, in _run_main_loop
  File "zim/main/__init__.py", line 268, in run
  File "zim/main/__init__.py", line 227, in build_notebook
  File "zim/main/__init__.py", line 158, in get_notebook_argument
  File "zim/notebook/info.py", line 48, in resolve_notebook
  File "zim/notebook/info.py", line 159, in __init__
  File "zim/newfs/base.py", line 255, in __init__
TypeError: Cannot convert None to a FilePath
```

**To Reproduce**
1. Mount an ext4 partition with WSL
2. Attempt to open any Zim project stored on the WSL drive
3. Error results

**Expected behavior**
It should just open the project folder without issue

- Version 0.75.1
 - OS: Windows 10
 - Language: US English

**Additional context**
- This bug does not occur in Linux.  Opening zim projects stored on windows drives works just fine under Linux.
- This bug does not occur when mounting ext4 drives with ext2fsd, though this is not useful as ext2fsd does not have functioning file writing under Windows 10.

agu2347:
Fixes #2387

### Problem

`zim.newfs.base._joinuri()` validates that a path starts with a UNC host (`\\host`), but the branch that actually builds the URI only matched hosts against `^\\\\\w+$` (or a drive letter). UNC hosts containing characters outside `\w` pass the initial validation but don't match either `elif`, so the function falls through and implicitly returns `None`.

This affects real-world paths: Windows exposes WSL distributions as `\\wsl$\Ubuntu-20.04\...` or `\\wsl.localhost\Ubuntu\...`, where `$` and `.` aren't matched by `\w`. Opening a notebook that lives on such a mount fails with a confusing `TypeError: Cannot convert None to a FilePath` deep inside `FilePath.__init__`, far from the actual root cause.

### Fix

Replace the overly strict `elif re.match(r'^\\\\\w+$', ...)` with a plain `else`. Once the earlier check has confirmed the path starts with a drive letter or a UNC-host prefix, and it isn't a drive letter, it must be a UNC host — so build the URI without re-validating the host name against `\w`. Ordinary UNC hosts (e.g. `\\myserver\share\...`) are unaffected since they already matched the old regex too.

### Test

Added `testWindowsUNCHostWithSpecialCharacters` in `tests/newfs.py`, covering `\\wsl$\...`, `\\wsl.localhost\...`, and a plain UNC host as a no-regression check. Since `_joinuri`/`_joinabspath` in `zim.newfs.base` are only defined when `os.name == 'nt'` at import time, the test patches `os.name` and reloads the module (via `importlib.reload`) to exercise the Windows-specific code path on any platform, restoring the original state in a `finally` block.

Verified locally: the new test fails against the current `develop` code (assertion error from `_joinuri` returning `None`/raising) and passes with the fix applied.

### DIFF
--- a/tests/newfs.py
+++ b/tests/newfs.py
@@ -98,6 +98,46 @@ class TestFilePath(tests.TestCase):
 		))
 
 
+	def testWindowsUNCHostWithSpecialCharacters(self):
+		# Regression test for the Windows-specific `_joinuri()` helper in
+		# zim.newfs.base, which used to silently return None (later causing
+		# "TypeError: Cannot convert None to a FilePath" much further down
+		# the call stack) for UNC hosts containing characters outside \w,
+		# such as the "\\wsl$\..." and "\\wsl.localhost\..." paths Windows
+		# uses to expose WSL mounts.
+		#
+		# This exercises the Windows-specific code path (`_joinuri` /
+		# `_joinabspath` in zim.newfs.base), which is only defined when
+		# `os.name == 'nt'` at import time. To test it on non-Windows
+		# platforms too, reload zim.newfs.base with `os.name` patched to
+		# "nt", then restore it afterwards.
+		import importlib
+		import zim.newfs.base as base
+
+		orig_os_name = os.name
+		orig_home = os.environ.get('HOME')
+		try:
+			os.name = 'nt'
+			os.environ['HOME'] = 'C:\\Users\\fakehome'
+			importlib.reload(base)
+
+			for path, expected_uri in (
+				(r'\\wsl$\Ubuntu-20.04\home\user\notebook',
+					'file://wsl%24/Ubuntu-20.04/home/user/notebook'),
+				(r'\\wsl.localhost\Ubuntu\home\user\notebook',
+					'file://wsl.localhost/Ubuntu/home/user/notebook'),
+				(r'\\myserver\share\notebook', # plain UNC host - no regression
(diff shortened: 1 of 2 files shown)


jaap-karssenberg:
As a side effect of the importlib usage in the test, other test cases now start failing :(

agu2347:
Thanks for catching that and cleaning it up — the `importlib.reload` approach in my test was clumsy and I should have avoided mutating the shared module object. Appreciate the follow-up fix, and nice touch extending UNC handling to the non-Windows branch too.

